PR #6, opened 2020-09-04
Update retris to 0.5.4-1

LinusCDE:
Hi,
I added some cli integration to the software.

The most interesting change is probably that retris will no more clash with remux by recognizing xochitl, failing to kill it but then only offer to exit by starting xochitl.service (which will mean to xochitl instances).

### DIFF
--- a/package/retris/package
+++ b/package/retris/package
@@ -1,13 +1,13 @@
 # vim: set ft=sh:
 pkgname=retris
-pkgver=0.5.4
+pkgver=0.5.4-1
 pkgdesc="Implementation of rust tetris_core on the reMarkable using libremarkable"
 section=games
 maintainer="Linus K. <linus@cosmos-ink.net>"
 license=MIT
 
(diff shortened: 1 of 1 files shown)


LinusCDE:
Note: The workflows [can't create docker images](https://github.com/LinusCDE/toltec/runs/1074174499) for me because my name contains capital letters. Searched a lot how I can make them lowercase, but [there doesn't seem to be one](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#functions).

It's only a really minor thing for me though.

LinusCDE:
Also just added plato to the mix.

Since the program likes a standalone directory and some may want to use `/opt/opt/` for compatibility purposes, I put it into `/opt/lib/plato/`. The config is then just symlinked to `/opt/etc/plato.toml` and media to `/home/root/plato-media`.

I've done several test so far and it seems to work well for me.

PR #7, opened 2020-09-05
Add plato

LinusCDE:
Hi,
also just added plato to the mix.

Since the program likes a standalone directory and some may want to use /opt/opt/ for compatibility purposes, I put it into /opt/lib/plato/. The config is then just symlinked to /opt/etc/plato.toml and media to /home/root/plato-media.

I've done several test so far and it seems to work well for me.

### DIFF
--- a/package/plato/postrm
+++ b/package/plato/postrm
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+# Opkg doesn't seem to remove that.
+# Or I've done something wrong (probably).
+rm -r /opt/lib/plato/
--- a/package/plato/package
+++ b/package/plato/package
@@ -1,0 +1,52 @@
+# vim: set ft=sh:
+pkgname=plato
+pkgver=0.9.1-1
+pkgdesc="Port of the plato document reader to the reMarkable"
+section=remarkable-apps
+maintainer="Linus K. <linus@cosmos-ink.net>"
+license=AGPL-v3
+
+origin=https://github.com/LinusCDE/plato.git
+revision=014c935b4fb19918ce608c36ca45343dd3815b82
+
+build() {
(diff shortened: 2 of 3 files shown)


package/plato/postrm
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# Opkg doesn't seem to remove that.
```
raisjnn, line 3: is the newly created symlink preventing removal of this dir? (because opkg wouldn't know to remove the symlink because opkg didn't add it). i'm curious if you can check the symlink into the package? (or if that's technically impossible)

LinusCDE: Which symlink do you mean? I have defined 3. But this is not a link itself. So even if it contains some symlinks itself, I don't see the problem why removing shouldn't work. Symlinks are just files, but I'm not sure whether they may affect opkgs checks.

Symlinks are:
- `/opt/bin/plato` -> `/opt/lib/plato/plato.sh`
- `/opt/lib/plato/media` -> `/home/root/plato-media/`
- `/opt/lib/plato/Settings.toml` -> `/opt/etc/plato.toml`

raisjnn: I'm assuming that opkg has logic to only remove empty directories, so if you add a symlink inside `/opt/lib/plato/`, it will prevent `/opt/lib/plato` from being removed. 

Looking at opkg source (to validate assumption), I see this special logic: https://git.yoctoproject.org/cgit/cgit.cgi/opkg/tree/libopkg/opkg_remove.c#n110

raisjnn: my tentative recommendation is to check if your symlinks still exist and if they are syminks. if so, remove the symlinks. otherwise, leave the whole dir.

PR #10, opened 2020-09-06
Add appmarkable

LinusCDE:
This is a fairly dump ui, just to run a process that doesn't have one.

I use this as a dependency for my other package rmServeWacomInput (PR follows shortly)

### DIFF
--- a/package/appmarkable/package
+++ b/package/appmarkable/package
@@ -11,10 +11,12 @@ origin=https://github.com/LinusCDE/appmarkable.git
 revision=b4226e896f441af9895ed5a4ff183af7f93b11ae
 
 build() {
-    cargo build --release
-    $STRIP target/armv7-unknown-linux-gnueabihf/release/appmarkable
+    rm -r .cargo/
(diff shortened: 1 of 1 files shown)


raisjnn:
i believe that packages now require an 'image=' section. in your case, it will be 'image=rust' i believe (the Dockerfiles are located in `images/<image>/Dockerfile` if you want to adjust it

you can test building locally by running `make appmarkable` (and it should build inside the specified docker image)

LinusCDE:
Cool. I'll do that later.

After those two PRs should I then create a single PR to update all my other packages accordingly at once?

raisjnn:
i think matteo already added the image to the existing packages in the repo

>

LinusCDE:
I added the image both this and PR #11 . I couldn't really test it now though as docker is acting up on my machine and, after a full day of coding, I don't feel like switching the kernel now or fixing it otherwise.  
My Pi seems to have some problems build the dockers too (not architecture related). So I'll revisit getting that to work later.

raisjnn:
i'll test it on my machine

PR #11, opened 2020-09-06
Add rmservewacominput

LinusCDE:
This is the package of my old program that just sends out ev events from the pen on a tcp server.

Since I use this still regularly but wanted to do away with all my button shortcut c programs as systemd services, I appified this. For that I made a dumb UI which you can find in PR #10


raisjnn:
tested that the build works locally, gonna merge

PR #17, opened 2020-09-08
Update oxide to v1.3-beta

Eeems:


package/oxide/package
```diff
@@ -1,6 +1,6 @@
 # vim: set ft=sh:
 pkgname=oxide
-pkgver=0.0.0-4
+pkgver=0.0.0-5
```
matteodelabre, line 3: I think that when upstream has a versioning scheme (as it is seems to be the case with oxide), it makes sense to use it in the package version. We just need to be careful with the _beta_ component of your version number, since we need to make sure that if a non-beta v1.3 comes out, it is recognized as a newer version. For example, using `1.3-beta-1` would be bad because it would be sorted as a more recent version than `1.3-1` according to [the Debian version comparison rules](https://manpages.debian.org/wheezy/dpkg-dev/deb-version.5.en.html). We can leverage the special `~` symbol which is always sorted lower than any other character by those same rules.

TL;DR: I’d suggest using `pkgver=1.3~beta-1` as the package version.

matteodelabre (changes requested):

package/oxide/package
Eeems: Thanks, I decided not to touch it since I wasn't sure what version numbering is allowed in the ipkg format.

matteodelabre (approved):

matteodelabre:
Excellent, thanks!

PR #29, opened 2020-09-11
[WIP] Update oxide to 1.3.3-beta

Eeems:


Eeems:
Didn't get a chance to fully test this since `make oxide` doesn't work when trying to build locally. Looks like the Makefile doesn't handle creating the docker images.

matteodelabre:
In principle, `make oxide` should automatically pull the Docker images from `ghcr.io` if you don’t have them locally. If you really want to build them by yourself (can take a couple of hours), you can run `./scripts/build-images` (I’ll add a Makefile target for that, so that it’s easier to discover). Did you encounter a specific error when running `make oxide`?

Eeems:
> In principle, `make oxide` should automatically pull the Docker images from `ghcr.io` if you don’t have them locally (this does not require logging into `ghcr.io`). If you really want to build them by yourself, you can run `./scripts/build-images`. It can take a couple of hours however if you made significant changes to the Dockerfiles, otherwise it’ll just reuse cached layers pulled from `ghcr.io`. I’ll add a Makefile target for the `build-images` script so that it’s easier to discover.

Yeah, I'm still waiting on them to build.

> Did you encounter a specific error when running `make oxide`?
```
 eeems@shark  ~/git/Github/Eeems/toltec   oxide  make oxide
rm build/packages/"oxide" -fr
./scripts/build-package package/"oxide" build/packages/"oxide"
=> Fetching source files
Unable to find image 'ghcr.io/matteodelabre/toltec-qt:oxide' locally
docker: Error response from daemon: manifest unknown.
See 'docker run --help'.
make: *** [Makefile:38: oxide] Error 125
```

Eeems:
Looks like the automated build is failing to add the static library?
> main.cpp:(.text.startup+0x14): undefined reference to `qt_static_plugin_QsgEpaperPlugin()'

matteodelabre:
> ```
>  eeems@shark  ~/git/Github/Eeems/toltec   oxide  make oxide
> rm build/packages/"oxide" -fr
> ./scripts/build-package package/"oxide" build/packages/"oxide"
> => Fetching source files
> Unable to find image 'ghcr.io/matteodelabre/toltec-qt:oxide' locally
> docker: Error response from daemon: manifest unknown.
> See 'docker run --help'.
> make: *** [Makefile:38: oxide] Error 125
> ```

Oops, I didn’t foresee that building from a different branch than stable or testing could cause this kind of problem. The package build script uses an heuristic that pulls the Docker image that has the same tag as the current Git branch (so if you build from stable, it’ll use `toltec-qt:stable`, and from testing it’ll use `toltec-qt:testing`). The reasoning behind that was that changing the Dockerfiles in testing should not immediately affect the build in stable. Here it seems that you’re building from a branch called `oxide`, but of course there’s no Docker image tagged like that in the published repository, which explains the error.

I think this could be solved by moving the images to a different Git repo, versioning them properly and including that version into the `image=` field of each package. I’ve been thinking about doing that for a few days anyway, because a nice side effect of this strategy is that it makes it easier for other projects to reuse the images.

Do you or @raisjnn have any opinion on this?

Eeems:
> I think this could be solved by moving the images to a different Git repo, versioning them properly and including that version into the image= field of each package. I’ve been thinking about doing that for a few days anyway, because a nice side effect of this strategy is that it makes it easier for other projects to reuse the images.

@matteodelabre I think that's a good idea.

raisjnn:
I think that idea is fine, but there is a step "publish-images" that will fail for everyone who doesn't have access to those repos. 
I also propose that the individual repos should go under a single organization name (vs. putting them into matteo namespace).

matteodelabre:
> I think that idea is fine, but there is a step "publish-images" that will fail for everyone who doesn't have access to those repos.

I assume you’re referring to the GitHub workflow job that builds and publishes the images whenever a Dockerfile is updated. This part of the workflow will be moved to the new repo hosting the images, so this should not cause any failure. Pulling the pre-built images will always work even for people who don’t have committer access to the repo.

> I also propose that the individual repos should go under a single organization name (vs. putting them into matteo namespace).

I agree that it’s best if the repository is not tied to my personal GitHub account. Do you suggest creating an org specific to toltec?

matteodelabre:
(I’ll move these discussions to separate issues so that we don’t loose track of the original PR.)

Issue #20, opened 2020-09-08
Define a process for moving packages from testing to stable


matteodelabre:
* See [how Debian does it](https://wiki.debian.org/DebianTesting).
* See [how Arch Linux does it](https://wiki.archlinux.org/index.php/Official_repositories#Testing_repositories).

Eeems:
So do we also want an unstable repo where devs can manage their packages for testing deployments?

matteodelabre:
Can you clarify what would be the difference of aim between the testing branch and the unstable branch? To make things manageable, I’d say that it would be best to find a solution that minimizes the number of branches while still achieving the goal of actual stability for the stable branch.

Eeems:
From the Debian wiki:
> Packages from Debian Unstable enter the next-stable testing distribution automatically, when a list of requirements is fulfilled:
>
>  -  The package has been in "unstable" at least for 2-10 days (depending on the urgency of the upload).
>  -  The package has been built for all the architectures which the present version in testing was built for.
>  -  Installing the package into testing will not make the distribution more uninstallable.
>  -  The package does not introduce new release critical bugs. 

So basically unstable is being actively worked on by developers/maintainers and contains bleeding edge. Testing contains packages at versions that are deemed "stable" enough to move forward. After that they move from testing to stable.

If we don't want the overhead we could be like ArchLinux and just have testing and stable though.

LinusCDE:
I find this idea good, but I doubt, that we'll stay active all the time to maintain even more branches.

Although we also need to consider having some kind of separation between rM1 and rM2 software when they diverge. The simplest idea in my mind would be to have a check in a preinst file and abort the installation (if possible) when a user wants to install a software that makes not sense for his device (e.g. software that disables buttons on the rM2).

One check to find out whether it's a reMarkable 1 or 2 could be with:
```bash
if grep deviceid=RM100- /home/root/.config/remarkable/xochitl.conf >/dev/null; then
  echo "This is a first gen"
else
  echo "This is a second gen"
fi
```

My initial idea was to have a stable-rm1 and stable-rm2 branch, but it think the above idea is more maintainable.

What are your opinions on this?

Eeems:
Until we get a toolchain for the rM2 and can actually start building applications for it, I think we should hold off on any decisions relating to it.
I think we likely should setup some sort of automation around promoting things to stable to help with when we aren't as active.

matteodelabre:
[...]

PR #38, opened 2020-09-15
Add bootstrap script

LinusCDE:
Hi,

as suggested by @okeh, I created a script that combines installation, re-enabling, adding toltec and installing a package.

It currently uses [my fork of remarkable_entware](https://github.com/LinusCDE/remarkable_entware/). But in theory it should also work with the original one as it also detects whether toltec is added and adds it automatically.

Since the script my be shared on serveral servers due to the HTTPs implications with the default wget, it has a version number in the comments.

The usage for the one-liner is also in the script. For testing you can use this one-liner:

```
# wget -qO- https://static.cosmos-ink.net/remarkable/bootstrap.sh | sh -s
```

Packages to be installed can be added to the end.

---

@matteodelabre I also saw that you are hosting some other files on toltec as well. You can gladly copy [the wget binary repo](http://static.cosmos-ink.net/remarkable/artifacts) to your repo and use that repo in the file as well.  
My fork for remarkable_install.sh will automatically use that bootstrapped wget. I can also change the url in remarkable_install.sh as well then to have a universal server for all the packages, installers and thirdparty stuff.


bootstrap.sh
```diff
@@ -0,0 +1,204 @@
...
+
+
+ensure_opkg_available() {
+  if ! which opkg >/dev/null 2>&1; then # opkg command not found
+    log WARN "Failed to find opkg command." \
+        "" \
+        "Pleae add /opt/bin & /opt/sbin to your PATH by executing" \
+        'ssh root@10.11.99.1 echo '\\\'\''PATH=/opt/bin:/opt/sbin:$PATH'\'\\\'\'' >> ~/.bashrc'\' \
+        ""
+  fi
```
raisjn: this should be automatic if /opt/bin exists (don't ask user to run the command by hand)

bootstrap.sh
```diff
@@ -0,0 +1,204 @@
...
+
+
+  # Verify internet connection
+  if ! ping -c1 1.1.1.1 >/dev/null; then
+    throw "Couldn't connect to the internet. Please ensure that your reMarkable is connected to a WiFi-Network with internet access."
+  fi
+
+
+  # Verify dns
+  if ! ping -c1 remarkable.com >/dev/null; then
```
raisjn: no need to ping remarkable.com, just giving away info to them. maybe hit some other server

bootstrap.sh
```diff
@@ -0,0 +1,204 @@
+#!/bin/sh
+
+#
+# Bootstrap 0.1.0 for reMarkable (rM2 untested)
+# wget -qO- '<URL_to_bootstrap.sh>' | sh -s [packages]...
+#
+
+bootstrap_wget() {
+  if [ ! -d /home/root/.wget_bin ]
```
raisjn: is there a common dir, like ~/.config/ or ~/.cache that can be used instead of putting .wget_bin in the toplevel namespace?

bootstrap.sh
```diff
@@ -0,0 +1,204 @@
...
+  export PATH="/opt/bin/:/opt/sbin/:$PATH"
+}
+
+
+ensure_toltec_repo_added() {
+  if ! grep "^src/gz\b.*\bhttps://toltec\.delab\.re/" /opt/etc/opkg.conf >/dev/null 2>&1; then
+    # No active toltec repo found in opkg.conf
+    echo "src/gz toltec https://toltec.delab.re/stable" >> /opt/etc/opkg.conf
+    opkg update || throw "Failed to update opkg after adding toltec stable repo"
+    log INFO "Added the toltec stable repo to opkg"
```
raisjn: toltec stable has almost nothing in it. i feel like we should default to testing or move more stuff to stable. perhaps in the testing -> stable process we have a weekly or biweekly promotion cycle

bootstrap.sh
```diff
@@ -0,0 +1,204 @@
...
+  fi
+}
+
+
+main() {
+  # Select source for remarkable_entware repo
+  if [ -z "$REMARKABLE_ENTWARE_REPO_AUTHOR" ]; then
+    # Use LinusCDE's fork for the time beeing
+    # TODO: Switch to "Evidlo" as soon as PRs are merged
+    REMARKABLE_ENTWARE_REPO_AUTHOR=LinusCDE
```
raisjn: this should be pointed to toltec-dev/remarkable_entware whenever we can make that repo (imo)

raisjn (commented):
we should also have a CI test that uses this script to install a package and verifies that: 1) package was installed, 2) opkg was installed, 3) toltec was enabled

bootstrap.sh
LinusCDE: Agreed. I'll add that.

bootstrap.sh
LinusCDE: Is there a good server you'd recommend? I took remarkable.com since it should never be down and everyone would probably trust it. If I'd just taken google.com, some people probably wouldn't have liked that.

Maybe github.com (though there are companies that actually block that domain)?

bootstrap.sh
LinusCDE: .cache is a good idea. I took this folder, so a user would see it easily. I'll need to change that in my entware_fork then, too.

bootstrap.sh
LinusCDE: Then I'll remove the explicit mention of Evildo from there and we'll decide that later.

bootstrap.sh
LinusCDE: Yeah, but the users should have the stable repo. We should really just merge to stable more often. Like, when there are no bug reports for a software in 2 weeks or similar, it's considered stable. I think some distros do it similarly.

bootstrap.sh
Eeems: Maybe toltec.delab.re or one.one.one.one?

bootstrap.sh
Eeems: See #20

bootstrap.sh
LinusCDE: Cloudflare (one.one.one.one) would also be good. I don't want to take toltec.delab.re as this might at some point misreport no internet when there are in fact just maintenance works or similar. The big servers typically are all good for that since they're practicially never down at any given point.

We could also make the point, that I could discard those checks, since in order to run that script, most would download it with wget anyway.

bootstrap.sh
LinusCDE: Changed to ~/.cache/wget_bin in commit [b712a400](https://github.com/toltec-dev/toltec/pull/38/commits/b712a400c5ffe258e9cc5070adf0c5556af6c323) and [remarkable_entware/52e7c156](https://github.com/LinusCDE/remarkable_entware/commit/52e7c1566d83149894bb4a40b17432bd1c185038).

bootstrap.sh
LinusCDE: Adjusted in commit [d3b53182](https://github.com/toltec-dev/toltec/pull/38/commits/d3b531825d8b46a1b1b2ab89926e0f39b3e4c61c).

bootstrap.sh
LinusCDE: I removed those checks. Wouldn't matter if a one-liner tries to download that script in the first place.

raisjn:
I've used this script on three remarkable tablets to great effect! I've put it at https://rmkit.dev/bootstrap and am going to test it from there (not just cosmos-ink) next. 

The only thing that messed  up the script was in one tablet it found /opt/ that was a non opkg /opt and it aborted. I then deleted /opt (because I created it) and re-ran and it worked. When I ran on a tablet with opkg, it correctly recognized /opt and opkg and moved forward. 

My main issue with script is that I had to tell someone to use `testing` repo and that added 2 or 3 extra manual steps.

LinusCDE:
@raisjnn Nice to hear.

> I've put it at https://rmkit.dev/bootstrap and am going to test it from there (not just cosmos-ink) next.

Cool! As it's intended. Also a nice idea to trash the `.sh` extension. Would make any url cleaner.

> My main issue with script is that I had to tell someone to use testing repo and that added 2 or 3 extra manual steps.

I had that yesterday when writing my wall of text which is why the latest commit addresses that. You can add `TOLTEC_BRANCH=testing ` in front of the one-liner to use testing by default.

Though we should really pay attention that this doesn't become the default. It should only ever be used by devs, testers or maintainers. Never appear in a one-liner in some readme.

bootstrap.sh
```diff
@@ -0,0 +1,205 @@
...
+#
+
+bootstrap_wget() {
+  if [ ! -d /home/root/.cache/wget_bin ]
+  then
+      log INFO "Installing a current wget binary with better and current tls support" \
+          "to /home/root/.cache/wget_bin/ for this installation..." \
+          ""
+      # Bootstrap a current version of wget
+      WGET_BINARIES_PATH='http://static.cosmos-ink.net/remarkable/artifacts'
```
raisjn: where is this hosted?

LinusCDE: On my Raspberry Pi 4 at home:
- Nginx on Ubuntu Server 20.04
- 30 MBit/s Uplink
- Automatic updating of DNS entries should the ISP have problems (was lately down about 3 times a week for about 3 mins with new ip afterwards)

I can also easily move that to a VPS I have. Or better yet, we just host it under thirdparty on the toltec server.

The file was [built here](https://git.cosmos-ink.net/linus/wget-remarkable-pipeline). You can check out the latest artifact from the CI/CD section there. Should be same hash and commit and build logs are also all there.

Eeems: I'd feel much better if it was hosted somewhere that had redundancy.

matteodelabre: Is there a reason for not using the [wget package from Entware](http://bin.entware.net/armv7sf-k3.2/wget_1.20.3-4_armv7-3.2.ipk)?

LinusCDE: To install entware (mostly atm) from https in the first place.

matteodelabre: Here’s an alternative idea: we could run the install script from the context of a host connected to the rM through SSH. That way, we could leverage the up-to-date wget/curl from the host. I believe this would make things simpler. What do you think?

LinusCDE: In theory yes, but then we would also need to ensure that user has wget installed. Either downloading it on Windows (and installing on Mac) or requiring docker for just one small installation.  
Considering most end users don't run linux, this would be a hassle for both us and them.

matteodelabre: Since the initial `wget -qO- https://static.cosmos-ink.net/remarkable/bootstrap.sh | sh -s` invocation uses the insecure default `wget` from the reMarkable, doesn’t that make all further efforts towards avoiding MITM moot? It seems to me that an attacker can simply MITM the bootstrap.sh file in this case.

Eeems: We should include a hash check in that command. Instead of just piping it into sh.

LinusCDE: Agreed. That's the dilemma I see currently: Short one-liner vs secure script snippet. We should really go with the latter though.

LinusCDE: I hope that the reMarkable team notices this and gives their BusyBox a update (wget is part of that). Then we would use a longer script for about 2 months and could switch to a oneliner afterwards. We could also get rid of any wget bootstrapping.

Eeems: The secure snippet wouldn't really be that long though.

raisjn:
are we waiting on anything here or is this ready to go?

matteodelabre:
Overall this looks good, but there are a few nits to fix regarding formatting, variable naming and shellcheck compliance. I’ll review it more thoroughly tomorrow.

bootstrap.sh
matteodelabre: What about embedding the secure wget binary inside the `bootstrap` script? This way we 1) remove an additional HTTP request, and 2) only need to check the SHA-sum of the downloaded script, instead of both the script and the secure wget binary retrieved by the script.

The embedding could be done automatically by a build script, and the final `bootstrap` script could be automatically published to the Toltec server.

matteodelabre: The embedding could be done automatically by a build script, and the final `bootstrap` script could be automatically published to the Toltec server.

LinusCDE: > What about embedding the secure wget binary inside the `bootstrap` script? This way we 1) remove an additional HTTP request, and 2) only need to check the SHA-sum of the downloaded script, instead of both the script and the secure wget binary retrieved by the script.
> 
> The embedding could be done automatically by a build script, and the final `bootstrap` script could be automatically published to the Toltec server.

We could probably do that. This would make the script fairly big though. And the fact that the script already has the hash of the weg file builds a chain of trust in that case anyway. If bootstrap.sh has a known hash, that would change if we were to change the hash in that script. It's similar in essence to how git commits work as well, if I'm not mistaken.

EDIT: I also changed the url from above to a github one by now, btw. See the last commit. We could also move that repo into toltec. And have the CI by done by GitHub as well (though I don't know GitHub Actions well enough to to that rn which is why I use the GitLab CI on my server instead).

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+            echo "FATAL: Invalid hash" >&2
+            exit 1
+        fi
+
+        # Ensure to /home/root/.cache/wget_bin exists and is empty
+        if [ -d /home/root/.cache/wget_bin ]
+        then
+            rm -rf /home/root/.cache/wget_bin/*
+        else
+            mkdir -p /home/root/.cache/wget_bin
```
matteodelabre: Note that this will execute if /home/root/.cache/wget_bin exists and is not a directory. We should probably change the condition to `if [ -f /home/root/.cache/wget_bin/bin ]`.

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+        if ! echo "$WGET_BINARIES_SHA256  /home/root/$WGET_BINARIES_FILENAME" | sha256sum -c -
+        then
+            echo "FATAL: Invalid hash" >&2
+            exit 1
+        fi
+
+        # Ensure to /home/root/.cache/wget_bin exists and is empty
+        if [ -d /home/root/.cache/wget_bin ]
+        then
+            rm -rf /home/root/.cache/wget_bin/*
```
matteodelabre: This seems dangerous: what if the user happens to have existing data in a folder named like this? I think the script should fail if the directory already exists.

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+#
+
+bootstrap_wget() {
+    if [ ! -d /home/root/.cache/wget_bin ]
+    then
+        log INFO "Installing a current wget binary with better and current tls support" \
+            "to /home/root/.cache/wget_bin/ for this installation..." \
+            ""
+        # Bootstrap a current version of wget
+        WGET_BINARIES_PATH='http://github.com/LinusCDE/wget-remarkable-pipeline/releases/download/job254'
```
matteodelabre: For consistency, uppercase should be reserved for environment variables, leaving lowercase to normal variables. By the way, all those variables should probably be marked `local`.

matteodelabre: It’s still not clear to me why we can’t manually download the [wget from Entware](http://bin.entware.net/armv7sf-k3.2/wget_1.20.3-4_armv7-3.2.ipk) and extract it to get a secure `wget`, instead of relying on another repository.

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+LD_LIBRARY_PATH="/home/root/.cache/wget_bin/dist" /home/root/.cache/wget_bin/dist/wget \$@
+EOF
+
+        chmod +x /home/root/.cache/wget_bin/wget
+    fi
+
+    # Ensure this binary is used
+    if [[ $(command -v wget) != /home/root/.cache/wget_bin/wget ]]; then
+        PATH="/home/root/.cache/wget_bin:$PATH"
+    fi
```
matteodelabre: Shouldn’t the `/home/root/.cache/wget_bin` directory get cleaned up when the script ends?

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+    ensure_toltec_repo_added
+
+    if [ $# -ge 1 ]; then
+        install_packages "$@" || throw "Failed to install requested packages"
+    fi
+}
+
+entware_install() {
+    bootstrap_wget || throw "bootstrap_wget() failed!"
+    (wget -O- "https://raw.githubusercontent.com/$REMARKABLE_ENTWARE_REPO_AUTHOR/remarkable_entware/master/entware_install.sh" | sh) \
```
matteodelabre: Couldn’t we merge the `entware_install.sh` script into this script?

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+        || throw "Failed to download and/or install remarkable_entware."
+    log INFO "Entware was installed successfully."
+}
+
+entware_reenable() {
+    bootstrap_wget || throw "bootstrap_wget() failed!"
+    (wget -O- "https://raw.githubusercontent.com/$REMARKABLE_ENTWARE_REPO_AUTHOR/remarkable_entware/master/entware_reenable.sh" | sh) \
```
matteodelabre: Can we move the `entware_reenable.sh` script into this repository?

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+    log INFO "Entware was reenabled successfully."
+}
+
+ensure_opkg_available_from_bashrc() {
+    # Run by ensure_opkg_available
+    if ! grep '.*PATH=.*/opt/bin.*' /home/root/.bashrc | grep -v '^#' >/dev/null; then
+        # Path is not in .bashrc
+        echo -e '\n# Path added by bootstrap.sh' >> /home/root/.bashrc
+        echo 'PATH="/opt/bin:/opt/sbin:$PATH"' >> /home/root/.bashrc
+        log WARN '/opt/bin and /opt/sbin were not added to $PATH in ~/.bashrc. Fixed automatically.'
```
matteodelabre: Isn’t this redundant with `entware_install.sh`?

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+    if [ $# -eq 1 ]; then
+        log INFO "Installing package: $1"
+    else
+        log INFO "Installing packages: $*"
+    fi
+
+    opkg install "$@"
+}
+
+log() {
```
matteodelabre: The `log()` and `throw()` functions in this script could be merged with the `status()`, `section()` and `error()` functions in `scripts/package-lib`.

scripts/bootstrap
```diff
@@ -0,0 +1,200 @@
...
+
+ensure_opkg_available() {
+    if command -v opkg &> /dev/null; then # opkg command not found
+        export PATH="/opt/bin/:/opt/sbin/:$PATH"
+        messages="$(ensure_opkg_available_from_bashrc)"
+        if [[ -n $messages ]]; then
+            echo "$messages" # Print the caught output
+            log WARN "Please re-connect to your reMarkable after this" \
+                     "script is finished. Otherwise opkg and other" \
+                     "installed binaries won't be recognized!"
```
matteodelabre: Isn’t this script supposed to run on the reMarkable? How can this message be shown if the connection to the reMarkable has been lost?

matteodelabre (dismissed):
Thanks for your work @LinusCDE. Here are some comments.

scripts/bootstrap
LinusCDE: > For consistency, uppercase should be reserved for environment variables, leaving lowercase to normal variables. By the way, all those variables should probably be marked local.

Good find! Should be addressed with commit 35bf74fb.

> It’s still not clear to me why we can’t manually download the wget from Entware and extract it to get a secure wget, instead of relying on another repository.

The "problem" is that entware does a proper job with their packages. They work exclusivly in the /opt directory and don't assume anything about the rest of the system. Even though binaries are dynamically linked, the linker lib itself is always hardcoded in the binary. In case of entwares wget it is on /opt.

Here is an example of me running entwares wget binary without having /opt mounted.
```
reMarkable: ~entwget/ ls
wget      wget-ssl
reMarkable: ~entwget/ ./wget
-sh: ./wget: No such file or directory
reMarkable: ~entwget/ file wget
wget: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /opt/lib/ld-linux.so.3, for GNU/Linux 3.2.0, stripped
reMarkable: ~entwget/ 
```

The `file not found` also confused me first when tinkering with c a while ago. But the interpreter lib `ld-linux.so` is meant with that.

scripts/bootstrap
Eeems: It's instructing the user to reconnect after it's done running. Not that the user will lose connection.

scripts/bootstrap
LinusCDE: It seems pretty unlikely that a user would happen to have such a directory to me. Though you're right that this is a concern. Since I added a check for that path to be a file. Doing a rm on a directory is unnecessary and a error for a file is a good idea.

Addressed in commit d8731cd.

scripts/bootstrap
LinusCDE: That is probably a good idea. I didn't do it here, so we may reuse it afterwards if needed.

This commit 6330e6e should address this. It's not perfect but do you find it okay this way (feel free to make changes).

scripts/bootstrap
LinusCDE: Probably. I didn't feel comfortable to make that decision here though. We should talk with @evildo again to maybe get his repo either into [toltec-dev](https://github.com/toltec-dev/) (and him, too) or merge the script outright.

scripts/bootstrap
LinusCDE: See https://github.com/toltec-dev/toltec/pull/38/files/515ea8a7dafcb877ba87a870acc7945f87e1a047#r498383687

scripts/bootstrap
LinusCDE: I don't think so, but I'm not sure. `entware_install.sh` may add `/opt/bin:/opt/sbin` temporarily (not sure on that) and just tells the user how to add the path permanently. The part here does that for the user if opkg was installed automatically or the user forgot it.

scripts/bootstrap
LinusCDE: I'm not entirely sure where see the problem with where it runs. The bootstrap script is always meant to run only on the reMarkable and don't make any assumptions on the users computer os that way.

That said, checking whether the function echo'ed something is still dirty. I also should clean up or reduce or entirely remove `ensure_opkg_available()` and `ensure_opkg_available_from_bashrc()` since they're only run once anyway. Probably a case of YAGNI.

scripts/bootstrap
LinusCDE: Good idea. Though we would then also supply those scripts as well. Do have an idea on how to use them without straight copying them?

scripts/bootstrap
LinusCDE: > It's instructing the user to reconnect after it's done running. Not that the user will lose connection.

Yes. (I didn't saw the comment when I wrote the previous one.)

Is there a way to apply the updated $PATH to the users bash directly. But aside from spawning a child bash with exec I didn't find a good way to do that. So this message was my only solution.

scripts/bootstrap
LinusCDE: See https://github.com/toltec-dev/toltec/pull/38#discussion_r498378708

scripts/bootstrap
Eeems: `export PATH=$PATH:...`?

LinusCDE: Only if you source the script. While we could do that in the one liner if someone downloads the script and runs it manually (which we shouldn't discourage imo), it wouldn't work since the export is applied the the child process the script is run in.

Eeems: When running the script source it instead of calling it and you can export variables.

LinusCDE: We could use that for the bootstrap command. But some people may wonder why. Also should people download and execute the script themselves, it wouldn't work.  
Is there a way to detect whether a script was sourced?

Eeems: https://stackoverflow.com/a/2687092

scripts/bootstrap
matteodelabre: Good catch, I didn’t think about that! Then I think that it would be best to move the wget build over to this repo, and to publish it on `https://toltec.delab.re`. I’ll do that in a separate PR if it’s OK with you.

scripts/bootstrap
matteodelabre: I think a good solution would be to have a generic script which installs Opkg and accepts as arguments a list of repositories to enable. Then it could be used both to install Toltec and to install Entware only. (By the way you probably meant @Evidlo.)

scripts/bootstrap
matteodelabre: Okay, so this is related to the thread above about merging with `remarkable_entware`.

scripts/bootstrap
matteodelabre: We could have a build step which simply `cat`s the lib and this script, then publishes the result to `toltec.delab.re`.

scripts/bootstrap
LinusCDE: No problem. If you want, I can also move the [wget-remarkable-pipeline](https://github.com/LinusCDE/wget-remarkable-pipeline) over to the toltec-dev domain. Though a custom gitlab actions build would probably be helpful at that point anyway. So we might as well just recreate that repo. What do you think?

matteodelabre: A transfer sounds great, I can add a GitHub workflow to it if you want.

LinusCDE: Is tranferred to toltec-dev. We just to get `buildroot/base:20200814.2228` (from docker hub) to run `ci_build.sh` and zipping the `dist/` folder afterwards.

scripts/bootstrap
LinusCDE: If I'm not mistaken, remarkable_entware basically runs a provided installer from entware (which installs it and automatically adds entware) and fixes some libs as well for the remarkable.  
So we would basically need a combination of this one, entware_install.sh and entwares installer.

scripts/bootstrap
matteodelabre: Thanks! I’m not really familiar with buildroot. I wonder if it could be used to simplify the Dockerfiles in toltec-dev/toolchain?

scripts/bootstrap
LinusCDE: Good idea. This would make the repo more complex though. Would be nice if you could add that as I'm not that comfortable with GH Actions, yet.

scripts/bootstrap
LinusCDE: > I wonder if it could be used to simplify the Dockerfiles in toltec-dev/toolchain?

Only if you want to build your base image yourself I guess. Buildroot meant to build a Busybox or Systemd bases linux system from scratch along with a bootloader. It uses a similar buildsystem to the Linux kernel and basically downloads and builds every single binary from scratch for many architectures and boards. I use it at work for a RPi-Like board. That way our software will be built into a ready-to-flash sd.img file.

It may be useful for your custom oecore-like toolchain. But I'm not sure whether getting the header files is even a usecase.

scripts/bootstrap
matteodelabre: Another drawback is that the resulting file's checksum will change anytime we change either of the `package-lib` or the `bootstrap` script, and therefore we will need to remember to manually update the install command accordingly even if we only update the `package-lib`.

matteodelabre:
I’ve made some changes to the [wget building workflow](https://github.com/toltec-dev/wget). It is now fully standalone (no need to download additional shared libs), which should hopefully make this script a bit simpler. Binaries are automatically pushed to <https://toltec.delab.re/thirdparty/bin> when a release is tagged in toltec-dev/wget. The checksum of the latest version is `8d447c6eb0a39e705f45bea900b12eef07142ea3da0809aca4dd44fe4110cdfd`.

scripts/bootstrap
matteodelabre: Leaving this as is for now to move forward.

matteodelabre:
Just changed the script to use Toltec’s wget and to incorporate Entware’s installation script. External dependencies are now reduced to just Entware, which should make auditing the script easier.

It works on my install, but I would appreciate if any of you had time to test it, to find if I introduced regressions. ~~My initial tests seem to show that the error recovery mechanism does not call `error-cleanup` when suitable—something that will need to be fixed before merging.~~ (Fixed.)

Scenarios to test:

- [x] Without any pre-existing Entware install.
- [x] With extra packages to install, passed as arguments to `bootstrap`.
- [x] With the `testing` branch (note that I've lowercased the toltec_branch variable since it's not exported).
- [x] With an existing, enabled remarkable_entware install (it should add Toltec on top).
- [x] With an existing, disabled remarkable_entware install (it should re-enable the /opt mount and add Toltec on top).
- [x] With an existing, enabled Toltec install (it should be a no-op).
- [x] With an existing, disabled Toltec install (it should re-enable the /opt mount).
- [x] With a faulty network that corrupts the initial wget download.
- [x] With a network that becomes unavailable partway through the installation.

Remaining work:

- [x] Automatically publish the `bootstrap` script to `https://toltec.delab.re/bootstrap`, through the `.github/workflow/stable.yml` workflow.
- [x] Update install instructions in README.
- [x] Add license header to the `bootstrap` script.

raisjn:
can you add doc to the top of the script with invocation line? (and example of using wget)

here's what I get with various invocations while testing re-enabling a disabled entware installation:

```
reMarkable: ~/ wget -qO- https://rmkit.dev/bs | sh -s
wget: note: TLS certificate validation not implemented
sh: line 39: syntax error near unexpected token `<'
reMarkable: ~/ wget -qO- https://rmkit.dev/bs | sh -C
wget: note: TLS certificate validation not implemented
sh: line 39: syntax error near unexpected token `<'
reMarkable: ~/ wget -qO- https://rmkit.dev/bs | bash -C
wget: note: TLS certificate validation not implemented
INFO:  Re-enabling existing Entware install
INFO:  Adding /opt/bin to the PATH
INFO:  Use '/home/root/entware-reenable' to re-enable Toltec after a system update
reMarkable: ~/ wget -qO- https://rmkit.dev/bs | bash -C harmony
harmony: harmony: Is a directory
wget: note: TLS certificate validation not implemented
reMarkable: ~/
```

my main feedback here is that i think the `bash -C harmony` line installs harmony, but i don't actually have any confirmation or notices about it. it would be nice for a message to appear indicating which packages were installed. there should also be indication of which repo is being enabled (stable or testing).

matteodelabre:
The errors with `sh` are expected since this is a Bash script.

Note that piping wget output to bash is insecure since it does not support TLS. I couldn’t come up with a simple single-line chain of commands that verifies the script checksum before running it, so for now it’s [documented as three commands to execute](https://github.com/toltec-dev/toltec/pull/38/files#diff-04c6e90faac2675aa89e2176d2eec7d8R15-R17).

You probably meant to use the `-s` Bash flag instead of `-C`. The line you ran tried to read a script from a file called `harmony` which did not exist. If you run `wget -qO- https://rmkit.dev/bs | bash -s harmony`, you’ll see that there is actually feedback when additional packages are installed.

matteodelabre:
> there should also be indication of which repo is being enabled (stable or testing).

Fixed that.

raisjn:
did you add the example invocation to the top of the file? 

> errors expected with sh

unexpected for me, as the script started as an sh script and invocations mention sh.

matteodelabre:
Does this work for you @raisjnn?

~~Ideally we’d also need to make sure that the script’s checksum is always up-to-date in the README.~~ Check added.

scripts/bootstrap/bootstrap
```diff
@@ -0,0 +1,336 @@
...
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+#
+# bootstrap
+#
+# Install and configure Opkg, Entware and Toltec
+#
+# Arguments: Additional packages to install after setting up Toltec
+# Example: ./bootstrap harmony draft
```
raisjn, line 11: i was hoping it would be a line showing how to use it alongside curl or wget, f.e.: `wget -Qo- <url> | bash -s harmony draft`

raisjn (approved):
i just want this thing out the door (sorry it's taken so long, linus and matteo)

this script looks good (assuming it works for everyone else) and we can iterate on it after it lands. the bootstrap should ideally still be a one liner in the README, but its good enough for now.

scripts/bootstrap/bootstrap
matteodelabre: I understand, but I don’t feel comfortable recommending that users run a potentially unsafe chain of commands. And if we make it safe (i.e. include the file’s checksum in the example), then it’s no longer possible to include that line in the file itself.

scripts/bootstrap/bootstrap
```diff
@@ -0,0 +1,336 @@
...
+    fi
+}
+
+# Mount Entware to /opt
+entware-mount() {
+    mkdir -p /opt
+    mkdir -p "$entware_path"
+
+    # Create systemd mount unit to mount over /opt on reboot
+    cat > "$systemd_mount_path" << UNIT
```
Evidlo, line 114: Change this to `<<-` and you can indent the heredoc.

matteodelabre: But only if we indent with tabs :(

PR #39, opened 2020-09-18
Update oxide to v2.0-alpha

Eeems:


Eeems:
I pushed up a fix to the Makefile after making the release tag. Not sure how to use the git repository itself as the source anymore. Help?

okayzed:
```
Digest: sha256:99c14897e6e871ec7f601024d6d1de3406ccc668234f663bd4c0e35adfedc017
Status: Downloaded newer image for ghcr.io/toltec-dev/qt:v1.0
rm -rf .build
rm -rf release
mkdir -p .build/process-manager
cp -r applications/process-manager/* .build/process-manager
cd .build/process-manager && qmake erode.pro
Info: creating stash file /src/.build/process-manager/.qmake.stash
WARNING: Failure to find: ../system-service/dbussettings.h
make -C .build/process-manager all
make[1]: Entering directory '/src/.build/process-manager'
arm-linux-gnueabihf-g++ -c -pipe -O2 -std=gnu++11 -Wall -W -D_REENTRANT -fPIC -DQT_DEPRECATED_WARNINGS -DQT_NO_DEBUG -DQT_QUICK_LIB -DQT_GUI_LIB -DQT_QML_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -I. -I../../docker-toolchain/qtcreator/files/libraries -I../system-service -isystem /opt/x-tools/arm-remarkable-linux-gnueabihf/arm-remarkable-linux-gnueabihf/include/QtQuick -isystem /opt/x-tools/arm-remarkable-linux-gnueabihf/arm-remarkable-linux-gnueabihf/include/QtGui -isystem /opt/x-tools/arm-remarkable-linux-gnueabihf/arm-remarkable-linux-gnueabihf/include/QtQml -isystem /opt/x-tools/arm-remarkable-linux-gnueabihf/arm-remarkable-linux-gnueabihf/include/QtNetwork -isystem /opt/x-tools/arm-remarkable-linux-gnueabihf/arm-remarkable-linux-gnueabihf/include/QtCore -I. -I/usr/mkspecs/linux-arm-gnueabihf-g++ -o main.o main.cpp
main.cpp:13:10: fatal error: dbussettings.h: No such file or directory
 #include "dbussettings.h"
          ^~~~~~~~~~~~~~~~
compilation terminated.
make[1]: Leaving directory '/src/.build/process-manager'
make[1]: *** [Makefile:579: main.o] Error 1
make: *** [Makefile:60: erode] Error 2
```

looks like we need dbus dev package. you can install it inside the package (for now), and we'll also install it into the base image for qt (but it might take a while to build that image)

raisjn:
taking from nao package, to use a particular commit, the lines should look like:

```
source=(https://github.com/rmkit-dev/rmkit/archive/f70f78eea2446958239c4b26fc2795fdd87f189e.zip)
sha256sums=(71153c40d20e3982601fe6faa55e382a2f84e04280235ee4ee53ccd1be254437)
```

the .zip file is the commit hash you want to use, sha256sum is checksum of that file



looking at the compilation, it seems that lib dbus header is missing from the qt image. you can install it in your package file (with apt-get) and we'll also add it to the qt base image

Eeems:
The error is actually because I'm building the applications out of order. Tarnish needs to be built before erode.

Eeems:
I could use some help here. I believe I'm installing the correct headers package, but it's still failing to build.

raisjn:
You likely need to install the systemd deps + headers into the qt toolchain location since this is using a cross compilation chain for QT. if you look at the compilation command can see flags like "-I" and "-i" and "-L" that are pointing to the toolchain directories.

i'm not sure how to do the above off the top of my head though, might require compiling that systemd dependency using the arm compiler.

package/oxide/package
```diff
@@ -2,22 +2,22 @@
 pkgname=oxide
 pkgdesc="Launcher application"
 url=https://github.com/Eeems/oxide
-pkgver=1.3.3~beta-3
+pkgver=2.0.0~alpha-1
```
matteodelabre: To match upstream as closely as possible, I think we should version this as “2.0\~alpha-1”. The rules for comparing versions will work even if a third version component is added in subsequent releases (e.g. “2.0.1\~alpha-1” will be correctly sorted higher than “2.0\~alpha-1”).

package/oxide/package
```diff
@@ -2,22 +2,22 @@
...
 timestamp=2020-09-11T05:33Z
 section=launchers
 maintainer="raisjn <of.rajsn@gmail.com>"
 license=MIT
 
 image=qt:v1.0
-source=(https://github.com/Eeems/oxide/archive/v1.3.3-beta.zip)
-sha256sums=(83a0872c1757dd883b578145841987a11dcd6a274d8203b7c91e6fb61de3cc39)
+source=(https://github.com/Eeems/oxide/archive/045c16943c82ba8c30ee22b507d8b710754a74e0.zip)
```
matteodelabre, line 12: Using the SHA-1 sum of the commit seems redundant now that we have added the `sha256sums` field. Maybe we should favor clarity and use the version tag in the `source` field instead of the SHA-1 sum.

package/oxide/package
```diff
@@ -2,22 +2,22 @@
...
+sha256sums=(b119b8e2d57715089c6f67240933236e8d0c63e6a8bed947fc51e29afc6149e3)
 
 build() {
+    apt-get update && apt-get install libsystemd-dev
```
matteodelabre: Unfortunately, as mentioned in the conversation, this installs the x86 version of libsystemd files, which cannot be used by the ARM cross-compiler. We need to build the package from source instead.

Eeems:
Ah, so the toolchain we build is missing things that the toolchain provided by reMarkable has.

@matteodelabre Any chance you could add systemd headers to the base image?

matteodelabre (changes requested):
Only minor nitpicks. Thanks for the PR!

matteodelabre:
> Ah, so the toolchain we build is missing things that the toolchain provided by reMarkable has.
> 
> @matteodelabre Any chance you could add systemd headers to the base image?

Yup, good catch! I’m looking into this.

package/oxide/package
Eeems: I'm not using the tagged commit as I had to push a fix to the makefile.

package/oxide/package
Eeems: I've removed this for now and I'll wait on the updated image with systemd's headers.

matteodelabre (approved):

PR #42, opened 2020-09-20
package: add koreader 2020.08.01

raisjn:
what do you think of binary releases?

i also started going down source release path, but koreader build process actually needs a git repo because it uses submodules to manage dependencies.


raisjn:
i'm going to use a different feature branch for this

PR #43, opened 2020-09-21
package: update rmkit suite to latest

raisjn:
* fixes to remux for interacting with plato
* fixes to remux for power management
* add sync, check and upgrade to nao package manager

### DIFF
--- a/package/nao/package
+++ b/package/nao/package
@@ -2,8 +2,8 @@
 pkgname=nao
 pkgdesc="nao package manager"
 url=https://github.com/rmkit-dev/rmkit
-pkgver=0.0.1-12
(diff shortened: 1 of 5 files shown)


raisjn:
Noticed I need to bump versions (and should eventually pick a versioning scheme for rmkit)

package/nao/package
```diff
@@ -2,7 +2,7 @@
 pkgname=nao
 pkgdesc="nao package manager"
 url=https://github.com/rmkit-dev/rmkit
-pkgver=0.0.1-12
+pkgver=(0.0.1-11)
```
Eeems: Shouldn't this be bumped to `0.0.1-13`?

PR #44, opened 2020-09-21
package: add koreader 2020.08.01

### LINKED ISSUE #41: package koreader

raisjn:
> what do you think of binary releases?

> i also started going down source release path, but koreader build process actually needs a git repo because it uses submodules to manage dependencies.

### DIFF
--- a/package/koreader/koreader
+++ b/package/koreader/koreader
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+bash /opt/koreader/koreader.sh
--- a/package/koreader/koreader.draft
+++ b/package/koreader/koreader.draft
@@ -1,3 +1,4 @@
 name=KOReader
 desc=ebook reader
 call=/opt/bin/koreader
+term=killall -9 luajit
--- a/package/koreader/koreader.draft
+++ b/package/koreader/koreader.draft
@@ -1,4 +1,6 @@
 name=KOReader
 desc=ebook reader
+imgFile=koreader
 call=/opt/bin/koreader
+term=killall -9 luajit
 term=killall -9 luajit
--- a/package/koreader/koreader.draft
+++ b/package/koreader/koreader.draft
@@ -1,0 +1,3 @@
+name=KOReader
+desc=ebook reader
+call=/opt/bin/koreader
--- a/package/koreader/koreader.draft
+++ b/package/koreader/koreader.draft
@@ -3,4 +3,3 @@ desc=ebook reader
 imgFile=koreader
 call=/opt/bin/koreader
 term=killall -9 luajit
-term=killall -9 luajit
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -22,6 +22,6 @@ package() {
   cp -R "$srcdir"/* "$pkgdir"/opt/koreader/
 
   install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
-  install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png"
+  install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
   install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
 }
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -1,0 +1,26 @@
+# vim: set ft=sh:
+pkgname=koreader
+pkgdesc="KOReader"
+url=https://github.com/koreader/koreader
+pkgver=2020.08.01-1
+timestamp=2020-09-19T03:30Z
+section=utils
+maintainer="raisjn <of.raisjn@gmail.com>"
+license=AGPL3
+
+image=base
+source=(https://github.com/koreader/koreader/releases/download/v2020.08.1/koreader-remarkable-v2020.08.1.zip)
+sha256sums=(447c151e11bf56d71145cecc731dfa2a39eb026fcce4c4d0571cc8e431d5c491)
+
+build() {
+  # Using prebuilt binaries for koreader
+  return
+}
+
+package() {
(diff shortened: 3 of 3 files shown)


Eeems:
> 
> 
> > what do you think of binary releases?
> 
> > i also started going down source release path, but koreader build process actually needs a git repo because it uses submodules to manage dependencies.

Low priority for now, until we need to push out an EBF, but I've opened #48

package/koreader/koreader.draft
```diff
@@ -0,0 +1,3 @@
+name=KOReader
+desc=ebook reader
+call=/opt/bin/koreader
```
Eeems, line 4:
```suggestion
call=/opt/bin/koreader
term=killall -9 luajit
```

Eeems (changes requested):
The draft configuration should have `term=` code to properly kill KOReader.

Eeems (approved):

package/koreader/koreader
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
```
LinusCDE, line 1: A bit nitpicky, but using bash for running another command seems overkill to me.

package/koreader/package
```diff
@@ -0,0 +1,26 @@
...
+sha256sums=(447c151e11bf56d71145cecc731dfa2a39eb026fcce4c4d0571cc8e431d5c491)
+
+build() {
+  # Using prebuilt binaries for koreader
+  return
+}
+
+package() {
+  install -d "$pkgdir"/opt/koreader
+  cp -R "$srcdir"/* "$pkgdir"/opt/koreader/
```
LinusCDE, line 22: For some reason, `$srcdir` also contains the zip (`koreader-remarkable-v2020.08.1.zip`) again (and gets packaged to `/opt/koreader`) which makes the package basically twice as large for no reason.

package/koreader/koreader.draft
LinusCDE: Running this change on oxide showed a me for the first 3 times a crash screen for a short while and made koreader restart before I could actually quit by holding the left button.

LinusCDE: Image of crash screen (only shown for about 1 second):
20200923_004209_lq

(I also verified that my `/opt/etc/draft/koreader.draft` actually contains the `term=` entry from above.)

package/koreader/package
LinusCDE: Also should we really put standalone apps directly into `/opt` as if we didn't have another rootfs there? Alternativly I would suggest `/opt/lib` for that as many linux packages also put their additional data there when `(/opt)/usr/share/` is not fitting for all their content.

LinusCDE (changes requested):
Installed and test-ran it. Looks great! If found a few issues though. We also may want to add an icon.

package/koreader/koreader.draft
Eeems: Oh right, the script needs to be killed as well. Forgot to test with just draft. There is a variable you can pass into KOReader to keep it from restarting itself, but that wont play nice with oxide.

package/koreader/package
Eeems: Maybe `/opt/usr/local/koreader` makes more sense?

raisjn: where would you recommend putting standalone apps? 

good find on the .zip, should definitely be removed

Eeems: Under Archlinux's rules you'd put them in `/opt/{name}` but since we are already here it might make more sense to use `/usr/local/{name}`. That said, reviewing the guidelines here https://wiki.archlinux.org/index.php/Arch_package_guidelines#Directories leads me to believe that doesn't make sense either.

LinusCDE: `/opt/usr/local/koreader` is a really good idea, too. Though if I'm not mistaken, `/usr/local/bin/` usually means "Not managed by the system package manager". Looking at my system that doesn't seem to hold much merit though.

So I think it doesn't really matter for us. I have no problem with `/opt/usr/local/...` for the package as well.

package/koreader/koreader
LinusCDE: Just looked up the links and it wouldn't make sense to change this to `/bin/sh` since they default that to bash anyway.

package/koreader/package
Eeems: we could do `/opt/usr/local/apps/koreader` or someting

package/koreader/koreader.draft
```diff
@@ -0,0 +1,3 @@
+name=KOReader
+desc=ebook reader
```
LinusCDE, line 2:
```suggestion
desc=ebook reader
imgFile=koreader
```

package/koreader/package
```diff
@@ -0,0 +1,26 @@
...
+
+  install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
```
LinusCDE, line 24:
```suggestion
  install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
  install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png"
```

LinusCDE (commented):
These changes also add the icon. 600x600 pixels also seems alright and renders good on oxide.

package/koreader/package
```diff
@@ -0,0 +1,27 @@
...
+  install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png"
```
raisjn: this has a trailing double quote

LinusCDE:
```suggestion
  install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
```

LinusCDE (approved):

package/koreader/package
LinusCDE: Missed that. Good call.

Issue #28, opened 2020-09-10
Plato build is not reproducible


matteodelabre:
Plato generates a different package when building locally and when building using GitHub Actions. The offending file seems to be `/opt/etc/draft/icons/plato.png`.

LinusCDE:
My guess is that, the `convert` tool from imagemagick has a slightly different version.  
EDIT: Maybe GitHub Actions uses slightly outdated/different versions to popular docker base images in their image repository.

I suggest to stop creating the icon on the fly and put it either in my fork directly or to the package file here. So far I went with putting icons and oxide files into my source code. With plato I just wanted to avoid many changed files (though the icon wouldn't make a difference at this point) and the icon was basically there anyway.

~Also with what env would I reference the directory that contains `package` from `package()`?~ Found it. It's `$recipedir`.

Should we put metadata like icons and launcher files always into the repo, or do you think that would pollute the repo too much? Then again .service-Files fit into the same category and draft has one in the recipe directory.  
Seeing as everybody else does it, it seems fit to put the metadata files into the source directory as well.

Eeems:
@LinusCDE with #49 being merged, this is solved right?

raisjn:
it still needs someone to build locally using 'make plato' and compare the
package checksum to the one from the repository build and verify they are
the same

Eeems:
> 
> it still needs someone to build locally using 'make plato' and compare the package checksum to the one from the repository build and verify they are the same

So it shouldn't have been merged yet?

raisjn:
merge is fine so we can compare, but this issue stays open until build is
verified as reproducible

raisjn:
```
$ md5sum build/packages/plato/plato_0.9.1-7_armv7-3.2.ipk
5ef8bfd19fa43266b61046e24d6b3cc8  build/packages/plato/plato_0.9.1-7_armv7-3.2.ipk
$ md5sum plato_0.9.1-7_armv7-3.2.ipk
974467f36d8ee08cdb9f164bbf237ec0  plato_0.9.1-7_armv7-3.2.ipk
```

but i don't know how matteo checked originally and figured out that it was the icon. it makes me think that i should be doing a diff across the files instead of just hashing package

Eeems:
You should be able to extract and compare the contents.

matteodelabre:
Indeed, I did it manually by comparing the contents of both archives. I’ll improve the `scripts/check-repo` script to do that automatically.

[...]

PR #49, opened 2020-09-21
Update plato to 0.9.1-7

LinusCDE:
This update should address issue #28 . Please consider reviewing this also in the sense of that issue, too and close that afterwards as well.

In commit [9b18b767](https://github.com/LinusCDE/plato/commit/9b18b76769751e7aa760ae3c31ff8dd263911b8d), the convert command that resulted in inconsistent results was ran once (saved as `oxide-icon.png`) to always have the same hash for that file.

### DIFF
--- a/package/plato/package
+++ b/package/plato/package
@@ -2,15 +2,15 @@
 pkgname=plato
 pkgdesc="Document reader"
 url=https://github.com/LinusCDE/plato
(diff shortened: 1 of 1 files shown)


raisjn:
if the remote package has the same checksum as my local package, i'll consider that bug closed

PR #63, opened 2020-09-24
Make launchers installable alongside each other

matteodelabre:
This PR aims at removing conflicts between the launcher packages and declaring the runtime conflicts that are inherent to them.

* Currently, `draft` and `oxide` cannot be installed alongside each other, because they both provide `/opt/etc/draft/xochitl.draft` and `/opt/etc/draft/icons/xochitl.png`.  This PR removes those files from the two packages and creates a new `xochitl` package to provide those files, on which both `draft` and `oxide` depend.
* The systemd services files for `draft`, `remux` and `oxide` are changed to declare that they are all mutually conflicting.


Eeems:
Should draft+oxide also have xochitl set as a dependency?

matteodelabre:
I’ll add that, thanks!

Eeems:
The conflict originally was due to both provoding a button-capture binary. Since that's no longer the case they no longer need to conflict aside from you can't really use both at the same time.

matteodelabre:
Maybe we should add a `Conflicts=` field to the respective systemd units, since it’s a runtime conflict.

Eeems:
Could you open the PRs?

matteodelabre:
Do we want to do that upstream or here? Changing the systemd unit file upstream would require opening PRs and waiting for them to be merged every time a new launcher is added to Toltec. If we do it in this repo, we already have an overview of existing launchers and can update the `Conflicts=` fields as needed.

Eeems:
> 
> 
> Do we want to do that upstream or here? Changing the systemd unit file upstream would require opening PRs and waiting for them to be merged every time a new launcher is added to Toltec. If we do it in this repo, we already have an overview of existing launchers and can update the `Conflicts=` fields as needed.

I think we want to do both. And since I'm upstream for oxide, I do want to include that in my code :)

matteodelabre:
> > Do we want to do that upstream or here? Changing the systemd unit file upstream would require opening PRs and waiting for them to be merged every time a new launcher is added to Toltec. If we do it in this repo, we already have an overview of existing launchers and can update the `Conflicts=` fields as needed.
> 
> I think we want to do both. And since I'm upstream for oxide, I do want to include that in my code :)

Alright. I’ve added it to this PR and I’ll look into opening upstream PRs tomorrow.

Eeems (dismissed):

raisjn:
what is conflicting about draft and remux? i don't use draft, tbh, so i'm unsure if it launches on middle button press. 

i'm unsure about how to handle the .service files. seems like extra work for me to maintain them in two places, but not a big deal

a side point is the descriptions and naming: i notice you are following some convention (but i'm not sure what) and i think that convention is going to make the listing pages look bare / uninformative. maybe there should be a way to supply long description + notes + changelog to the person installing the software. 

otherwise, everything looks fine to me, but i still haven't tested the packages

matteodelabre:
> what is conflicting about draft and remux? i don't use draft, tbh, so i'm unsure if it launches on middle button press.

You probably don’t want to run both at the same time, as remux will automatically start xochitl on top of draft and both apps will compete for the screen.

> i'm unsure about how to handle the .service files. seems like extra work for me to maintain them in two places, but not a big deal

I think it’s best if we can control the systemd units at package level, to ensure consistency between them.

> a side point is the descriptions and naming: i notice you are following some convention (but i'm not sure what) and i think that convention is going to make the listing pages look bare / uninformative. maybe there should be a way to supply long description + notes + changelog to the person installing the software.

See [this file](https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#pkgname-required) for the current conventions on pkgname and pkgdesc. Currently, I’m mostly reusing the upstream descriptions while removing uninformative words. What information would you add to the descriptions?

Not sure if Opkg supports changelogs. We can probably add long descriptions. Can you please open an issue to discuss this further?

matteodelabre:
I think we should also add `xochitl.service` to the conflicts declaration.

raisjn (approved):
I'm fine with this, one request is: please test remux installation / uninstallation and that you can launch it by holding down middle press. 

for now: it's fine that remux competes with draft, but actually remux should be able to coexist alongside draft / oxide because it's not a "shell" - it is not the first app that launches when starting the remarkable (but oxide / draft are). the main behavior you are saying is problematic is: when remux starts, it restarts xochitl. the proper behavior should be: when remux starts, if xochitl is running, it restarts xochitl.

Eeems:
> 
> for now: it's fine that remux competes with draft, but actually remux should be able to coexist alongside draft / oxide because it's not a "shell" - it is not the first app that launches when starting the remarkable (but oxide / draft are). the main behavior you are saying is problematic is: when remux starts, it restarts xochitl. the proper behavior should be: when remux starts, if xochitl is running, it restarts xochitl.

Right now oxide would catch all long presses on the middle button and remux would never launch. So technically they do conflict right now.

matteodelabre:
> I'm fine with this, one request is: please test remux installation / uninstallation and that you can launch it by holding down middle press.

It works for me.

PR #65, opened 2020-09-25
Move maintainer scripts definition to recipe file

### LINKED ISSUE #60: Rework pre/post scripts
I think we should abstract away a bunch of the complexity of the pre/post scripts. Ideally we should have the following scripts that a package can define:

- preinstall
- postinstall
- preremove
- postremove
- preupgrade
- postupgrade

The tooling that builds a package should automatically create the wrapper script that would be required to determine which of these scripts should be run at what time (see https://github.com/toltec-dev/toltec/issues/55#issuecomment-697986350)

matteodelabre:
Fixes #60.

Recipe files can define 5 new functions, `preinstall()`, `configure()`, `preremove()`, `postremove()`, `preupgrade()` and `postupgrade()`, which are used by the packaging script to create the `preinst`, `postinst`, `prerm` and `postrm` maintainer scripts.

When installing a new package, the following happens:

* The package files are unpacked (but not installed)
* `preinstall` is called
* The package files are installed into place
* `configure` is called

When removing an installed package, the following happens:

* `preremove` is called
* The package files are removed (except configuration files)
* `postremove` is called

When upgrading a package from version A to B, the following happens:

* `preupgrade B` is called from version A
* Old package files are removed (except configuration files)
* `postupgrade B` is called from version A
* New package files are unpacked and installed
* `configure` is called from version B

This PR also updates draft, remux, oxide and plato which were previously using old-style maintainer scripts.


matteodelabre:
Note that `configure()` cannot directly distinguish between a first install and an upgrade situation. This is due to Opkg calling the underlying `postinst` maintainer script without additional context in both cases. But this is not necessary in our current use cases.

Eeems:
> 
> 
> Note that `configure` cannot directly distinguish between a first install and an upgrade situation, but this is due to the fact that Opkg calls the underlying `postinst` maintainer script without additional context in both cases. But this is not necessary in our current use cases.

Any chance we could handle that by doing something in the postupgrade step?

matteodelabre:
I guess `postupgrade()` could create a temporary signal file which gets picked up by `configure()`. But this is quite messy, and I can’t think of a situation where it would be useful to distinguish between install and upgrade in `configure()`. Do you have something specific in mind?

Eeems:
> 
> 
> I guess `postupgrade()` could create a temporary signal file which gets picked up by `configure()`. But this is quite messy, and I can’t think of a situation where it would be useful to distinguish between install and upgrade in `configure()`. Do you have something specific in mind?

Not really, we can make that the package's problem to solve. If they need to hack around with a signal file they can do so.  Ideally if they need to run migrations they will be setup in such a way that it can handle all of the detection itself without that hack anyways.

raisjn:
why is `configure` called and not `postinstall` in the installation process? is that because you use "configure" idea in multiple places?

matteodelabre:
Indeed. `configure()` is used both after installing and after upgrading, so it wouldn’t make sense to call it `postinstall()` IMHO.

matteodelabre:
Some limitations of this implementation:

* During packaging, the recipe is copied verbatim four times into the generated `{pre,post}{inst,rm}` files. This is unnecessary, because each maintainer script only needs the function that it executes. However, removing the unneeded parts of the recipe would require parsing Bash (or at least, robustly detect and remove selected functions from a Bash file), something I didn’t want to go into for now. See [the generated scripts in the draft package for an example](https://github.com/toltec-dev/toltec/suites/1246129157/artifacts/18913566).
* I had to duplicate two functions, `get-conflicts()` and `is-enabled()`, that I use in draft, oxide and remux to detect whether any conflicting units are already enabled and should be disabled before running the respective launcher. It would be better to have some kind of way to share those functions between the packages. Maybe create a common helper package?

Eeems:
> 
> 
> Some limitations of this implementation:
> 
>     * During packaging, the recipe is copied verbatim four times into the generated `{pre,post}{inst,rm}` files. This is unnecessary, because each maintainer script only needs the function that it executes. However, removing the unneeded parts of the recipe would require parsing Bash (or at least, robustly detect and remove selected functions from a Bash file), something I didn’t want to go into for now. See [the generated scripts in the draft package for an example](https://github.com/toltec-dev/toltec/suites/1246129157/artifacts/18913566).
> 
>     * I had to duplicate two functions, `get-conflicts()` and `is-enabled()`, that I use in draft, oxide and remux to detect whether any conflicting units are already enabled and should be disabled before running the respective launcher. It would be better to have some kind of way to share those functions between the packages. Maybe create a common helper package?

Maybe have a `setup` section that you can create the common helpers? That or they just need to copy it into each of them.

matteodelabre:
I’m not sure what you mean @Eeems. Are you referring to one of the two limitations specifically?

Eeems:
> 
> 
> I’m not sure what you mean @Eeems. Are you referring to one of the two limitations specifically?

Right so instead of 
```bash
some-function(){
  ...
}

configure(){
  some-function
}
```
You would do 
```bash
setup(){
  some-function(){
    ...
  }
}

configure(){
  some-function
}
```
This would allow you to only copy in the portions you care about into the resulting scripts.

raisjn:
i think this diff should include the shared helpers file you mention, i don't like the redundant code in the different files.

regarding dead code: i think you could do something like: `source package/$PACKAGE/package; declare -f prerm` to see the definition of `prerm`. you can use this to place just prerm into a specific file, maybe

matteodelabre:
> i think this diff should include the shared helpers file you mention, i don't like the redundant code in the different files.

Yup! Working on this.

> regarding dead code: i think you could do something like: `source package/$PACKAGE/package; declare -f prerm` to see the definition of `prerm`. you can use this to place just prerm into a specific file, maybe

Very cool suggestion, I didn’t think about this way of doing it, thanks!

matteodelabre:
Done. Can one of you review @raisjnn, @Eeems? Thanks!

The final generated maintainer scripts look like this (example with oxide’s postinst):

```sh
#!/usr/bin/env bash
set -e
declare -- pkgname="oxide"
declare -- pkgdesc="Launcher application"
declare -- url="https://github.com/Eeems/oxide"
declare -- pkgver="2.0~beta~3"
declare -- timestamp="2020-09-26T20:53Z"
declare -- section="launchers"
declare -- maintainer="Eeems <eeems@eeems.email>"
declare -- license="MIT"
declare -a depends=([0]="xochitl")
declare -- arch="armv7-3.2"
declare -a source=([0]="https://github.com/Eeems/oxide/releases/download/v2.0-beta/oxide.zip")
declare -a sha256sums=([0]="d3bac44b96c78726fcbfbd8f7a23d3bd4e4ae3f1f794714190100ebe602a8433")
# vim: set ft=sh:

#
# install-lib
#
# Common functions used by the install scripts
#

# Check whether a systemd unit exists and is enabled
#
# Arguments:
#
# $1 - Full name of the systemd unit, e.g. "xochitl.service"
#
# Exit code:
#
# 0 if the unit exists and is enabled, 1 otherwise
is-enabled() {
    # Note that we must invert the condition so that the exit code indicates
    # a failure when the unit exists and is disabled or when the unit does
    # not exist
    ! systemctl list-unit-files "$1" | awk "/$1/{exit \$2 == \"enabled\"}"
}

# Get a list of systemd units with which the given unit conflicts
#
# Arguments:
#
# $1 - Full name of the systemd unit, e.g. "xochitl.service"
#
# Output:
#
# List of conflicting units
get-conflicts() {
    systemctl show "$1" | awk -F'=' '/^Conflicts=/{print $2}'
}
configure () 
{ 
    systemctl daemon-reload;
    if ! is-enabled "$pkgname.service"; then
        echo "";
        echo "Run the following command(s) to use $pkgname as your launcher";
        for conflict in $(get-conflicts "tarnish.service");
        do
            if is-enabled "$conflict"; then
                echo "$ systemctl disable --now ${conflict/.service/}";
            fi;
        done;
        echo "$ systemctl enable --now tarnish";
        echo "";
    fi
}
if [[ $1 = configure ]]; then
    configure
fi
```

package/draft/package
```diff
@@ -43,3 +43,30 @@ package() {
...
+        for conflict in $(get-conflicts "$pkgname.service"); do
+            if is-enabled "$conflict"; then
+                echo "$ systemctl disable --now ${conflict/.service/}"
+            fi
+        done
+
+        echo "$ systemctl enable --now $pkgname"
+        echo ""
+    fi
+}
```
raisjn: can this be pushed into some common library or is it fine to duplicate this across launcher's packages?

package/oxide/package
```diff
@@ -2,7 +2,7 @@
 pkgname=oxide
 pkgdesc="Launcher application"
 url=https://github.com/Eeems/oxide
-pkgver=2.0~beta~2
+pkgver=2.0~beta~3
```
raisjn: while you are in here... i heard that the versioning was wrong

scripts/package-build
```diff
@@ -135,16 +140,71 @@ ctlfile="$ctldir"/control
...
-[[ -f "$recipedir"/postinst ]] && install -D -m 755 "$recipedir"/postinst "$ctldir"/postinst
-[[ -f "$recipedir"/prerm ]] && install -D -m 755 "$recipedir"/prerm "$ctldir"/prerm
-[[ -f "$recipedir"/postrm ]] && install -D -m 755 "$recipedir"/postrm "$ctldir"/postrm
+# Create maintainer scripts
+maintainer-script-header() {
+    cat <<'SCRIPT'
+#!/usr/bin/env bash
+set -e
+SCRIPT
+    dump-metadata
```
raisjn, line 149: cool

raisjn (dismissed):
i really like that the package recipe now has everything self contained! overall looks good, i'm debating whether to build and test the packages locally or wait for it to hit testing and then upgrade and see if i hit any pain, but i'm leaning towards waiting for it to hit testing (out of laziness)

matteodelabre:
Just tested and it looks like remux’s preremove script does not run. I’m investigating this.

raisjn (dismissed):
i haven't looked through the docs in detail, but other stuff lgtm

EDIT: aside from remux stuff, apparently

matteodelabre:
Should be fixed now!

raisjn (approved):

PR #67, opened 2020-09-25
Remove old package versions

### LINKED ISSUE #57: Remove old package versions from published repository

matteodelabre:
Fixes #57. Rationale for doing a non-time-based removal:

> Doing a date-based removal (e.g. keep versions from the last 2 weeks) is not trivial since the mtimes of published files can be arbitrary (they follow the timestamp= field defined in the recipes, to ensure build reproducibility). I think I’ll simply remove all versions except the most recent one, which should not cause problems if people update before upgrade-ing, and even less so since stable will move more slowly than testing.


Eeems:
I'd still feel a little better keeping a couple versions around just in case we push out a package and we missed a bug. This would easily allow people to downgrade back to the safe version.

matteodelabre:
Does it work for you if we keep the current version plus the two previous ones for each package?

Eeems:
> 
> 
> Does it work for you if we keep the current version plus the two previous ones for each package?

Since we can't do time based, that'll do.

raisjn (dismissed):
putting back on matteo's plate for eeems changes

Eeems:
Quick check, is this suppose to be removing all but the last two versions? It looks like the only change was to add `--delete` to the `rsync` call.

Eeems:
@matteodelabre poke

matteodelabre:
> Quick check, is this suppose to be removing all but the last two versions? It looks like the only change was to add `--delete` to the `rsync` call.

Yup. I can’t think of a way to implement the last-two-versions logic with `rsync`, or in Bash more generally. I agree that it would be better to keep more than one version, though.

Eeems:
Fair point. I'm just worried about how this will be for end users if they forget to `opkg update` I guess.

matteodelabre:
If I remember correctly, `opkg` spits out an error saying something along the lines of “did you forget to run `opkg update`” when a package from the local listing is missing from the remote server.

Eeems:
In that case, I think this is likely fine.

Eeems (approved):

PR #68, opened 2020-09-26
Update KOReader to v2020.09

Eeems:


package/koreader/package
```diff
@@ -2,15 +2,15 @@
 pkgname=koreader
 pkgdesc="KOReader"
 url=https://github.com/koreader/koreader
-pkgver=2020.08.01-1
-timestamp=2020-09-19T03:30Z
+pkgver=2020.09-1
+timestamp=2020-09-25T01:42Z
 section=utils
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=AGPL3
```
matteodelabre: The correct [SPDX identifier](https://spdx.org/licenses/) for this license is `AGPL-3.0-or-later`.
```suggestion
license=AGPL-3.0-or-later
```

package/koreader/package
```diff
@@ -2,15 +2,15 @@
 pkgname=koreader
 pkgdesc="KOReader"
 url=https://github.com/koreader/koreader
-pkgver=2020.08.01-1
-timestamp=2020-09-19T03:30Z
+pkgver=2020.09-1
+timestamp=2020-09-25T01:42Z
 section=utils
```
matteodelabre: Maybe this should go into the `readers` section? See [the `section` documentation](https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#section-required).
```suggestion
section=readers
```

package/koreader/package
```diff
@@ -2,15 +2,15 @@
 pkgname=koreader
 pkgdesc="KOReader"
```
matteodelabre: We could make this description more informative. Taking from koreader’s GitHub repo description, it could be something like _“Ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats”_.
```suggestion
pkgdesc="An ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats"
```

matteodelabre (changes requested):

package/koreader/package
```diff
@@ -2,15 +2,15 @@
...
 url=https://github.com/koreader/koreader
-pkgver=2020.08.01-1
-timestamp=2020-09-19T03:30Z
+pkgver=2020.09-1
+timestamp=2020-09-25T01:42Z
```
matteodelabre: This timestamp [should match the upstream release date](https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#timestamp-required), i.e. `2020-09-17T01:52Z` if I’m not mistaken.
```suggestion
timestamp=2020-09-17T01:52Z
```

matteodelabre (changes requested):

raisjn:
I went through and used the 'add suggestion' feature to augment matteo's comments with suggestions (which can be batch committed by eeems)

mea culpa here on a lot of the issues matteo surfaced (i put in wrong timestamp, wrong license, wrong desc, etc), so thanks for fixing and upgrading

package/koreader/package
Eeems: Sure, it wasn't clear since the timestamp is previous to the last one.

package/koreader/package
Eeems: We could use the one that the KOReader repository has:

> An ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats.

matteodelabre (approved):

PR #74, opened 2020-09-27
Switch oxide to binary

Eeems:
Toltec builds are crashing on device.

### DIFF
--- a/package/oxide/package
+++ b/package/oxide/package
@@ -15,6 +15,8 @@ sha256sums=(d3bac44b96c78726fcbfbd8f7a23d3bd4e4ae3f1f794714190100ebe602a8433)
 package() {
     cp -R "$srcdir"/* "$pkgdir"
     rm "$pkgdir"/oxide.zip
+    
+    tree "$pkgdir"
 
     install -D -m 644 -t "$pkgdir"/lib/systemd/system "$pkgdir"/etc/systemd/system/tarnish.service
     rm -r "$pkgdir"/etc/systemd
--- a/package/oxide/package
+++ b/package/oxide/package
@@ -13,6 +13,8 @@ source=(https://github.com/Eeems/oxide/releases/download/v2.0-beta/oxide.zip)
 sha256sums=(d3bac44b96c78726fcbfbd8f7a23d3bd4e4ae3f1f794714190100ebe602a8433)
 
 package() {
+    tree "$srcdir"
+
     cp -ar "$srcdir"/* "$pkgdir"
     rm "$pkgdir"/oxide.zip
     
--- a/package/oxide/package
+++ b/package/oxide/package
@@ -13,11 +13,8 @@ source=(https://github.com/Eeems/oxide/releases/download/v2.0-beta/oxide.zip)
 sha256sums=(d3bac44b96c78726fcbfbd8f7a23d3bd4e4ae3f1f794714190100ebe602a8433)
 
 package() {
-    tree "$srcdir"
+    unzip "$srcdir"/oxide.zip -d "$pkgdir"
 
-    cp -ar "$srcdir"/* "$pkgdir"
-    rm "$pkgdir"/oxide.zip
-    
     tree "$pkgdir"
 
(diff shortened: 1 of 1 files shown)


Eeems:
@matteodelabre FYI I had to switch to a binary release because the build output of the toolchain is crashing.

matteodelabre:
I think this indicates a problem either in the toolchain or in oxide, which I’d rather fix instead of working around.

matteodelabre:
When compiling oxide in debug mode, I get no crash, which makes it harder to see what’s going on. The crash I get on release mode (“terminate called without an active exception”) might indicate that you’re destructing a thread without joining it first.

matteodelabre:
I’m able to reproduce the crash with debug symbols when compiling in release mode with forced debug symbols. Seems that the crash happens in the constructor of AppsAPI. But it could come from another thread, and I’m not able to switch threads with gdb.

raisjn (approved):
> I think this indicates a problem either in the toolchain or in oxide, which I’d rather fix instead of working around.

agree, but there are people who have broken builds, so accepting for now while debugging happens

Eeems:
> I’m able to reproduce the crash with debug symbols when compiling in release mode with forced debug symbols. Seems that the crash happens in the constructor of AppsAPI. But it could come from another thread, and I’m not able to switch threads with gdb.

Could you open an issue on oxide about this? I'm surprised that this doesn't happen with to rM toolchain.

package/oxide/package
```diff
@@ -2,24 +2,20 @@
 pkgname=oxide
 pkgdesc="Launcher application"
 url=https://github.com/Eeems/oxide
-pkgver=2.0~beta~1
+pkgver=2.0~beta~2
```
matteodelabre, line 5: The `pkgrel` part of the package version should be separated from the rest using a dash, not a tilde, i.e. `2.0~beta-2` instead of `2.0~beta~2`. Otherwise, if a version is released that drops the `~beta` suffix, we get `2.0~2`, which is actually sorted lower than `2.0~beta~2`. You can use `opkg compare-versions` to test if a version is sorted lower or higher than another.

package/oxide/package
```diff
@@ -2,24 +2,20 @@
...
-    make release
-}
+source=(https://github.com/Eeems/oxide/releases/download/v2.0-beta/oxide.zip)
+sha256sums=(d3bac44b96c78726fcbfbd8f7a23d3bd4e4ae3f1f794714190100ebe602a8433)
 
 package() {
-    cp -ar "$srcdir"/release/* "$pkgdir"
+    unzip "$srcdir"/oxide.zip -d "$pkgdir"
+
+    tree "$pkgdir"
```
matteodelabre, line 18: Why is this necessary? `tree` is already called [in the packaging script](https://github.com/toltec-dev/toltec/blob/249e558582c88ca66194d8acf53738500c2266c6/scripts/package-build#L146).

matteodelabre (commented):
Some little things I didn’t catch before the PR was merged.

matteodelabre:
> > I’m able to reproduce the crash with debug symbols when compiling in release mode with forced debug symbols. Seems that the crash happens in the constructor of AppsAPI. But it could come from another thread, and I’m not able to switch threads with gdb.
> 
> Could you open an issue on oxide about this? I'm surprised that this doesn't happen with to rM toolchain.

I’m surprised too. The gcc version in the reMarkable toolchain is 7.3.0, and the gcc in the toltec images is 8.3.0, which might explain the difference. I’ll open an issue.

package/oxide/package
Eeems: If you look at the history you'll see why it happened. The code that handles extracting the zip automatically does some funky things about assuming that you don't care about the top level directory in the zip file.

matteodelabre: I see. I tried to maintain consistency with Arch’s makepkg while implementing it, so the behavior should be roughly the same. Do you have any suggestion on how to improve the extracting part of the building script?

Eeems: I'd have to look at what you are doing, but at the very least we should document the behaviour.
Ideally we should only extract assuming the zip contains a single directory if that's actually the case.

PR #78, opened 2020-09-27
package: update remux, nao, simple to latest

raisjn:
* remux: use process groups for halting programs
* nao: add exit button, add manage package item from main menu
* simple: support for displaying newlines properly

### DIFF
--- a/package/nao/package
+++ b/package/nao/package
@@ -2,8 +2,8 @@
 pkgname=nao
 pkgdesc="nao package manager"
 url=https://github.com/rmkit-dev/rmkit
-pkgver=0.0.1-14
-timestamp=2020-09-21T01:41+00:00
+pkgver=0.0.1-15
+timestamp=2020-09-27T01:41+00:00
 section=utils
 maintainer="raisjn <of.raisjn@gmail.com>"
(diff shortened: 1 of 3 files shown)


package/nao/package
```diff
@@ -2,17 +2,17 @@
...
 url=https://github.com/rmkit-dev/rmkit
-pkgver=0.0.1-14
-timestamp=2020-09-21T01:41+00:00
+pkgver=0.0.1-15
+timestamp=2020-09-27T01:41+00:00
 section=utils
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
 depends=simple
 conflicts=rmkit
```
matteodelabre, line 11: Do we still need this `conflicts=` declaration since `rmkit` does not exist anymore and has never been moved to `stable`? Ditto for `simple`.

matteodelabre (approved):
Looks good to me!

package/nao/package
raisjn: people may have installed rmkit from testing, as most people are using testing (since nothing is in stable). will remove in some time from now (maybe a month or two)

PR #80, opened 2020-09-28
Add draft configuration for calculator

### LINKED ISSUE #51: Add draft configuration file and icon to calculator

matteodelabre:
Fixes #51.


Eeems:
You are including both svg and png?

matteodelabre:
Would it be better to generate the PNG on the fly? That would require adding `inkscape` to the build environment, but it seems feasible.

Eeems:
> 
> 
> Would it be better to generate the PNG on the fly? That would require adding `inkscape` to the build environment, but it seems feasible.

Probably, otherwise it's not clear what the svg is being used for.

matteodelabre:
That would be so that others can iterate on the icon if needed.

LinusCDE:
Please also ensure though that builds are repeatable. Converting svg to png with inkscape also seems overkill.

They have indeed some custom metadata in their svg files which can make converting/rendering faulty with other software. But then this is akin to a IDEA-Specific project file and one should usually export the plain svg file in my opinion, which should convert fine with any svg to pdf converting utility.

matteodelabre:
> They have indeed some custom metadata in their svg files which can make converting/rendering faulty with other software. But then this is akin to a IDEA-Specific project file and one should usually export the plain svg file in my opinion, which should convert fine with any svg to pdf converting utility.

Agreed. This PR contains the plain SVG file.

.github/workflows/testing.yml
```diff
@@ -12,7 +12,7 @@ jobs:
               uses: actions/checkout@v2
             - name: Install missing tools
               run: |
-                sudo apt-get install bsdtar tree
+                sudo apt-get install bsdtar inkscape tree
```
Eeems: Do we still need this change?

matteodelabre: Good catch!

Eeems (approved):

PR #86, opened 2020-09-29
Handle recipe metadata fields more safely

### LINKED ISSUE #64: Replace recipe metadata variables by associative array
Currently, recipes declare a handful of metadata variables such as `pkgname`, `url` and `source` which have more or less generic names. This causes two problems:

* We must take care of not overwriting them inside the build scripts.
* `shellcheck` is not able to automatically detect that those variables are defined, unless we declare them manually.

I propose placing all these variables in a single Bash associative array to solve the above two problems. Here’s an example of how that could look like for plato.

_Before:_

```
pkgname=plato
pkgdesc="Document reader"
url=https://github.com/LinusCDE/plato
pkgver=0.9.1-8
timestamp=2020-09-21T18:40Z
section=readers
maintainer="Linus K. <linus@cosmos-ink.net>"
license=AGPL-3.0-or-later

image=rust:v1.1
source=(https://github.com/LinusCDE/plato/archive/9b18b76769751e7aa760ae3c31ff8dd263911b8d.zip)
sha256sums=(478333e0f32096471950ef3f1f76d7cc42f6802dc417227518b6ffc5593b4457)
```

_After:_

```
declare -A metadata=(
    [name]=plato
    [description]="Document reader"
    [url]=https://github.com/LinusCDE/plato
    [version]=0.9.1-8
    [timestamp]=2020-09-21T18:40Z
    [section]=readers
    [maintainer]="Linus K. <linus@cosmos-ink.net>"
    [license]=AGPL-3.0-or-later

    [image]=rust:v1.1
    [sources]=https://github.com/LinusCDE/plato/archive/9b18b76769751e7aa760ae3c31ff8dd263911b8d.zip
    [sha256sums]=478333e0f32096471950ef3f1f76d7cc42f6802dc417227518b6ffc5593b4457
)
```

Then we can:

* Retrieve a piece of metadata: `${metadata[name]}`
* Check if an attribute is defined: `[[ -v metadata[nonexisting] ]]`
* Define a default value for an attribute: `[[ ! -v metadata[nonexisting] ]] && metadata[nonexisting]="default value"`.

matteodelabre:
Fixes #64.

* Add basic typechecking to the `load-recipe` function.
* Make sure fields from a recipe do not leak to others when building the full repo.
* Mark fields as read-only so they cannot be accidentally overwritten in the script.

### DIFF
--- a/docs/package.md
+++ b/docs/package.md
@@ -16,24 +16,68 @@ For consistency, those fields must be declared in the order they are described b
 
 > **Note:** The field names and semantics is inspired both by the [Debian control file format](https://www.debian.org/doc/debian-policy/ch-controlfields.html) and the [Arch Linux PKGBUILD format](https://wiki.archlinux.org/index.php/PKGBUILD).
 
-#### `pkgname` (required)
+#### `pkgname`
+
+<table>
+    <tr>
+        <th>Required?</th>
+        <td>Yes</th>
+    </tr>
+    <tr>
+        <th>Type</th>
+        <td>String</td>
+    </tr>
+</table>
 
 Name of the built package.
 Must only contain ASCII lowercase letters, digits and dashes.
 Should match the upstream name as closely as possible.
 
-#### `pkgdesc` (required)
+#### `pkgdesc`
+
+<table>
+    <tr>
+        <th>Required?</th>
+        <td>Yes</th>
+    </tr>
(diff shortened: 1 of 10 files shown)


raisjn (approved):
nice! i don't really have anything to add other than code looks great

PR #87, opened 2020-09-30
Update koreader to v2020.09.29-1

raisjn:
update to koreader nightly build which has the multi-task friendly patches for disabling input grabbing and resolution changing. 

this removes: 'term=killall luajit -9' because KOReader can now be paused properly.


Eeems:
I'm not convinced that `term=killall luajit -9` can be removed, since that's not for pausing, but for exiting the application. My expectation is that you would still have a problem when killing the parent script where the `luajit` instances wouldn't exit.

raisjn:
removed the removal of `term` in last push

Eeems (approved):
lgtm, I haven't tested, but I'm assuming you will?

raisjn:
yes, i've tested the build and it respects the two environment variable changes around setting depth and grabbing inputs

PR #88, opened 2020-09-30
Add web build scripts

raisjn:
first pass at making webpage for toltec index page and package show page

i went with bash since i thought i might as well stick with the theme. can see a preview at https://opkg.rmkit.dev/testing/


raisjn:
an idea is: a package could supply a file in their directory that gets dumped into the webpage for the package. the file would need to be sanitized (or examined during pull request time)

raisjn:
misc things:

* should have a link to the github for toltec on the listing page
* should have links to project homepages on each package's page (if applicable)
* ???

raisjn:
I think a changelog per package is also interesting (for people who want to populate it). i believe debian has this concept

Eeems:
> 
> 
> I think a changelog per package is also interesting (for people who want to populate it). i believe debian has this concept

I'd rather just link to the project's changelog.

scripts/repo-build-web
```diff
@@ -0,0 +1,103 @@
+
```
matteodelabre: There’s an extra line before the shebang here.

matteodelabre:
That’s a good start @raisjnn, thanks! :-)

matteodelabre:
I feel like `repo-build-web` really belongs to (or at least should be called by) the `repo-build` script, since the latter already creates the machine-readable `Packages` and `Packages.gz` indexes. Also, I’m not sure about the `repo-build-web` naming, since it breaks the `noun-verb` or `noun-lib` convention of other scripts in this folder.

raisjn:
if you have idea for naming, please feel free to fix it (thank you!) - I have no preference and it took me too long to come up with the current names (noun verb is awkward for people who speak SVO language, i think)

some feedback I'm looking for currently is around what is missing from the web site and what people would like to see added

matteodelabre:
I'm not sure we actually need a separate page for each package. The info is already on the main page, and we can turn the link on each package name into a link to the homepage for that package.

Maybe you could also add a link to the parent directory, i.e. Toltec's home page?

raisjn:
I'm going to go through and add websites to each package.

currently the description is shown nowhere, I'm thinking about where to put that too

matteodelabre:
The `url` field is already meant to be the homepage of each package, so maybe you could use that?

raisjn:
works for me, I'll use that field

raisjn:
i'm wondering if we should split out the table into one table per category of application. 

EDIT: maybe leave that for another diff

matteodelabre:
Yeah, that sounds like a good idea, but maybe we should indeed move forward with a first version.

package/appmarkable/package
```diff
@@ -1,6 +1,6 @@
 # vim: set ft=sh:
 pkgname=appmarkable
-pkgdesc="Turn your program into a very simple app for draft similar"
+pkgdesc="Turn your program into a very simple app for draft"
 url=https://github.com/LinusCDE/appmarkable
```
matteodelabre: For consistency with the other changes to `url=` fields.

```suggestion
url="https://github.com/LinusCDE/appmarkable"
```

package/mines/package
```diff
@@ -1,6 +1,6 @@
 # vim: set ft=sh:
 pkgname=minesweeper
-pkgdesc="mine detection game"
+pkgdesc="Mine detection game"
 url=https://github.com/rmkit-dev/rmkit
```
matteodelabre:
```suggestion
url="https://rmkit.dev/apps/minesweeper"
```

scripts/repo-build
```diff
@@ -93,3 +93,7 @@ touch --no-dereference --date="@$lastcommitdate" \
     "$repodir"/Packages.stamps
 
 section "Done. Result is in $repodir"
```
matteodelabre: Perhaps move the “Done…” message to the final line of the build script?

scripts/repo-build
```diff
@@ -93,3 +93,7 @@ touch --no-dereference --date="@$lastcommitdate" \
...
+
+section "Making packages web listing"
+echo "${BASH_SOURCE%/*}"/repo-build-web "$recipesdir" "$repodir"
+"${BASH_SOURCE%/*}"/repo-build-web "$recipesdir" "$repodir"
```
matteodelabre, line 99: I’ve been changing the other occurrences of `${BASH_SOURCE%/*}/` to simply `scripts/` in other scripts lately, because I think it looks confusing and because we only run the scripts from the repo root anyway. If you agree with this, could you please change those two lines to have `scripts/` instead of `${BASH_SOURCE%/*}/`? Otherwise, we can discuss about it and maybe change it back in the other places.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+Build new or updated packages defined in RECIPESDIR and create a package index.
+The WORKDIR must be a non-existent directory in which the builds are performed.
+If REMOTEREPO is omitted 'https://toltec.delab.re/testing' is used.
+
+Options:
+
+    -h              Show this help message.
+    -f              Force building all the packages, even those for which
+                    an already built version exists on the remote."
```
matteodelabre, line 16: This help message probably needs to be updated.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+
+helpflag=
+
+while getopts hf name; do
```
matteodelabre: Since there is no `-f` flag for this script.

```suggestion
while getopts h name; do
```

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+if [[ $# -eq 1 ]]; then
+    error "Missing WORKDIR argument. Use the -h flag for more information."
+fi
+
+if [[ $# -gt 2 ]]; then
+    error "Extraneous arguments. Use the -h flag for more information."
+fi
+
+recipesdir="$1"
+workdir="$2"
```
matteodelabre, line 45: For consistency with the `repo-build` script and to avoid confusion with the directory that is used for building the individual packages.

```suggestion
repodir="$2"
```

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+
+pkgsworkdir="$workdir"/
```
matteodelabre, line 47: Isn’t this variable equivalent to `workdir` (or `repodir` by my previous comment)?

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+mkdir -p "$pkgsworkdir"
+
+cat > "${pkgsworkdir}/index.html" << WEBS
+<html>
```
matteodelabre, line 51: To make it HTML5-compliant.

```suggestion
<!doctype html>
<html>
```

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+<head>
+<style>
+  body {
```
matteodelabre, line 55: In the interest of consistency, I think the HTML should also be 4-spaces-indented.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+        load-recipe "$recipedir"
+
+        section "Adding table entry for $pkgname $pkgver"
+
+        pkgworkdir="$pkgsworkdir"/"$pkgname"
+        if [ -z "${url}" ]; then
+          href="${url}";
+        else
+          href="<a target='_new' href='${url}'>${pkgname}</a>"
+        fi
```
matteodelabre, line 107: You should use 4 spaces for indentation in the two branches of this condition.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+
+# Build each package or get it from the remote server
+for recipedir in "$recipesdir"/*; do
+    (
+        load-recipe "$recipedir"
+
+        section "Adding table entry for $pkgname $pkgver"
+
+        pkgworkdir="$pkgsworkdir"/"$pkgname"
+        if [ -z "${url}" ]; then
```
matteodelabre: For consistency with the other conditions, which all use the double-brackets syntax, I think this should be changed to `[[ -z $url ]]`.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+          href="${url}";
+        else
+          href="<a target='_new' href='${url}'>${pkgname}</a>"
+        fi
+
+cat >> "${workdir}/index.html" <<WEBS
+<tr>
+  <td>
+    ${href}
+  </td>
```
matteodelabre: Why is this tag on three lines while the others are one a single line?

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+</head>
+
+<body>
+
+<a href="../">Back to Repository Home Page</a>
+
+<h1>Toltec Package Listing</h1>
+<table>
+
+<thead> <th>Name</th> <th>Description</th> <th>Version</th> <th>Category</th> <th>License</th> </thead>
```
matteodelabre: I think you need to wrap the `<th>`s in a `<tr>`.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+
+# Build packages index
+section "Making packages index", "${pkgsworkdir}/index.html"
+
+# Build each package or get it from the remote server
+for recipedir in "$recipesdir"/*; do
+    (
+        load-recipe "$recipedir"
+
+        section "Adding table entry for $pkgname $pkgver"
```
matteodelabre: `section` is used for sectioning messages in other scripts, but here I think this belongs more as a normal-level message since nothing is echo-ed while generating the table entry for a specific package. I would therefore suggest to turn it into a `status` call.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+<a href="../">Back to Repository Home Page</a>
+
+<h1>Toltec Package Listing</h1>
+<table>
+
+<thead> <th>Name</th> <th>Description</th> <th>Version</th> <th>Category</th> <th>License</th> </thead>
+WEBS
+
+# Build packages index
+section "Making packages index", "${pkgsworkdir}/index.html"
```
matteodelabre: There are two arguments to `section` here, and the second is ignored.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+
+cat >> "${workdir}/index.html" <<WEBS
+<tr>
+  <td>
+    ${href}
+  </td>
+  <td>${pkgdesc}</td>
+  <td>${pkgver}</td>
+  <td>${section}</td>
+  <td>${license}</td>
```
matteodelabre: Since the `license` fields follow the SPDX convention, we can direct the user to each license simply by linking to `https://spdx.org/licenses/$license.html`.

scripts/repo-build-web
```diff
@@ -0,0 +1,128 @@
...
+
+recipesdir="$1"
+workdir="$2"
+
+pkgsworkdir="$workdir"/
+mkdir -p "$pkgsworkdir"
+
+cat > "${pkgsworkdir}/index.html" << WEBS
+<html>
+<head>
```
matteodelabre, line 52: To avoid any encoding problems in the future.

```suggestion
<head>
<meta charset="utf-8">
```

matteodelabre (dismissed):
Thanks @raisjnn! Here are my comments.

raisjn:
@matteodelabre do you know the "suggest change" feature?

scripts/repo-build
matteodelabre, line 99: I’ve been changing the other occurrences of `${BASH_SOURCE%/*}/` to simply `scripts/` in other scripts lately, because I think it looks confusing and because we only run the scripts from the repo root anyway. Do you agree with this? Otherwise, we can discuss about it and maybe change it back in the other places.

```suggestion
echo scripts/repo-build-web "$recipesdir" "$repodir"
scripts/repo-build-web "$recipesdir" "$repodir"
```

scripts/repo-build-web
matteodelabre, line 16: This help message probably needs to be updated.

matteodelabre:
> @matteodelabre do you know the "suggest change" feature?

Now I do :-) Thanks!

scripts/repo-build-web
raisjn: meh, i don't think html5 matters much here

matteodelabre: Well, it should make it less likely to hit weird bugs related to compatibility with old versions of the standard or old implementations, plus it’s not much effort.

scripts/repo-build-web
raisjn: "a foolish consistency is the hobgoblin of little minds"

package/remux/package
```diff
@@ -1,9 +1,9 @@
 # vim: set ft=sh:
 pkgname=remux
 pkgdesc="App launcher that supports multi-tasking applications"
-url=https://github.com/rmkit-dev/rmkit
+url="https://rmkit.dev/apps/remux"
 pkgver=0.0.1-18
-timestamp=2020-09-29T01:41+00:00
+timestamp=2020-09-27T01:41+00:00
```
raisjn: hmmm

matteodelabre:
If I’m not mistaken, you did not address part of my comments.

matteodelabre (approved):

PR #89, opened 2020-09-30
Update remux to latest (2020.09.29)

raisjn:
* read xochitl command from systemd file
* use rgb565 when drawing suspend screen (better image quality)
* don't use `term` command to pause apps any more (KOReader integration)


matteodelabre (approved):
Looks good to me.

PR #90, opened 2020-09-30
Add recrossable

matteodelabre:
Still need to figure out why the app does not receive pen events.

### DIFF
--- a/package/recrossable/package
+++ b/package/recrossable/package
@@ -1,6 +1,6 @@
 # vim: set ft=sh:
 pkgname=recrossable
-pkgdesc="Crossword game with simplistic handwriting recognition and automatic generation of crosswords"
+pkgdesc="Solve crossword puzzles"
 url=https://github.com/sandsmark/recrossable
 pkgver=0.0.0-1
 timestamp=2019-01-08T17:41Z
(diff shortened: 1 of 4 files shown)


Eeems:
It's probably not handling them. Here is what I have to do in oxide to handle most pen events: https://github.com/Eeems/oxide/blob/master/applications/launcher/eventfilter.cpp#L105-L142

matteodelabre:
But surely it is supposed to handle them, because [it ignores touch events](https://github.com/sandsmark/recrossable/blob/master/drawablecell.cpp#L54) and there is no way to interact with the app otherwise. Or am I missing something?

Eeems:
Looks like there is stuff here: https://github.com/sandsmark/recrossable/blob/master/main.cpp#L23

matteodelabre:
It works, but I must be missing something since that workaround is targeted to Qt ⩽ 5.6 and we’re on Qt 5.11.

Eeems:
> 
> 
> It works, but I must be missing something since that workaround is targeted to Qt ⩽ 5.6 and we’re on Qt 5.11.

He likely hasn't tested on the new toolchain and realized that it also needs the workaround.

matteodelabre:
Plausible indeed since the previous Qt version in the toolchain was 5.6.

matteodelabre:
Reported to upstream in sandsmark/recrossable/issues/3.

Eeems (approved):
lgtm

PR #92, opened 2020-10-01
use HEAD request instead of GET when checking file existence on remote server

raisjn:
what do you think? (it could probably be better in terms of cleanliness)

### DIFF
--- a/scripts/package-lib
+++ b/scripts/package-lib
@@ -119,6 +119,18 @@ rsecurl() {
     rcurl --proto '=https' "$@"
 }
 
+# checks if a file exists by issuing HEAD request
+# this avoids downloading the request body
+curlcheck() {
+    response=$(curl -n --silent -X HEAD "$@" --write-out "\n%{http_code}\n" -D /dev/null)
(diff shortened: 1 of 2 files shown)


Eeems:
lgtm

matteodelabre:
I'm always for saving network usage but with this change the `Packages`, `Packages.gz` and `Packages.stamps` files generated at the end of the `repo-build` script will only contain the newly updated packages. When the GitHub actions are run, those incomplete indexes will overwrite the previous ones, and therefore users of Opkg will only have access to a subset of the packages.

raisjn:
there should be some way of using it for local building so we don't re
download files when running `make repo'

raisjn:
alternatively, the generation of packages.gz can happen on remote server

matteodelabre:
What about changing the `repo-build` script and `make repo` target so that they do not remove the build working directory every time they are run? Existing packages from the previous runs could be kept in place without doing anything, and the download would only happen on the first run.

raisjn:
yes, that's a good idea. will save my bandwidth but not the ci servers, but I'm fine with that

PR #93, opened 2020-10-02
Reuse local repo artifacts in repo-build

### LINKED ISSUE #91: Use HEAD request instead of GET when checking if package exists on remote server
this will save many megabytes

matteodelabre:
Fixes #91, closes #92.

* When running the repo-build script with an existing `build` directory, reuse the previous artifacts from that directory if they are still valid (i.e. matching versions).
* Rename the --force flag in the repo-build script to --local to avoid confusion since the flag does not force rebuilding the local artifacts, but only forces to build them locally when applicable.
* Add a new `clean` target to the Makefile.
* Make packages build workdirs unique per package version, so that they don’t conflict with existing previous versions, should the build directory be kept.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,11 +7,13 @@ Available targets:
 
     repo            Build the repository and reuse archives from the remote
                     repository for existing package versions.
-    repo-force      Build the repository without reusing existing archives.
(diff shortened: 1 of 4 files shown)


raisjn (approved):
lgtm

PR #96, opened 2020-10-04
Add and enforce a style guide

### LINKED ISSUE #95: Define and automatically enforce a code style
* [x] Define a common code style for the project (e.g. as a shfmt config).
* [x] Add a `make lint` target for checking code style issues and optionally automatically fixing them (e.g. using shfmt).
* [x] Add a step in the `pr` workflow for automatically checking code style.
* [x] Create a contributing guide in which this code style is mentioned.

matteodelabre:
Fixes #95.

* Define an automatically-enforceable style guide by adding an .editorconfig file (configures shfmt).
* Replace Vim modelines by Bash shebangs so that shfmt can detect all shell scripts as such, and update docs/package.md accordingly.
* Add `make lint` target to run `shellcheck` and report errors.
* Add `make format` target to run `shfmt` and report errors.
* Add `make format-fix` target to run `shfmt` and automatically fix errors.
* Fix existing formatting & lint errors.
* Check format and lint conformity in the `pr` workflow.
* Mention the style guide in docs/contributing.md


package/.shellcheckrc
```diff
@@ -0,0 +1,15 @@
...
+# Configuration for Shellcheck in package recipes
+# See <https://github.com/koalaman/shellcheck/blob/master/shellcheck.1.md#rc-files>
+
+# Do not warn about unguarded `cd`s since recipes are always executed
+# with the -e flag set
+disable=SC2164
+
+# Do not warn about unused variables since metadata fields are expected to
+# not necessarily be used in the recipes
+disable=SC2034
```
raisjn, line 11: is there a way to add just the variables to a list of ones to ignore? this seems like a large benefit of shellcheck to turn off

package/.shellcheckrc
```diff
@@ -0,0 +1,15 @@
...
+
+# Do not warn about undeclared variables since recipes may use build variables
+# such as `srcdir` or `pkgdir`
+disable=SC2154
```
raisjn, line 15: same as above

scripts/repo-build-web
```diff
@@ -30,98 +29,105 @@ if [[ -n $helpflag ]]; then
...
-</table>
-</body>
+cat >> "$repodir/index.html" << WEBS
+        </table>
+    </body>
 </html>
 WEBS
-section "Done. Result is in ${pkgsworkdir}"
+
+status "Done. Result is in $repodir/index.html"
```
raisjn, line 133: most of these changes (that are unrelated to style) should be in a separate PR, but i'll allow it ;)

docs/contributing.md
```diff
@@ -0,0 +1,52 @@
+## Contributing to Toltec
+
+Thank you for taking time to contribute to Toltec!
+We welcome contributions from anyone, regarding for example [reporting bugs](#reporting-a-bug), [requesting](#request-a-package) or [adding new packages](#adding-a-new-package), [upgrading existing ones](#upgrading-a-package), [improving the docs](#improving-the-documentation), or other topics.
+
+To make a request or report a bug, you simply need to [open a new issue](../../../issues/new/choose).
+
+To directly propose changes, the basic procedure is to fork this repository, make the desired changes in your newly created local copy, and open a pull request.
+When proposing changes, please make sure that you follow the [style guide](#style-guide).
+After you submit your pull request, a maintainer will take time to review your changes and merge them into the repository if they fit.
```
raisjn:
```suggestion
After you submit your pull request, a maintainer will take time to review your changes, request modifications and then merge your changes into the repository if they fit.
```

I feel like "fit" is ending the sentence too early - what are the changes supposed to fit?

raisjn:
conceptually, style is something that will be non-destructive to fix (since style is solely about formatting) and should not prevent a merge. instead, the fixes should be integrated into PR process and auto applied or run as a bot on some regular basis. 

it does make sense for lint to stop a PR, though (and fixes can be auto suggested / applied)

package/.shellcheckrc
matteodelabre: Unfortunately there is no direct way of doing that yet, as far as I know. Issue koalaman/shellcheck#2020 proposes to add an `is_used` directive that would do exactly that. Two workarounds come to mind:

1. `export` all the metadata fields in recipes. IMO, this would add too much noise.
2. Instead of passing each recipe as-is to shellcheck, append dummy lines that use the fields in question, and pass the result to shellcheck. To do that we’ll need to write a separate script that applies this workaround to each file as needed and then calls shellcheck.

package/.shellcheckrc
matteodelabre: Issue koalaman/shellcheck#2020 would also help here since it proposes the dual `is_set` directive. Otherwise, we can apply a workaround similar to my comment above, but prepending `declare` lines instead.

scripts/repo-build-web
matteodelabre: Thanks :-)

docs/contributing.md
matteodelabre: I was not sure how to express concisely that the reviewer can also choose to reject the changes if they feel that they don’t fit with the project goals.

matteodelabre:
> conceptually, style is something that will be non-destructive to fix (since style is solely about formatting) and should not prevent a merge. instead, the fixes should be integrated into PR process and auto applied or run as a bot on some regular basis.

Agreed. We ought to use [lint-action](https://github.com/marketplace/actions/lint-action) in the long run, but we first need to wait until it supports shfmt/create a PR to add support for shfmt (issue wearerequired/lint-action#59).

raisjn (approved):
let's iterate

raisjn:
why is a check still running?

matteodelabre:
It looks like this is because I renamed the previous _“Check that a pull request correctly builds”_ check to _“Check that it builds without error”_ in this PR, and therefore I need to update the branch protection settings.

PR #97, opened 2020-10-04
Add chessmarkable 0.3.3-1

LinusCDE:


raisjn:
consider not taking the toplevel name `chess` and using a variant of it? (in case another chess app comes along) 

(I am very excited about this app, can't wait t install it)

LinusCDE:
I also thought about it a long time. I couldn't come up with a decent play on the word. I initially had called it `remarkable-chess` (some of the early release binaries are still called that). But I ended up renaming it in chess since the package shouldn't start with `remarkable-` (there was some discussion if I'm not mistaken).

Do you have any other idea for it? Btw it is available on my [test repo](https://rmtestrepo.cosmos-ink.net/) since I used that to QA my packages anyway.

matteodelabre:
Just tested the package and it works great, thanks! I don’t have strong opinions on the name.

raisjn:
You really can't think of any other name than chess? it seems unfair to other packages to pick the most generic name (imo, no package should be called "chess")

lChess, LinChess, LinusChess, rustChess, reChess, rmChess, Chesslet, ChessMarkable...

LinusCDE:
I'm also really torn on it. My initial guess is that it will probably be fairly unlikely that we end up with two chess versions and my only other idea for a name `ress` didn't sound good or like chess at all imo. Going with `rmChess` is also getting boring for me and the casing just is messy.

(I didn't see the previous reply at this point)

---

> You really can't think of any other name than chess? it seems unfair to other packages to pick the most generic name (imo, no package should be called "chess")
> 
> lChess, LinChess, LinusChess, rustChess, reChess, rmChess, Chesslet, ChessMarkable...

Agreed. ChessMarkable sounds nice. I'll use that if you don't mind.

raisjn:
go for it!

re: two chess apps: you'd be surprised

LinusCDE:
I already got some suspicion based on your response. :laughing:

raisjn:
where did the chess icons come from / what is the license on them?

LinusCDE:
> where did the chess icons come from / what is the license on them?

I took them from pixabay (orig image and url [here](https://github.com/LinusCDE/chessmarkable/tree/master/res/sources)). They don't seem to care what you do with those images. Crediting them is even a bonus only. The disadvantage is that they are fairly low res and the transparency is a bit messed up. If I have time I'll consider remaking my own ones from scratch (hopefully). For now they suffice.

package/chessmarkable/package
```diff
@@ -0,0 +1,27 @@
+# vim: set ft=sh:
```
raisjn: should be the env bash line

package/chessmarkable/package
```diff
@@ -0,0 +1,27 @@
...
+maintainer="Linus K. <linus@cosmos-ink.net>"
+license=MIT
+
+image=rust:v1.1
+source=(https://github.com/LinusCDE/chessmarkable/archive/e97cb952cbbd806a94cea527ccd652fcf2302e50.zip)
+sha256sums=(4088c54d2db5de9fe56be99e3969dc5c1eb6a52d6462be67262839b84064cf22)
+
+build() {
+    # Fall back to system-wide config
+    rm .cargo/config
```
raisjn, line 17: I don't get it - did you need this to build in the docker image?

raisjn (dismissed):
lgtm

raisjn:
make sure to run through shellcheck

package/chessmarkable/package
LinusCDE: I prefer to use the config to use the oecore toolchain and believe it's easier for users who want to compile it and don't have a cross compiler anyway. Mostly probably a habit of mine. I would like to keep it this way for a while though.

package/chessmarkable/package
LinusCDE: Didn't see that we changed that. Good catch!

LinusCDE:
Sorry again for dismissing your review again by accidentally by committing changes @raisjnn 😅. I still don't quite get then GH does that.

The update is mostly stability fixes for the software. And hopefully the last change for at least a few days.

raisjn (approved):
no problem, i think its the repo setting that matteo pointed out on discord

package/chessmarkable/package
raisjn: if i understand right, this is falling back to the system wide config *inside* the docker image - that's what you intend, right?

PR #81, opened 2020-09-28
Implement the process for moving packages from testing to stable

matteodelabre:
This creates the `docs/branches.md` document which explains what the two release branches are, which one users should choose, and how the transition of packages from one branch to another is handled.

It also updates `docs/package.md` to add related information about the `maintainer` field.

This implements the proposal in #20, and closes #20 as a consequence.

PR #98, opened 2020-10-04
Stable merge window for week 41 of 2020

matteodelabre:
Following the (still RFC) process outlined in #20 and #81, this PR is the first merge-to-stable window. All packages which have not been changed since last Friday (i.e. all of them) have been included (including `appmarkable`, `harmony`, `mines`, `nao`, `remux` and `simple` in which only the `url=` field was changed over the week-end).

We need to test that all those packages work, both on their own and with each other.

> Here are important things to check when testing a package:
> 
> 1. The package should work as intended by upstream.
> 2. It must not destroy any user data (e.g. from previous versions of the same package, from other packages, from the home directory).
> 3. It must not break other packages.

Tested:

* [x] appmarkable
* [x] calculator
* [x] draft
* [x] harmony
* [x] mines
* [x] nao
* [x] oxide
* [x] remux
* [x] simple
* [x] fingerterm
* [x] plato
* [x] recrossable
* [x] remarkable-splash
* [x] retris
* [x] xochitl
* [x] rmservewacominput


matteodelabre:
Looks like Draft does not provide a way to scroll when there are more than 7 apps installed. See dixonary/draft-reMarkable#24.

Eeems:
I've been using oxide as my daily driver for a while now. I can confirm that the following other applications also work with it:
 - calculator
 - harmony
 - mines
 - nao
 - simple
 - plato
 - recrossable
 - retris
 - xochil

KOReader has an outstanding issue with it grabbing all input. There is an environment variable that needs to be defaulted to true to disable the grabbing that would keep it from having issues with Oxide.

matteodelabre:
Let’s defer adding KOReader until the next merge window then.

matteodelabre:
Harmony is unable to save drawings if the `/home/root/harmony` folder does not exist, and the `harmony` package does not create this folder automatically. See rmkit-dev/rmkit#30.

matteodelabre:
`rmservewacominput` does not show up in Oxide even after manually importing apps. It is listed in Draft though.

raisjn:
> Let’s defer adding KOReader until the next merge window then.

Why? This is an issue that requires changes to oxide, not KOReader

> Harmony is unable to save drawings if the /home/root/harmony folder does not exist, and the harmony package does not create this folder automatically. See rmkit-dev/rmkit#30.

This can be fixed in harmony source code, but I think package should make this directory for now

matteodelabre:
> > Let’s defer adding KOReader until the next merge window then.
> 
> Why? This is an issue that requires changes to oxide, not KOReader

Isn’t that environment variable specific to KOReader? If it is, the solution would be to add it to KOReader’s draft file, and this change is not related to Oxide.

> > Harmony is unable to save drawings if the /home/root/harmony folder does not exist, and the harmony package does not create this folder automatically. See [rmkit-dev/rmkit#30](https://github.com/rmkit-dev/rmkit/issues/30).
> 
> This can be fixed in harmony source code, but I think package should make this directory for now

Ok, so let’s move harmony to next week’s merge window too?

raisjn:
> Ok, so let’s move harmony to next week’s merge window too?

would prefer to just fix in the package for now

> Isn’t that environment variable specific to KOReader? If it is, the solution would be to add it to KOReader’s draft file, and this change is not related to Oxide.

this is not necessarily applicable to all launchers, not really sure. but putting KOReader in now would be fine, imo, as its current behavior is what the behavior has been all along until the recent upstream patches that were submitted. it doesn't make sense to me that oxide can block KOReader

matteodelabre:
> > Ok, so let’s move harmony to next week’s merge window too?
> 
> would prefer to just fix in the package for now

But this would go against the rule in RFC #81 stating that the merge PR “can only contain package changes dating from the previous Friday or older”. To be consistent we should either change that rule or follow it.

> > Isn’t that environment variable specific to KOReader? If it is, the solution would be to add it to KOReader’s draft file, and this change is not related to Oxide.
> 
> this is not necessarily applicable to all launchers, not really sure. but putting KOReader in now would be fine, imo, as its current behavior is what the behavior has been all along until the recent upstream patches that were submitted. it doesn't make sense to me that oxide can block KOReader

The thing is that, with the current set of packages, KOReader does not play well with Oxide. For example:

1. Start `tarnish`.
2. Tap `KOReader`.
3. Hold the left key to get back to the launcher.
4. KOReader does not yield input back to the launcher, and the screen image is distorted:

<p align="center">Photo showing the distorted screen image after launching KOReader and pressing the left key</p>

So we can’t include both in the stable branch, IMO, and therefore we need to choose between the two.

Eeems:
> 
> 
> `rmservewacominput` does not show up in Oxide even after manually importing apps. It is listed in Draft though.

It's passing arguments as part of the `call=` spec (https://github.com/LinusCDE/rmWacomToMouse/blob/master/oxide#L3), which oxide doesn't support importing. I don't think this should block it from going forward. I'd a known breaking change in Oxide 2.0+. Oxide has instructions on how to work around it.

> 
> > > Isn’t that environment variable specific to KOReader? If it is, the solution would be to add it to KOReader’s draft file, and this change is not related to Oxide.
> > 
> > 
> > this is not necessarily applicable to all launchers, not really sure. but putting KOReader in now would be fine, imo, as its current behavior is what the behavior has been all along until the recent upstream patches that were submitted. it doesn't make sense to me that oxide can block KOReader
> 
> The thing is that, with the current set of packages, KOReader does not play well with Oxide. For example:
> 
>     1. Start `tarnish`.
> 
>     2. Tap `KOReader`.
> 
>     3. Hold the left key to get back to the launcher.
> 
>     4. KOReader does not yield input back to the launcher, and the screen image is distorted:
> 
> 
> Photo showing the distorted screen image after launching KOReader and pressing the left key
> 
> So we can’t include both in the stable branch, IMO, and therefore we need to choose between the two.

See Oxide's known issues for the workaround: https://github.com/Eeems/oxide/releases/tag/v2.0-beta We could try to script it in the package. I agree with @raisjnn here, since this is a KOReader issue that has existed from the start and not a new regression, we should ship it and work on resolving it in the next release window.

raisjn:
* recrossable has a bug where the numbering is incorrectly starting in the bottom right and the game is difficult to actually play because of this (because the clues and numbering do not use the usual crossword direction)

matteodelabre:
> It's passing arguments as part of the `call=` spec (https://github.com/LinusCDE/rmWacomToMouse/blob/master/oxide#L3), which oxide doesn't support importing. I don't think this should block it from going forward. I'd a known breaking change in Oxide 2.0+. Oxide has instructions on how to work around it.

Agreed.

> See Oxide's known issues for the workaround: https://github.com/Eeems/oxide/releases/tag/v2.0-beta We could try to script it in the package.

I see. So we just need to change the .draft file in the KOReader package, right?

>  I agree with @raisjnn here, since this is a KOReader issue that has existed from the start and not a new regression, we should ship it and work on resolving it in the next release window.

Since you both agree on this, I’ll align with you and add it back. We’ll need to make #81 more explicit about this in the future.

raisjn:
> But this would go against the rule in RFC #81 stating that the merge PR “can only contain package changes dating from the previous Friday or older”. To be consistent we should either change that rule or follow it.

I don't think you need to be strict on the first merge window. If we are following the agreed upon rules (what prevents a package from being moved from testing -> stable), it would be that the original author of the package (or maintainer?) would say "yes, i think it's behaving as intended" and does not break any existing packages or delete user data.

Eeems:
> 
> 
> > It's passing arguments as part of the `call=` spec (https://github.com/LinusCDE/rmWacomToMouse/blob/master/oxide#L3), which oxide doesn't support importing. I don't think this should block it from going forward. I'd a known breaking change in Oxide 2.0+. Oxide has instructions on how to work around it.
> 
> Agreed.
> 
> > See Oxide's known issues for the workaround: https://github.com/Eeems/oxide/releases/tag/v2.0-beta We could try to script it in the package.
> 
> I see. So we just need to change the .draft file in the KOReader package, right?

No, we'd need to add a script that registers the application in oxide and provides the onPause and onResume calls. We should make sure that the appropriate `term=` setting is in place though.

matteodelabre:
> No, we'd need to add a script that registers the application in oxide and provides the onPause and onResume calls. We should make sure that the appropriate `term=` setting is in place though.

Would you be open to supporting onPause and onResume fields on the .draft file in a future release of Oxide? I’d rather avoid having launcher-specific code in Toltec.

matteodelabre:
> Harmony is unable to save drawings if the `/home/root/harmony` folder does not exist, and the `harmony` package does not create this folder automatically. See [rmkit-dev/rmkit#30](https://github.com/rmkit-dev/rmkit/issues/30).

Can confirm that #99 fixes this.

matteodelabre:
>     * recrossable has a bug where the numbering is incorrectly starting in the bottom right and the game is difficult to actually play because of this (because the clues and numbering do not use the usual crossword direction)

I was wondering why it was so hard to find the words by number while playing, so this is the answer. :)
I think this ought to be reported upstream though.

Eeems:
Opened #102 to deal with some of the concerns with KOReader.

> 
> 
> > No, we'd need to add a script that registers the application in oxide and provides the onPause and onResume calls. We should make sure that the appropriate `term=` setting is in place though.
> 
> Would you be open to supporting onPause and onResume fields on the .draft file in a future release of Oxide? I’d rather avoid having launcher-specific code in Toltec.

No, I'd like to move away from the draft configuration file format completely. I will support a different format for packages to declare applications that will be automatically handled though. That is future state though.

package/remux/remux.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=App launcher that supports multi-tasking applications
+After=xochitl.service opt.mount
+StartLimitInterval=30
+StartLimitBurst=5
+Conflicts=draft.service tarnish.service xochitl.service
```
raisjn, line 6: can it conflict with `xochitl.service` and also use `after xochitl.service` - what's that mean?

raisjn (approved):

package/remux/remux.service
matteodelabre: This looks like an error indeed. Remux is not supposed to conflict with xochitl. Systemd seems to handle this correctly however, so this is not blocking. I’ll open a separate issue (#104). Thanks!

raisjn:
> I think this ought to be reported upstream though.

https://github.com/sandsmark/recrossable/issues/4

PR #99, opened 2020-10-05
package: add mkdir -p /home/root/harmony to harmony installation

raisjn:


raisjn:
addresses harmony break in https://github.com/toltec-dev/toltec/pull/98

package/harmony/package
```diff
@@ -25,3 +25,7 @@ package() {
     mkdir -p "$pkgdir"/opt/etc/draft/
     install -D -m 644 "$srcdir"/src/harmony/harmony.draft "$pkgdir"/opt/etc/draft/harmony.draft
 }
+
+configure() {
+    mkdir -p /home/root/harmony 2> /dev/null
```
matteodelabre: This redirect to `/dev/null` is not needed since `mkdir -p` does not print anything, even if the directory already exists.

```suggestion
    mkdir -p /home/root/harmony
```

matteodelabre (changes requested):
Looks good to me, but you need to increment the package revision number (i.e. `0.0.1-14` → `0.0.1-15`). We should probably have an automated check for that (#100).

matteodelabre (approved):
Thanks!

PR #101, opened 2020-10-05
Make the package listings sortable

matteodelabre:
This PR adds a small JS snippet to the packages listing so that users with JS enabled can sort the packages by name, section, license, or other fields. This resolves the following comment: https://github.com/toltec-dev/toltec/pull/88#issuecomment-703116295.

Splitting the CSS and JS to separate files would enable them to be linted and formatted separately, plus enable clients to cache them. But at this point I’m not sure how to best structure this in the repository. Any ideas?

### DIFF
--- a/scripts/repo-build-web
+++ b/scripts/repo-build-web
@@ -89,10 +89,6 @@ cat > "$indexfile" << 'WEBS'
                 cursor: pointer;
             }
 
-            .sortable th:not(.sort-asc):not(.sort-desc)::after {
-                content: " ⇅";
-            }
-
             .sortable th.sort-asc::after {
                 content: " ↓";
             }
--- a/scripts/repo-build-web
+++ b/scripts/repo-build-web
@@ -44,7 +44,9 @@ recipesdir="$1"
 repodir="$2"
 mkdir -p "$repodir"
 
-cat > "$repodir/index.html" << 'WEBS'
+indexfile="$repodir/index.html"
+
+cat > "$indexfile" << 'WEBS'
 <!doctype html>
 <html lang="en">
     <head>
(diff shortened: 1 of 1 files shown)


raisjn:
if this is about the comment about splitting into separate tables for each section, this diff is not the same as the suggestion i made

the suggestion i made originally doesn't require js to use (as it can be done statically)

matteodelabre:
Indeed. But it achieves the same effect for users who enable JS. And it does not prevent us from splitting the table statically in a future PR.

okayzed:
did you understand the original suggestion and decide this was better?

re: static assets: that's a non-issue, this is maybe a few K of code and caching is not *that* relevant. in addition, inlining CSS / JS (in small amounts) is faster than splitting the page delivery over multiple requests because there's less overhead. this is a common tactic at google, for example, where they put the most important CSS into the main page delivery.

raisjn:
did you understand the original suggestion and decide this was better? my understanding from the comments / conversation is that we would add the table split in another diff. 

re: static assets: that's a non-issue, this is maybe a few K of code and caching is not that relevant. in addition, inlining CSS / JS (in small amounts) is faster than splitting the page delivery over multiple requests because there's less overhead. this is a common tactic at google, for example, where they put the most important CSS into the main page delivery.

matteodelabre:
> did you understand the original suggestion and decide this was better? my understanding from the comments / conversation is that we would add the table split in another diff.

Yes, I did understand the original suggestion. This one seems more flexible to me since it does not force a specific view of the listing on the user. But it can be combined with a static split by section with a separate PR. Right now, I don’t think there are enough packages to warrant this split though.

> re: static assets: that's a non-issue, this is maybe a few K of code and caching is not _that_ relevant. in addition, inlining CSS / JS (in small amounts) is faster than splitting the page delivery over multiple requests because there's less overhead. this is a common tactic at google, for example, where they put the most important CSS into the main page delivery.

Yup. But the argument about making it easier to lint the CSS and JS parts still stands.

okayzed:
> Indeed. But it achieves the same effect for users who enable JS. 

For users who enable JS. The initial suggestion is aimed at users who do not. Would prefer to just go that route than adding JS into the page

raisjn:
> Indeed. But it achieves the same effect for users who enable JS.

For users who enable JS. The initial suggestion is aimed at users who do not. Would prefer to just go that route than adding JS into the page

raisjn:
If you think we don't need to make the non-JS version, i can review this diff, but i wanted to make sure that you understood where the original feature idea and motivation are coming from

matteodelabre:
Here’s a compromise that statically splits the listing in sections and allows client-side per-section sorting.

scripts/repo-build-web
```diff
@@ -57,77 +59,185 @@ cat > "$repodir/index.html" << WEBS
...
 
+    cat >> "$indexfile" << 'WEBS'
+        </table>
+WEBS
 done
 
-cat >> "$repodir/index.html" << WEBS
-        </table>
+cat >> "$indexfile" << 'WEBS'
+        <script>
```
raisjn, line 175: why is this indented?

raisjn (commented):
looks good! 

https://imgur.com/KZ7rz7A.png - on the first page load, it shows both arrows for some columns. does both arrows mean unsorted?

scripts/repo-build-web
matteodelabre: This is to match the indent level of the rest of the generated HTML output.

matteodelabre:
> https://imgur.com/KZ7rz7A.png - on the first page load, it shows both arrows for some columns. does both arrows mean unsorted?

Indeed. But I fear it’s actually making things less clear. I’ll remove it.

raisjn (approved):
shipit

raisjn:
is the ordering of the sections non-deterministic?

matteodelabre:
Looks like Bash is using a non-deterministic hashtable of some kind for storing associative arrays. We’ll need to sort the list. See #106.

PR #102, opened 2020-10-05
Don't grab input in KOReader by default.

Eeems:


raisjn (dismissed):
lgtm

Eeems:
I've added a warning message on install if it detects that Oxide is installed. I tested the logic on device. This doesn't handle showing a warning when Oxide is installed after KOReader though.

raisjn (approved):
shipit

matteodelabre (approved):
Tested, can confirm it fixes the issue. Thanks!

PR #113, opened 2020-10-07
recrossable: Update and remove patch

matteodelabre:
recrossable has been updated so that the patch to fix pen input events is no longer needed. See sandsmark/recrossable#3.


raisjn (approved):
lgtmacro

PR #116, opened 2020-10-08
Improve source archive extraction

matteodelabre:
Behavior change
---------------

Previously, the topmost directories were always stripped when extracting source archives, which was confusing when the archive contained more than one top-level directory (see <https://github.com/toltec-dev/toltec/pull/74#discussion_r495615907>).

This PR changes that behavior to a more prudent one, which finds the longest folder prefix in the archive (potentially empty), and removes it when extracting the archive’s files. For example, when extracting an archive containing:

    a/b/c/file1.txt
    a/b/c/file2.txt

the `a/b/c/` prefix will be removed. When extracting an archive containing:

    a/file1.txt
    b/file2.txt

both the `a/` and `b/` folders will be kept.

Changes
-------

* Implement the new behavior
* Document it in docs/package.md#sources
* Update oxide to take advantage of this

### DIFF
--- a/docs/package.md
+++ b/docs/package.md
@@ -205,10 +205,10 @@ See <https://www.debian.org/doc/debian-policy/ch-relationships.html#conflicting-
     </tr>
 </table>
 
-Docker image used for building this package.
+Docker image to use for building the package.
 It can be omitted only for packages which do not require a build step (see [below](#build-section-optional)).
 
-#### `source` 
+#### `source`
 
 <table>
     <tr>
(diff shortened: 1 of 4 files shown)


raisjn:
do all packages build with this change?

raisjn (approved):
shipit

matteodelabre:
> do all packages build with this change?

~~Actually it breaks KOReader, thanks for reminding me to check that.~~ EDIT: It’s not because of this PR, it looks like KOReader’s source file has disappeared from the server.

raisjn:
>  EDIT: It’s not because of this PR, it looks like KOReader’s source file has disappeared from the server.

must be from using nightlies (so we get the upstreamed patches for better launcher integration). the next monthly release from KOReader should be happening within the two weeks. we can switch to new nightly, then switch to the monthly release in 2 weeks: http://build.koreader.rocks/download/nightly/

matteodelabre:
The other packages build fine.

PR #118, opened 2020-10-09
Fix KOReader build

matteodelabre:
KOReader nightly 2020.09-29 was removed from their servers, so this PR upgrades to 2020.09-53 to address that. We will need to update it to the latest stable release when it is released later in this month.

Also add the `--show-error` flag to curl invocations so that an explicit error message is shown in the build output when the source of a package is not available (previously, nothing was shown and the script simply exited).


raisjn (approved):
nice, thank you! i think the build will fall off the server one more time before their  monthly update

PR #119, opened 2020-10-10
Add reboot-guard

Eeems:
I'm likely going to add this as a dependency for Oxide.


raisjn (dismissed):
:cool:

how big is python? 30MBish download?

Eeems:
> cool
> 
> how big is python? 30MBish download?

Not quite sure, the python3 package is only 967 bytes compressed. Didn't check what else it pulls down.

Eeems:
Tested and it now works.

raisjn (approved):
shipit

PR #121, opened 2020-10-11
Fix handling packages with multiple dependencies

matteodelabre:
Previously only the first element of the $depends (resp. $conflicts) array was used to build the list of dependencies (resp. conflicting packages) of a packages, due to an error in the build script. This caused an error in the generated package for `reboot-guard`, where the `python` dependency was not automatically installed. This PR fixes that.

It also changes the version of `reboot-guard` from `1.0.1~1` to `1.0.1-1` to bring it in line with other package versions.

### DIFF
--- a/package/reboot-guard/package
+++ b/package/reboot-guard/package
@@ -5,7 +5,7 @@
 pkgname=reboot-guard
 pkgdesc="Block systemd-initiated poweroff/reboot/halt until configurable condition checks pass"
 url=https://github.com/stephanritscher/reboot-guard
-pkgver=1.0.1~1
+pkgver=1.0.1-1
 timestamp=2020-05-04T06:16Z
 section=utils
 maintainer="Eeems <eeems@eeems.email>"
--- a/scripts/package-lib
+++ b/scripts/package-lib
@@ -156,6 +156,29 @@ rsecurl() {
     rcurl --proto '=https' "$@"
 }
 
+# Join elements with the given delimiter
+#
+# Arguments:
+#
+# $1 - Separator string
+# $2... - Elements to join
+#
+# Output:
+#
+# Joined string.
+join-elements() {
+    local sep="$1"
+    local cursep=""
+    shift
+
+    for arg; do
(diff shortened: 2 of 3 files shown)


scripts/package-lib
```diff
@@ -156,6 +156,29 @@ rsecurl() {
...
+#
+# $1 - Separator string
+# $2... - Elements to join
+#
+# Output:
+#
+# Joined string.
+join-elements() {
+    local sep="$1"
+    local cursep=""
```
raisjn, line 171: i don't see how $2 relates to cursep

package/reboot-guard/package
```diff
@@ -5,7 +5,7 @@
 pkgname=reboot-guard
 pkgdesc="Block systemd-initiated poweroff/reboot/halt until configurable condition checks pass"
 url=https://github.com/stephanritscher/reboot-guard
-pkgver=1.0.1~1
+pkgver=1.0.1-1
```
raisjn, line 8: thanks. i need to get a better eye for this

scripts/package-lib
```diff
@@ -156,6 +156,29 @@ rsecurl() {
...
+    shift
+
+    for arg; do
```
raisjn, line 174: neat, i didn't know about this syntax

package/reboot-guard/package
matteodelabre: Or we need to work on #8 :)

scripts/package-lib
matteodelabre: $2 and the following arguments are the elements to be joined together.
$cursep is null on the first loop iteration, then is set to the separator in $1 (this is so that we don’t have a trailing separator on the left of the resulting string).

scripts/package-lib
matteodelabre: Learned about it today too!

scripts/package-lib
raisjn: that makes sense

raisjn (approved):
lgtm

PR #122, opened 2020-10-11
Fix reboot-guard dependency

matteodelabre:
Entware’s Python package is called `python3` and not `python`.


raisjn (dismissed):
shipit

matteodelabre:
Sorry to dismiss your review @raisjnn, but I forgot to increase the package version.

raisjn (approved):
np

PR #124, opened 2020-10-12
Configure Opkg to fetch Entware packages using TLS

matteodelabre:
When I merged remarkable_entware into the bootstrap script, I erroneously did not keep @LinusCDE’s sed line that changes `opkg.conf` so that Entware packages are fetched using TLS. Therefore, in the current state of the bootstrap script, Entware packages are fetched through raw HTTP.

This PR adds back [that missing line](https://github.com/LinusCDE/remarkable_entware/blob/4288859dd28ca57d91e16b96621261e20fee7807/entware_install.sh#L147).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ To automatically install Opkg, Entware and Toltec, run the bootstrap script in a
 
 ```sh
 $ wget http://toltec.delab.re/bootstrap
-$ echo "3c792fe4dcd034dcbf52ddd242fa3816d48a776acdbd78b3bec2a7741416ed35  bootstrap" | sha256sum -c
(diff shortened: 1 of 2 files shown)


raisjn (approved):
lgtm

it looks like someone can accidentally update bootstrap without updating readme checksum?

matteodelabre:
[scripts/bootstrap/checksum-check](https://github.com/toltec-dev/toltec/blob/testing/scripts/bootstrap/checksum-check) is part of the PR lint step, so that if someone forgets to update the checksum, the PR is blocked.

raisjn:
awesome!

PR #125, opened 2020-10-12
Update KOReader to version 2020.10

matteodelabre:
Contains rM-specific fixes: koreader/koreader#6772, koreader/koreader#6676, koreader/koreader#6769.

This also fixes the issue with the nightly versioning scheme.


raisjn (approved):
shipit

PR #127, opened 2020-10-16
Improve package docs and fix #120

### LINKED ISSUE #120: build step doesn't run if there is no image
Either we need to update our documentation to be clearer that build only happens if `image=` is populated, or still run it for simple things like patching files.

matteodelabre:
Fix typos and some unclear sentences. Add errors about missing `build()` function when `image` is set and vice versa. Fix undocumented sections. Add table of contents.

### DIFF
--- a/docs/package.md
+++ b/docs/package.md
@@ -1,18 +1,25 @@
 ## Writing a package recipe
 
-A **package recipe** is a Bash script containing metadata and instructions for building and installing a package.
-This recipe is used by the packaging script to generate installable archives for the Opkg package manager.
+A **package recipe** is a Bash script containing the metadata and instructions needed to build a package from source.
+These recipes are used by the packaging script to generate installable package archives for the Opkg package manager.
 
-> **Note:** Recipes should not be marked as executable because they are not meant to be executed directly, but rather to be sourced by the packaging script.
+> **Note:** Recipe scripts should not be marked as executable because they are not meant to be executed directly but rather meant to be sourced by the packaging script.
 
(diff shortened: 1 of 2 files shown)


docs/package.md
```diff
@@ -272,15 +283,15 @@ They can also use functions from the [install library](scripts/install-lib).
 When installing a new package, the following happens:
 
 * The package files are unpacked (but not installed)
-* `preinstall` is called, if it exists
+* `preinstall` is called if it exists
```
Eeems, line 286: Grammatically it makes more sense to leave the comma in.

raisjn (approved):
looks good

docs/package.md
raisjn: "if it exists, preinstall is called" and "preinstall is called if it exists" are valid, i think. something something dependent / independent clauses and commas

Eeems: True, I can just see us flip-flopping back and forth on this one depending on who is submitting the PR.

matteodelabre: I’m not sure about this since I’m not a native speaker. [This source](https://owl.purdue.edu/owl/general_writing/punctuation/commas/extended_rules_for_commas.html) says “don't put a comma after the main clause when a dependent (subordinate) clause follows it (except for cases of extreme contrast).” It looks like this rule applies here, doesn’t it?

Eeems: I would argue that this is a case of extreme contrast. That said English grammar is highly subjective and changes constantly.

Issue #104, opened 2020-10-05
Remove conflict declaration with xochitl.service from remux


matteodelabre:
They can run alongside each other perfectly fine, so they should not be declared as conflicting. In fact, the `Conflicts=` declaration in this case is inconsistent with the `After=` declaration which also contains `xochitl.service`.

PR #130, opened 2020-10-18
update remux.service to not conflict with xochitl.service

raisjn:
addresses #104


Eeems (approved):

Eeems (changes requested):
You'll need to update the package release number.

Eeems (approved):

PR #131, opened 2020-10-21
Implement split packages

### LINKED ISSUE #66: Split packages
Sometimes several packages can be generated from a single source. For example, harmony, nao, mines, remux and simple all come from the rmkit Git repo. It would make packaging simpler if we could combine the four recipes into a single one, _à la_ [Arch split packages](https://jlk.fjfi.cvut.cz/arch/manpages/man/PKGBUILD.5#PACKAGE_SPLITTING).

### LINKED ISSUE #106: Sort the sections alphabetically in the generated index page
[The list of existing sections is computed using a Bash associative array.](https://github.com/toltec-dev/toltec/blob/6a6839e4e5f4ab930eb045680845d1eeb9301927/scripts/repo-build-web#L107) The keys of that array are then enumerated to generate the list of sections in the index page. The ordering appears to change with each build, which is probably because Bash is using a non-deterministic hashtable of some kind for storing associative arrays. It would be best to ensure that the order is consistent across builds by sorting the list of sections alphabetically.

matteodelabre:
Loosely based on Arch Linux’s split packages, this PR implements a mechanism for generating multiple packages derived from the same set of build artifacts (fixes #66).

The build working directory is now organized into two levels: `build/package/<RECIPENAME>/<PACKAGENAME>`, which reflects the nesting of several packages inside a single recipe.

Changes:

* Make `package-build` accept arguments to select a subset of a recipe’s packages to build (by default, build all of them).
* Update `repo-build` to work with split packages.
* Update `rebo-build-web` to work with split packages (incidentally fixes #106 since the code to generate the list of categories was reworked).
* Merge recipes `harmony`, `mines`, `nao`, `remux` and `simple` into a single `rmkit` package.
* Split Oxide’s task manager to a separate package (@Eeems, are there any other binaries from Oxide that can be moved to standalone packages?).


Eeems:
For oxide you'll want to split it into the following packages:

- oxide (launcher application) (Depends on tarnish, erode, fret)
- rot (Command line api application) (Depends on tarnish)
- tarnish (System service)
- fret (Screenshot application) (Depends on tarnish)
- erode (Task manager) (Depends on tarnish)

Technically the dependency tree makes rot optional.

matteodelabre:
Thanks for the info! I have a some more questions if you don’t mind:

- Does it make sense to split tarnish and oxide at this point? tarnish still looks tied to oxide since it needs the `/etc/dbus-1/system.d/codes.eeems.oxide.conf` file to be installed to run properly.
- Is tarnish really a hard dependency for erode? I’m able to use it just fine even without having `tarnish` installed.
- It also seems like I can run oxide just fine without having erode or fret installed.
- I’m not sure how to test rot, I tried `rot wifi`, `rot apps`, but I only get segmentation faults. Is there more information somewhere?

Eeems:
- `/etc/dbus-1/system.d/codes.eeems.oxide.conf` should be part of the tarnish package, as tarnish provides that dbus service.
- Right now no, but very soon it'll be properly integrated with the API provided by tarnish.
- Correct, but the default oxide environment assumes they are installed. More realistically the dependencies for oxide is just tarnish, and tarnish depends on oxde, erode, and fret.
- `rot --help`, `rot wifi get state`, `rot apps get runningApplications`. Could you log an issue about the segfaults, It should really spit out the help message.

matteodelabre:
I’m a bit confused on how to best split the oxide package. For example, you say that tarnish depends on erode and that soon erode will depend on tarnish. Therefore if we split tarnish and erode into two packages, those two will create a circular dependency. Did I miss something?

Eeems:
Well tarnish assumes that erode, oxide, and fret all exist, but it shouldn't fall over without them. On the other hand, oxide and fret will fail to start without tarnish.

matteodelabre:
Thanks for clarifying! I updated the PR accordingly.

LinusCDE (dismissed):
Lgtm.
I mostly glanced over the code but did a lot of QA instead.

Some notes:
 - About 70% of all conditions seem to not require the bash extension (double brackets `[[ ]]`) at all. (I tried cleaning them all up, but removed probably too much as the code didn't run anymore.) I also understand that having those only double backets makes the code look cleaner, but may confuse people who are not all too familiar with all bash extensions and don't immediately realize that most are standard posix anyway. Maybe only change the files that exclusively use standard posix conditions to single brackets.
 - `make repo-local` builds fine and the file tree of every generated package seems alright to me.
 - Sidenote: Some packages still reference qt::v1.1. Is there any reason not to update them? I'll try to look out that new updates to affected packages to also update the qt-toolchain version along with it.
 - I successfully tested the configure() section to get properly executed as I didn't know that opkg runs `postinst` with the "configure" positional parameter.

All in all, it seems to work very well and I don't have any real issues with the code (I couldn't do it better myself anyway).

I'll approve this, but **maybe hold the merge back a bit** as @raisjnn may also make a probably more proper/deeper review on the weekend as well.

Makefile
```diff
@@ -2,8 +2,9 @@
 # SPDX-License-Identifier: MIT
 
 HOST?=10.11.99.1
-PACKAGES=$(shell ls package/)
-PUSH_PACKAGES=$(foreach app, $(PACKAGES), $(app)-push)
+RECIPES=$(shell ls package/)
+RECIPES_PUSH=$(foreach app, $(RECIPES), $(app)-push)
+RECIPES_CLEAN=$(foreach app, $(RECIPES), $(app)-clean)
```
raisjn, line 7: :+1:

package/appmarkable/package
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) 2020 The Toltec Contributors
 # SPDX-License-Identifier: MIT
 
-pkgname=appmarkable
+pkgnames=(appmarkable)
```
raisjn, line 5: is it easy to support both pkgname and pkgnames?

package/rmkit/package
```diff
@@ -0,0 +1,111 @@
...
+    pip3 install okp
+    make
+}
+
+harmony() {
+    pkgdesc="Procedural sketching app"
+    url="https://rmkit.dev/apps/harmony"
+    pkgver=0.0.1-17
+    section=drawing
+    conflicts=(rmkit)
```
raisjn: ok, ok, we can delete this :-D

scripts/package-lib
```diff
@@ -192,8 +218,8 @@ join-elements() {
 # $2 - Expected type of the field, 'string', 'array' or 'map'
 # $3 (optional) - Either 'nonempty' or 'optional'
 check-field() {
-    local signature
-    signature="$(declare -p "$1" 2> /dev/null)"
+    # shellcheck disable=SC2155
```
raisjn, line 221: (nit: i sometimes see people put a comment about what the lint message being disabled is, like so:)

```suggestion
    # shellcheck disable=SC2155
    # SC2155: Declare and assign separately to avoid masking return values.
```

scripts/repo-build-web
```diff
@@ -107,17 +107,22 @@ cat > "$indexfile" << 'WEBS'
...
+        (
+            load-recipe-header "$recipedir"
+            for pkgname in "${pkgnames[@]}"; do
+                (
+                    load-recipe-pkg "$pkgname"
+                    echo "$section"
+                )
+            done
+        )
+    done | sort | uniq
```
raisjn, line 122: nice

raisjn:
> I mostly glanced over the code but did a lot of QA instead.

this is super helpful, thanks! :-D

> About 70% of all conditions seem to not require the bash extension (double brackets [[ ]]) at all. (I tried cleaning them all up, but removed probably too much as the code didn't run anymore.) I also understand that having those only double backets makes the code look cleaner, but may confuse people who are not all too familiar with all bash extensions and don't immediately realize that most are standard posix anyway. Maybe only change the files that exclusively use standard posix conditions to single brackets.

i'm wondering if that is due to shellcheck compliance or something.

raisjn (dismissed):
changes look good to me

this is a super cool feature, thanks for adding it! <3

package/appmarkable/package
matteodelabre: It should be easy enough yes. When considering the two options, the possible confusion caused by having two similarly-named fields in the `docs/package.md` document came to mind. IMO it’s less confusing if we only have `pkgnames`.

scripts/package-lib
matteodelabre: Seems reasonable. Thanks!

matteodelabre:
Thanks for reviewing @raisjnn! I addressed your suggestions.

matteodelabre:
Also, thanks @LinusCDE for your review!

> About 70% of all conditions seem to not require the bash extension (double brackets `[[ ]]`) at all. (I tried cleaning them all up, but removed probably too much as the code didn't run anymore.) I also understand that having those only double backets makes the code look cleaner, but may confuse people who are not all too familiar with all bash extensions and don't immediately realize that most are standard posix anyway. Maybe only change the files that exclusively use standard posix conditions to single brackets.

> i'm wondering if that is due to shellcheck compliance or something.

This is just a personal preference of mine, which is not enforced by shellcheck actually (I don’t think they have a built-in check for that yet). Since `[[` is a superset of `[` feature-wise, I think it’s less surprising overall if we use it everywhere.

> Sidenote: Some packages still reference qt::v1.1. Is there any reason not to update them? I'll try to look out that new updates to affected packages to also update the qt-toolchain version along with it.

No particular reason that I know of. This is something that could be fixed in a PR, but is not top priority IMO.

> I tested the configure() section to get properly executed as I didn't know that opkg runs `postinst` with the "configure" positional parameter.

Great! Opkg follows dpkg as far as I know: https://www.debian.org/doc/debian-policy/ap-flowcharts.html

raisjn (approved):
doitlive

PR #135, opened 2020-10-31
Add VNSee 0.3.0

matteodelabre:
Since GitHub source code archives do not include submodules, I had to download the libvncserver submodule separately. This source needs to be placed in a specific subfolder of the $srcdir instead of the root, so we need to override the automatic archive extraction behavior in this case.

To enable that, the two following changes are also included in this PR:

* Add an optional `noextract` field to the recipes to allow disabling the automatic extraction of downloaded archives.
* Add a `prepare()` section for running tasks on the source files before the build Docker container is started.

### DIFF
--- a/docs/package.md
+++ b/docs/package.md
@@ -10,10 +10,11 @@ Sourcing a package recipe must have no side effects: the metadata section can on
 ### Contents
 
 1. [Metadata section](#metadata-section)
-2. [Build section](#build-section)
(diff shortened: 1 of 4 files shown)


raisjn (approved):
lgtm

i took a while to read and think about prepare(), seems reasonable.

matteodelabre:
Thanks for reviewing!

PR #136, opened 2020-11-06
Prevent installation of device-incompatible packages

### LINKED ISSUE #129: feature: prevent installation of rM2 incompatible apps on rM2
~~remux package should not recommend people disable xochitl, this caused someone on discord to break their rM2~~

it should be possible to prevent packages from being installed on rM2 if they are incompatible

LinusCDE:
Here is the first draft in solving the issue of packages working only on certain models.

Currently there is still a problem that when someone attempts to install the wrong package (later as a dependency), the fail of the script still marks the wrong rmx-only package to be installed. This causes an conflict with actually supported packages (that use the other rmx-only pkg). Does anyone have an idea on how to solve this?

```
reMarkable: /home/root/ opkg install rm2-only
Installing rm2-only (0.1.0-1) to root...
Downloading https://rmtestrepo.cosmos-ink.net/rm2-only_0.1.0-1_armv7-3.2.ipk
Your devices got identified as: reMarkable 1.0
You are attempting to install a package that only works on the reMarkable 2.
Collected errors:
 * pkg_run_script: package "rm2-only" preinst script returned status 1.
 * preinst_configure: Aborting installation of rm2-only.
 * opkg_install_cmd: Cannot install package rm2-only.
reMarkable: /home/root/ opkg remove retris   
Removing package retris from root...
reMarkable: /home/root/ opkg install retris
Installing retris (0.5.8-3) to root...
Downloading https://rmtestrepo.cosmos-ink.net/retris_0.5.8-3_armv7-3.2.ipk
Installing rm1-only (0.1.0-1) to root...
Collected errors:
 * check_conflicts_for: The following packages conflict with rm1-only:
 * check_conflicts_for:         rm2-only * 
 * opkg_install_cmd: Cannot install package retris.
reMarkable: /home/root/ 
```

We could probably just ensure the right package is installed in the bootstrap script. Not sure if this is the cleanest possible way though as we would duplicate the version checking. Most people would probably also just install new packages with opkg directly making the fix ineffective for them (though this are probably not many).


package/rmx-only/package
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(rm1-only rm2-only)
+pkgdesc="Ensure packages are not installed on the wrong device"
+pkgver=0.1.0-1
+url=http://0.0.0.0/
```
LinusCDE: Do we have a better way for expressing the missing url?

package/rmx-only/package
```diff
@@ -0,0 +1,50 @@
...
+timestamp=2020-11-06T09:35Z
+section=meta
+maintainer="Linus K. <linus@cosmos-ink.net>"
+license=MIT
```
LinusCDE: We should probably changes this to be equal to the repo maintainers and license. These packages will by more of an first-party/internal one.

package/rmx-only/package
Eeems: It might make more sense to use https://remarkable.com

raisjn:
should we add rm1-only to packages that aren't rm2 compatible yet? (plato and...?)

LinusCDE:
This PR isn't ready at the moment. If a user installed a rM 1 package on his rM 2 first, the rm1-only package would fail configuration because of the check, but it would still be technically installed and conflict with any future rM 2 only package (and vice versa ofc).

I'm not sure, but is there another way to prevent installation of these meta packages or make them just block others (triggers maybe)?

Funnily enough, I plan on making plato's input and battery work in a few hours.

Eeems:
Would it make more sense to have the check in preinstall to avoid that?

Edit: I see that it is in fact in the preinstall that you have this already, so it should work fine as is.

LinusCDE:
> Would it make more sense to have the check in preinstall to avoid that?
> 
> Edit: I see that it is in fact in the preinstall that you have this already, so it should work fine as is.

I thought this as well. But if I exit with an error. The package/dependency will fail, but count as installed the next time.

Eeems:
> > Would it make more sense to have the check in preinstall to avoid that?
> > Edit: I see that it is in fact in the preinstall that you have this already, so it should work fine as is.
> 
> I thought this as well. But if I exit with an error. The package/dependency will fail, but count as installed the next time.

Hmm odd. I guess the better way would be to have bootstrap install the package by default and use it as a conflicts? So instead of rm1-only/rm2-only the package could be rm1 and rm2 and have packages conflict with rm2.

LinusCDE:
> > > Would it make more sense to have the check in preinstall to avoid that?
> > > Edit: I see that it is in fact in the preinstall that you have this already, so it should work fine as is.
> > 
> > 
> > I thought this as well. But if I exit with an error. The package/dependency will fail, but count as installed the next time.
> 
> Hmm odd. I guess the better way would be to have bootstrap install the package by default and use it as a conflicts? So instead of rm1-only/rm2-only the package could be rm1 and rm2 and have packages conflict with rm2.

Thats a good idea as well. It would probably make current packages incompatible/buggy when using an older bootstrap script, but luckily this repo isn't announced yet. So this won't be a problem in my opinion.

Eeems:
> Thats a good idea as well. It would probably make current packages incompatible/buggy when using an older bootstrap script, but luckily this repo isn't announced yet. So this won't be a problem in my opinion.

Which is where implementing https://github.com/toltec-dev/toltec/issues/140 is a good idea still. We could also add some generic install scripts to warn users when it's missing.

LinusCDE:
That's a really good idea. Didn't even think of that idea anymore. The fact that I basically stated the problem, you want solved there, proves your point. :laughing:

LinusCDE:
I added your suggestions. Haven't tested it out yet.
We should probably take on #140 first and then add an autoinstall of the rm1 or rm2 package to the bootstrap script in this PR.

package/rmx-only/package
Eeems: toltec is MIT licensed, so I believe this is fine.

LinusCDE:
I think we should just make them two separte server-side-repos (not git repos; maybe keeping the current state for compatibility).

It would basically be an extra entry in the package which would decide where packages will get copied to on the server. the bootstrap would then decide to use an url such as https://toltec-dev.org/rm1/stable/ and so on. Even though we duplicate some files, on the server this could also simply be symlinks to /stable and /testing as before, keeping backwards compatibility and preventing duplicate files on the server.
it's probably cleaner than us trying to hack around the dependencies or similar to convince opkg to not install stuff that's not meant for a certain device.

LinusCDE:
To elaborate on the previous. Here is a mockup idea of how it could look on the server.
```
╭─linus@HomeLX /tmp/toltec-dev.org  
╰─$ tree
.
└── stable
    ├── Packages
    ├── Packages.gz
    ├── rm1
    │   ├── Packages
    │   ├── Packages.gz
    │   ├── simple_0.1.0-1_armv7-3.2.ipk -> ../simple_0.1.0-1_armv7-3.2.ipk
    │   └── tarnish_2.0.1~beta-3_armv7-3.2.ipk -> ../tarnish_2.0.1~beta-3_armv7-3.2.ipk
    ├── rm1packages
    ├── rm2
    │   ├── Packages
    │   ├── Packages.gz
    │   ├── rm2fb_1.0.0-3_armv7-3.2.ipk -> ../rm2fb_1.0.0-3_armv7-3.2.ipk
    │   ├── simple_0.1.0-1_armv7-3.2.ipk -> ../simple_0.1.0-1_armv7-3.2.ipk
    │   └── tarnish_2.0.1~beta-3_armv7-3.2.ipk -> ../tarnish_2.0.1~beta-3_armv7-3.2.ipk
    ├── rm2fb_1.0.0-3_armv7-3.2.ipk
    ├── rm2packages
    ├── simple_0.1.0-1_armv7-3.2.ipk
    └── tarnish_2.0.1~beta-3_armv7-3.2.ipk

3 directories, 16 files
```

I excluded the testing-branch in the above. Here are the mock files seen in the tree (all files are empty since that doesn't matter in the mockup): [toltec-dev.org.tar.gz](https://github.com/toltec-dev/toltec/files/5762576/toltec-dev.org.tar.gz)

in that example, users would have the following repo urls:
- rm1 + rm2 (as now): https://toltec-dev.org/stable/ 
- rm1 only: https://toltec-dev.org/stable/rm1/
- rm2 only: https://toltec-dev.org/stable/rm2/

Eeems:
> I think we should just make them two separte server-side-repos (not git repos; maybe keeping the current state for compatibility).

I'd prefer to avoid two git repos as much as possible. Two server side is okay, but we might want to create a package for handling updating the repo paths so that we have an upgrade path in the future.

LinusCDE:
Added the ideas above.

- I removed the old rmx package in commit 49e06f79
- Take a look at commit d949be4c. Maybe polish it, but check whether you agree with the definition of when to use any @Eeems 
- Take a look at commit f9b68d98 and check if I have made any mistake regarding what device a package supports @Eeems @raisjn and others 
- Commit 0292fa03 has the actual new changes. They may need some rework as it's the first draft. Since this changes the folder structure on the server @matteodelabre should take a look. I'm also not sure how I should handle the case of already existing rm1/rm2 sub-folders.

TODO:
 - [ ] Bootstrap: Automatically point to the correct repo (e.g. `https://toltec.delab.re/$toltec_branch/rm1`)
 - [ ] Bootstrap: Upgrade the old repo path (which still works) to the new one ↑
 - [ ] **Require a review from @matteodelabre** in the end, since these changes add additional content onto [his server](https://github.com/toltec-dev/toltec).

package/reboot-guard/package
```diff
@@ -5,9 +5,10 @@
 pkgnames=(reboot-guard)
 pkgdesc="Block systemd-initiated poweroff/reboot/halt until configurable condition checks pass"
 url=https://github.com/stephanritscher/reboot-guard
-pkgver=1.0.1-2
+pkgver=1.0.1-3
 timestamp=2020-05-04T06:16Z
 section=utils
+devices=(rm1 rm2)
```
Eeems:
```suggestion
devices=(any)
```

package/toltec-bootstrap/package
```diff
@@ -0,0 +1,27 @@
...
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(toltec-bootstrap)
+pkgdesc="Toltec installation script"
+url=https://toltec-dev.org/
+pkgver=0.0.1-1
+timestamp=2020-12-27T18:48Z
+section=utils
+devices=(rm1 rm2)
```
Eeems:
```suggestion
devices=(any)
```

Eeems: You can't install the current bootstrap without this, so I think it's fair for it to be marked as any.

package/xochitl/package
```diff
@@ -8,6 +8,7 @@ url=https://remarkable.com
 pkgver=0.0.0-1
 timestamp=2020-09-24T18:48Z
 section=readers
+devices=(any)
```
Eeems:
```suggestion
devices=(rm1 rm2)
```
Since we don't know what the next device we support will look like. Maybe it wont have xochitl?

package/toltec-bootstrap/package
LinusCDE: I would say it should stay non-any since it'll probably do specific tests based on the ...soc0/mashine file and compare it against known versions. So it would need an explicit update and therefore not classify for `any`.

Eeems: Since you use the bootstrap script to install, it by definition works on any device. It may refuse to continue the install on an unknown device, but the script itself should fully support any device out there.

package/xochitl/package
LinusCDE: I personally don't see a reason why this should ever fail on a future device. But I can see the reasoning and would be willing to apply this if you still want it.

LinusCDE: Maybe we should settle on the definition of `any` in the docs first: d949be4c

Eeems: I agree with the definition of `any` in that commit. Which, in that spirit, this doesn't fall under as xochitl is specific to the rm1/2 and might not exist on other devices.

Eeems:
I looked over the commits you called out and they look good.

Eeems:
@LinusCDE The current build is creating a 500MB archive :/

LinusCDE:
> @LinusCDE The current build is creating a 500MB archive :/

Then there is something not using the symlinks properly. Should around 190MB if the symlinks work correctly. Had the same issue when scp'ing. Not sure but the transfer method may need an update to prevent this unnecessary coping.

From the `man scp`
> -r      Recursively copy entire directories.  Note that scp follows symbolic links encountered in the tree traversal.

Though I think the transfer is done with something else. That may need some adjustment as well. The symlinks are relative, so coping them is fine.

package/xochitl/package
matteodelabre: rM seems committed to keeping old devices updated while moving forward with new hardware iterations. Therefore we could expect them to only introduce software differences between devices for hardware-related motives (as we’ve seen with rM2). It looks to me that this is the assumption that divides your two views on this @LinusCDE & @Eeems. If we accept this assumption, then there is no reason why `xochitl` would suddenly become unavailable for a specific hardware revision and stay available for others. If we reject it, there could be a future where a new hardware revision gets its own interface with a different naming.

In any case, I don’t think this is such a crucial issue, since for each new hardware revision there will necessarily be some period of time where we get to experiment with the new device before we move support for it to stable. In this period of time, we’ll have the opportunity to change the `devices` parameter of the `xochitl` package if need be.

Eeems: I'm also thinking that we may decide to support more than just the reMarkable line of devices in the future. Fair enough though about being able to change the devices line in the future.

LinusCDE:
----------

Third rework (currently in sync with testing-branch). Most of the above will be void.

Eeems:
@LinusCDE files changed show no changes?

LinusCDE:
The last commits reverted all of them. We didn't want to go that route as well, so I wanted to start the architecture approach.

But we haven't decided on the details yet. So the PR will probably be used when this gets done.

LinusCDE:
Will be resolved in another pr

PR #137, opened 2020-11-07
Implement a strategy for renaming packages

matteodelabre:
* Create an `old` section for deprecated or transitional packages. This section is not included in the package listing.
* Add a `replaces` recipe field, equivalent to the `Provides`, `Conflicts` and `Replaces` Debian fields combined.
* Use these new features to add an upgrade path for renaming `minesweeper` to `mines`.

To test:

* Make a clean install of Toltec stable.
* Install `minesweeper` (version 0.0.1-14).
* Change the branch to <https://toltec-dev.org/dev/feature/renaming> in `/opt/etc/opkg.conf`.
* Run `opkg update` then `opkg upgrade`.

This seems to work on my device: the `mines` package is installed, then the `minesweeper` package is removed. Opkg generates an error however, which seems like it can be safely ignored.

```
Upgrading minesweeper on root from 0.0.1-14 to 0.0.1-15...
Downloading https://toltec-dev.org/dev/feature/renaming/minesweeper_0.0.1-15_armv7-3.2.ipk
Installing mines (0.0.1-17) to root...
Downloading https://toltec-dev.org/dev/feature/renaming/mines_0.0.1-17_armv7-3.2.ipk
Removing package minesweeper from root...
Configuring mines.
Collected errors:
 * pkg_run_script: Internal error: minesweeper has a NULL  tmp_unpack_dir.
  * prerm_upgrade_old_pkg: prerm script for package "minesweeper" failed
```

I’m not sure yet whether this is an Opkg bug or if I’m doing something wrong.

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -2,7 +2,7 @@
 # Copyright (c) 2020 The Toltec Contributors
 # SPDX-License-Identifier: MIT
 
-pkgnames=(harmony mines nao remux simple)
+pkgnames=(harmony minesweeper mines nao remux simple)
 timestamp=2020-10-10T01:41+00:00
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
@@ -38,11 +38,24 @@ harmony() {
     }
 }
 
+minesweeper() {
+    pkgdesc="Transitional package for mines - can safely be removed"
+    url="https://rmkit.dev/apps/minesweeper"
+    pkgver=0.0.1-15
+    section=old
+    depends=(mines)
+
+    package() {
+        return
+    }
+}
+
 mines() {
     pkgdesc="Mine detection game"
     url="https://rmkit.dev/apps/minesweeper"
-    pkgver=0.0.1-16
+    pkgver=0.0.1-17
     section=games
+    replaces=(minesweeper)
 
     package() {
         install -D -m 755 "$srcdir"/src/build/mines "$pkgdir"/opt/bin/mines
--- a/docs/package.md
+++ b/docs/package.md
@@ -135,6 +135,7 @@ launchers       | Apps that present to the user a list of other apps that they c
 math            | Apps to assist the user in performing mathematical tasks.
 readers         | Apps for reading and annotating documents (PDF, EPUB, …).
 utils           | System tools and various apps.
+old             | Deprecated or transitional packages. Packages in this section will not be listed in the package index.
 
 If the package does not fit into one of the existing sections, you are free to create a new one and document it here.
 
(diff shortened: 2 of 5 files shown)


Eeems:
Looks like it might be an opkg bug: https://dev.archive.openwrt.org/ticket/18320.html#comment:1

matteodelabre:
I think you’re right. Since the old package gets removed nonetheless, I think we should go forward with this, unless someone can think of a cleaner workaround. Package renames should be marginal anyway.

Eeems:
Any chance we could print a message informing the user what has happened and to ignore the error?

matteodelabre:
Good idea! Inspecting the logs from the `opkg upgrade` command shows that the install scripts from the transitional package are not executed. We could however use an install script in the renamed `mines` package. We would need to make sure that the message only gets shown for users who had previously installed the old `minesweeper` package.

package/rmkit/package
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) 2020 The Toltec Contributors
 # SPDX-License-Identifier: MIT
 
-pkgnames=(harmony mines nao remux simple)
+pkgnames=(harmony minesweeper mines nao remux simple)
```
raisjn: it's a little confusing that `minesweeper` (the deprecated name) is placed at toplevel and `mines` is added as `depends` for minesweeper

raisjn (dismissed):
looks good to me, i'll try out the upgrade path when it hits testing

package/rmkit/package
Eeems: So if you have minesweeper installed it will automatically install `mines` which will then remove `minesweeper`. Eventually we can drop the `minesweeper` package.

Eeems:
I think it might be safe to remove the minesweeper package from this build since we've gone forward without it. But for future renames we'll be able to use this mechanism.

matteodelabre:
> Looks like it might be an opkg bug: https://dev.archive.openwrt.org/ticket/18320.html#comment:1

The bug is fixed in Yocto’s opkg (https://bugzilla.yoctoproject.org/show_bug.cgi?id=6819) but remains open in the OpenWrt fork (https://dev.archive.openwrt.org/ticket/18320.html). Unfortunately, Entware distributes the OpenWrt fork, not the Yocto one. I’ll try compiling the Yocto source to see if it really fixes the issue for our use case.

Eeems:
> 
> 
> > Looks like it might be an opkg bug: https://dev.archive.openwrt.org/ticket/18320.html#comment:1
> 
> The bug is fixed in Yocto’s opkg (https://bugzilla.yoctoproject.org/show_bug.cgi?id=6819) but remains open in the OpenWrt fork (https://dev.archive.openwrt.org/ticket/18320.html). Unfortunately, Entware distributes the OpenWrt fork, not the Yocto one. I’ll try compiling the Yocto source to see if it really fixes the issue for our use case.

We might want to just have the Yocto opkg in our repo until it's fixed.

matteodelabre:
Closing this since it would need to be reimplemented from scratch to work with the new tooling.

PR #138, opened 2020-11-07
[rmkit] update to latest version (11.07.2020)

raisjn:
* add memory stats to remux dialog
* show running app in remux
* fix harmony missing SAVE_DIR (finally)
* fix remux multi-launch issue (no bug, but clutters logs)

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -32,10 +32,6 @@ harmony() {
         install -D -m 755 "$srcdir"/src/build/harmony "$pkgdir"/opt/bin/harmony
         install -D -m 644 "$srcdir"/src/harmony/harmony.draft "$pkgdir"/opt/etc/draft/harmony.draft
     }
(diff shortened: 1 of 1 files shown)


matteodelabre:
Thanks! I think that you need to increase the pkgver of remux and harmony.

> * fix harmony missing SAVE_DIR (finally)

Does this mean that we can remove the `configure()` section on lines 37–39 of the recipe?

raisjn:
good catch, thanks! i'll bump pkgver for harmony and remux

raisjn:
I will have to test, but yes, this should address the package making the SAVE_DIR

matteodelabre:
Just tested and it works, even if I remove the save dir before launching harmony. Good job! I’ll remove the configure section from the package and merge.

matteodelabre (approved):

PR #146, opened 2020-11-28
Update retris to 0.6.1-1

LinusCDE:
Should have fully working input support for the rM 2.

@raisjn, could you test this with the shim?

A binary is available on [the release page](https://github.com/LinusCDE/retris/releases/tag/0.6.0-1). I also added the ipk to my test repo.

### DIFF
--- a/package/retris/package
+++ b/package/retris/package
@@ -5,15 +5,15 @@
 pkgnames=(retris)
 pkgdesc="Tetris game"
 url=https://github.com/LinusCDE/retris
-pkgver=0.6.0-1
-timestamp=2020-11-28T15:23Z
+pkgver=0.6.1-1
+timestamp=2020-12-24T12:32Z
 section=games
 maintainer="Linus K. <linus@cosmos-ink.net>"
 license=MIT
 
 image=rust:v1.2.1
-source=(https://github.com/LinusCDE/retris/archive/0.6.0-1.zip)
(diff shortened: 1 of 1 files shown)


Eeems:
@raisjn and @LinusCDE poke

LinusCDE:
Just waiting on my pr to get into libremarkable, so i can reference that instead.

Not sure if the stable window is still open. If so I can probably just reference my fork fork now and update this (and chessmarkable) tomorrow.

Eeems:
Technically I think you've missed the window already. We should work on getting it merged though.

LinusCDE:
Can do so tomorrow then. Also not sure if there is now a nicer way to start apps with immediate rm2 compatibility. And not sure if I have time to persue that tomorrow.

If it won't fit into the merge window I don't think it'll be a problem though. Didn't even plan on getting this into into the current one.

Eeems:
> Can do so tomorrow then. Also not sure if there is now a nicer way to start apps with immediate rm2 compatibility. And not sure if I have time to persue that tomorrow.
> 
> If it won't fit into the merge window I don't think it'll be a problem though. Didn't even plan on getting this into into the current one.

I believe there is a way to have systemd handle most of it for you. As long as your launcher is passing on the environment it should all work nicely.

LinusCDE:
If someone can review this for the rM2, I'll we might be able to squeeze it into the stable window.

Particulary check the following:
 - Input is mapped correctly on the rM2
 - Can launchers start this without this package specifying the shim?

raisjn (approved):
shipit

PR #150, opened 2020-12-10
[rmkit][rm2] update rmkit to latest version

raisjn:
* apps now support rm2 with rm2fb

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -9,11 +9,11 @@ license=MIT
 
 image=python:v1.1
 source=(
-    https://github.com/rmkit-dev/rmkit/archive/1736446aa9ab6196aa4b366f252cfc7661d11d19.zip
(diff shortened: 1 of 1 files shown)


LinusCDE (dismissed):
All sub-packages compiled and ran fine for me. Not sure if you also a rM2 owner review this as well.

LinusCDE (approved):
Retested remux on my rM1. Seems fine.

Btw: The harmony binary has changed (e.g. not bit-perfect producible). I don't see this as an issue, what's your opionion on this, @matteodelabre ?

raisjn:
thanks for reviewing!

One line in a shared file (`src/rmkit/fb/fb.cpy`) was changed which would explain why harmony has a new hash. I haven't verified that harmony builds with the same signature every time, though, so I'm not sure if we expect it to.

LinusCDE:
Ok. I don't think that we need to retest harmony then. It seemed pretty stable so far.

PR #153, opened 2020-12-18
[remux] update remux with shutdown fix for rM1

raisjn:
the new rM update (2.5) removes rtcwake from the system. this version of remux no longer depends on rtcwake

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -9,11 +9,11 @@ license=MIT
 
 image=python:v1.1
 source=(
-    https://github.com/rmkit-dev/rmkit/archive/fb8359fcac7eb22f481292265a01c01cf73e3279.zip
+    https://github.com/rmkit-dev/rmkit/archive/5ceb9d2faa4e145a8adb9bb32744c7c5a503ba23.zip
     remux.service
 )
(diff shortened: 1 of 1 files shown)


Eeems:
It would be great to start shipping rtcwake as a package since it's got a nicer interface

package/rmkit/package
```diff
@@ -62,7 +62,7 @@ nao() {
 remux() {
     pkgdesc="App launcher that supports multi-tasking applications"
     url="https://rmkit.dev/apps/remux"
-    pkgver=0.1.0-1
+    pkgver=0.1.0-2
```
Eeems: No version number change?

raisjn: i don't have internal version numbers for programs in rmkit,  I'll add that to the todo list

Eeems:
The other sub-packages should have their release number bumped to ensure that they will register as an update.
When trying to launch remux it's turning off my device. https://pastebin.com/n0866iKF

Eeems:
harmony, mines, and nao/simple all seem to work fine though.

Eeems (approved):

PR #154, opened 2020-12-21
[harmony] create saved_images dir in package configure script

raisjn:
when this dir is missing, sometimes harmony doesn't start. this
has been a bug i've been struggling with for a while, the best
solution has been to create it in this package


LinusCDE (approved):
:shipit: Ship it

PR #159, opened 2020-12-23
Update KOReader

Eeems:
Will only happen if `rot` is available

### DIFF
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -36,7 +36,7 @@ configure() {
     # Is Oxide installed?
     if systemctl list-units --full -all | grep -Fq 'tarnish.service'; then
         # Check to see if tarnish is running and rot is available
-        if command -v rot &> /dev/null && systemctl is-active --quiet tarnish; then
+        if command -v jq &> /dev/null && command -v rot &> /dev/null && systemctl is-active --quiet tarnish; then
             sh -c "$_applicationRegistered" && sh -c "$_unregisterApplication"
             result=$(rot apps call registerApplication 'QVariantMap:{"name": "KOReader", "displayName": "KOReader", "decription": "ebook reader", "bin": "/opt/bin/koreader", "icon": "/opt/etc/draft/icons/koreader.png", "onStop": "killall -9 luajit", "onPause": "killall -STOP luajit || true && /opt/koreader/fbdepth -d 16 -r 1", "onResume": "/opt/koreader/fbdepth -d 8 -r 1 || true && killall -CONT luajit"}')
             if [[ "$result" != '"/"' ]] && [[ "$result" != "" ]]; then
@@ -50,7 +50,7 @@ configure() {
 postremove() {
     # Check to see if tarnish is running and rot is available
     if systemctl list-units --full -all | grep -Fq 'tarnish.service' && command -v rot &> /dev/null && systemctl is-active --quiet tarnish; then
-        if sh -c "$_applicationRegistered" && ! sh -c "$_unregisterApplication"; then
+        if ! command -v jq &> /dev/null || sh -c "$_applicationRegistered" && ! sh -c "$_unregisterApplication"; then
             echo "You may need to manually remove the application entry for KOReader in Oxide."
         fi
     fi
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -34,3 +34,10 @@ configure() {
         fi
     fi
 }
+postremove() {
+    if command -v rot &> /dev/null && command -v jq &> /dev/null && systemctl is-active --quiet tarnish; then
+        rot apps get applications | jq -rc '.KOReader' | sed 's|/codes/eeems/oxide1/||' | xargs -I {} rot --object "Application:{}" apps call unregister
+    else
+        echo "You may need to manually remove the application entry for KOReader in Oxide."
+    fi
+}
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -27,17 +27,15 @@ package() {
 
 configure() {
     if systemctl list-units --full -all | grep -Fq "tarnish.service"; then
-        if command -v rot &> /dev/null && systemctl is-active --quiet tarnish; then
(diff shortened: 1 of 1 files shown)


Eeems:
Test scenarios:

- Oxide is not installed
- Oxide is installed but not enabled or running
- Oxide is installed and running (Should automatically register with the correct settings)
- Test on both rM1 and rM2

LinusCDE (changes requested):
**EDIT:** When speaking of suspending/resuming, I meant the KOReader app only and not the sleepscreen. I missed that this could be misinterpreted.

When testing with remux at first, I encountered some glitching when switching the folder pages.

&nbsp;

Some pages were fine, but some glitched like above without a clear pattern on which turn this happens.

It seems only to happen when remux is enabled on boot and regardless of whether oxide is installed. @raisjn 


**Oxide is not installed**

Works. Only when remux is used, the first time launching the files are glitched (even suspending is enough until probably a reboot).

**Oxide is installed but not enabled or running**

Same as above

**Oxide is installed and running (Should automatically register with the correct settings)**

Cool start works fine. When suspending directly to oxide, there is a chance of the screen getting misconfigured (probably mismatch between set size and assumed size of the framebuffer). Launching erode (middle button) then or either before returning to oxide seems to work around this problem. (Taking a screenshot got the expected result though. Maybe just another refresh needed?)



When resuming koreader, I get this view however (both images are updated).


**Test on both rM1 and rM2** => Only rM 1

raisjn:
thanks for the indepth testing! the latest remux from build.rmkit.dev/master/latest has fixes for koreader which might help here

Eeems:
@LinusCDE could you test the suspend/resume with the v2.1 branch of oxide? It looks like there might be timing issues with the onPause/onResume there.

LinusCDE:
> @LinusCDE could you test the suspend/resume with the v2.1 branch of oxide? It looks like there might be timing issues with the onPause/onResume there.

Built the v2.1 branch (`make release`) and copied all binaries (and the /etc/dbus thing) over my existing ones in `/opt/bin` (current toltect testing branch).

I still get the "screen corruption" when leaving directly to oxide (left button). It doesn't seem to happen any second time though when KOReader was not killed.

It also seems that you are not clearing the sceen as much anymore which leaves some residue when launching erode (middle button) and returning.

&nbsp;

So it behaves a bit differently, but still has problems.

---

> thanks for the indepth testing! the latest remux from build.rmkit.dev/master/latest has fixes for koreader which might help here

Couldn't reproduce the bug anymore with the updated binary. :+1:

LinusCDE:
Reinstalling koreader now fixes the issue. The missing onPause/onResume values in tarnish.conf were added and seem to have been the problem. Probably just ensure that people who update koreader get these changes as well and then it should work.

Eeems:
Hmm, well 

> Reinstalling koreader now fixes the issue. The missing onPause/onResume values in tarnish.conf were added and seem to have been the problem. Probably just ensure that people who update koreader get these changes as well and then it should work.

Added code to handle removing the application entry on remove that should handle this in the future. Maybe I'll also add it to the configure.

Eeems:
Still fighting with the code to make it more robust. It would be great if you could share functions.

Eeems:
Okay, so the configure script works properly, but the postremove script doesn't seem to run ever. When I source the package file on the device and run postremove it works properly. @matteodelabre thoughts?

LinusCDE:
~~If you mean bash functions, you can probably just nest them in this case.~~

LinusCDE (changes requested):
It somehow still does not add the onPause and onResume to `tarnish.conf`. Now I fail to get it even forced in by either deleting the `tarnish.conf`, removing & installing koreader, fore-reinstalling koreader or some combinations.

EDIT: Using oxide 2.0.1~beta-1

Eeems:
> It somehow still does not add the onPause and onResume to `tarnish.conf`. Now I fail to get it even forced in by either deleting the `tarnish.conf`, removing & installing koreader, fore-reinstalling koreader or some combinations.
> 
> EDIT: Using oxide 2.0.1~beta-1

With `jq` installed?

LinusCDE:
> > It somehow still does not add the onPause and onResume to `tarnish.conf`. Now I fail to get it even forced in by either deleting the `tarnish.conf`, removing & installing koreader, fore-reinstalling koreader or some combinations.
> > EDIT: Using oxide 2.0.1~beta-1
> 
> With `jq` installed?

Yes. But rot wasn't for this particular test. Maybe it should just be a dependency?

- With rot, entries get added and removed properly.
- Upgrading from 2020.10-2 to this version didn't update the onPause and onResume lines, so oxide was still broken when quitting.
- Removing and installing this version didn't work either.

EDIT: Giving the tarnish service a look makes it seem that it doesn't take notice of the app getting added or removed. Maybe oxide should get updated first?

Eeems:
> 
> Yes. But rot wasn't for this particular test. Maybe it should just be a dependency?
> 
>     * With rot, entries get added and removed properly.
> 
>     * Upgrading from 2020.10-2 to this version didn't update the onPause and onResume lines, so oxide was still broken when quitting.
> 
>     * Removing and installing this version didn't work either.
> 
> 
> EDIT: Giving the tarnish service a look makes it seem that it doesn't take notice of the app getting added or removed. Maybe oxide should get updated first?

rot is a dependency of oxide now, but maybe not for the version you have installed. If rot is making the calls, it will notice that changes, since all rot does is make calls out to tarnish and have it handle it.

raisjn:
i should have brought this up earlier, but why does koreader package need to do special calls for oxide? (is it the onPause/onResume commands?) it seems like this should really be solved by config file (i think you are doing this because i saw config file stuff in oxide commits) + a SIGHUP to tarnish.

Eeems:
> i should have brought this up earlier, but why does koreader package need to do special calls for oxide? (is it the onPause/onResume commands?) it seems like this should really be solved by config file (i think you are doing this because i saw config file stuff in oxide commits) + a SIGHUP to tarnish. 

Because KOReader changes the bitdepth of the display by default, and oxide didn't have pgroup signals yet. v2.1 of oxide has a new file format for applications installed in a package that will solve all of this though.

LinusCDE:
Retested with the testing version of oxide. Same behaviour as before.

Eeems:
> Retested with the testing version of oxide. Same behaviour as before.

What output are you getting during the install?

LinusCDE:
> > Retested with the testing version of oxide. Same behaviour as before.
> 
> What output are you getting during the install?

I get the Removed app and added app lines when removing/installing the app respectively. `journalctl -fu tarnish` doesn't show any lines regarding this lines though.

LinusCDE (approved):
Even though the resize problems still persist, they most likely do so for the older version as well. Fixing this will be a goal for a future PR.

The app registers fine now using the older draft method.

Eeems:
v2.1 will have a proper solution.

PR #161, opened 2020-12-25
Update plato to 0.9.1-11

LinusCDE:
This update makes plato compatible with the rM 2 as long as the shim is used. The "Quit to Xochitl"-Option will also only show up when the actual xochitl.service was running. Not just the potentially freezed xochitl binary. So users will not be able to start xochitl.service through the menu when using a launcher but can continue to use it standalone as well.

@raisjn The state is the basically same as the binary I sent in #rm2 for testing. Feel free to validate this again though.

### DIFF
--- a/package/plato/package
+++ b/package/plato/package
@@ -5,15 +5,15 @@
 pkgnames=(plato)
 pkgdesc="Document reader"
 url=https://github.com/LinusCDE/plato
-pkgver=0.9.1-10
-timestamp=2020-09-26T02:41Z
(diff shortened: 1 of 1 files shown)


raisjn (approved):
is the title right?

LinusCDE:
Oops. Copy&Paste failure. The commit was meant to be "Update plato to 0.9.1-11". I updated the version but not the name of the software. The title of this PR followed suit and I didn't notice it.

I'll leave the commit message but update the title.

PR #162, opened 2020-12-26
[rmkit] update remux with fix for broken wifi on suspend

raisjn:
this update fixes broken wifi on resume from suspend in remux

Test Plan:

* invoke suspend through remux's bottom right corner. 
* wake tablet up by pressing power button
* verify that network comes back up within a couple minutes (look at xochitl's network bar or use `ping` command, etc)

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -31,6 +31,8 @@ harmony() {
     package() {
         install -D -m 755 "$srcdir"/src/build/harmony "$pkgdir"/opt/bin/harmony
         install -D -m 644 "$srcdir"/src/harmony/harmony.draft "$pkgdir"/opt/etc/draft/harmony.draft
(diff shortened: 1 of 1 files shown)


LinusCDE (changes requested):
The wifi seems to work well for me (though the old version also worked well when following your steps).

A thing I guess I also omitted, was that the PR I made needs the icons added here.

The lines of the icons can be found in this diff (only the changes in package() sections): https://github.com/LinusCDE/toltec/compare/beac3d3...ce47601

LinusCDE (approved):

Eeems:
So we need someone to test with a rM2 now right?

LinusCDE:
Would be best, but I think it'll suffice. I'm not sure whether the bug is actually rM 2 exclusive. The build works fine for me, and whether the bug was fixed could've been tested by raisjn anyway.

Eeems:
The suspend issue is rM2 exclusive.

raisjn:
yes, i tested the suspend issue on rM2

PR #164, opened 2020-12-27
Micro text editor

Eeems:


Eeems:
I wonder if it would be better to install the licenses into `/opt` to avoid having to reinstall them, or filling up the root partition?

Eeems:
> I wonder if it would be better to install the licenses into `/opt` to avoid having to reinstall them, or filling up the root partition?

I've moved it to `/opt` and also added back the man pages in case the user installs `man-pages`

Eeems:
Test scenarios:

1. Run `micro` and make sure it works.
2. Install `man-pages` and ensure that you can run `man micro`.

LinusCDE (approved):
Works well :+1:

PR #165, opened 2020-12-28
Stable merge window for week 52 of 2020

raisjn:
Updated Packages:

* chessmarkable -> 0.5.1-1
* oxide -> 2.0.1~beta-2
* plato -> 0.9.1-11
* retris -> 0.6.1-1
* rmkit
  * harmony -> 0.1.0-3
  * mines -> 0.1.0-2
  * nao -> 0.1.0-2
  * remux -> 0.1.3-1
  * simple -> 0.1.0-2

New Packages:

* rm2fb -> 1.0.0

### DIFF
--- a/package/rm2fb/package
+++ b/package/rm2fb/package
@@ -8,7 +8,7 @@ maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
 pkgdesc="Remarkable2 Framebuffer"
 url="https://github.com/ddvk/remarkable2-framebuffer"
-pkgver=1.0.0-1
+pkgver=1.0.0-2
 section=utils
 version=1.0.0
 image=qt:v1.1
@@ -46,11 +46,20 @@ package() {
 configure() {
     systemctl daemon-reload
     systemctl enable rm2fb --now
-    systemctl restart xochitl
+    # Restart xochitl if it's running
+    if systemctl --quiet is-active xochitl; then
+        # Reset the crash count so we don't trigger remarkable-fail
+        echo "0" > /tmp/crashnum
+        systemctl restart xochitl
+    fi
 }
 
 preremove() {
-    echo "make sure to re-enable xochitl with 'systemctl enable xochitl --now'"
-    echo "and to disable / uninstall any launchers like draft, oxide or remux before"
+    echo -n "make sure "
+    if ! systemctl --quiet --is-enabled xochtil; then
+        echo "to re-enable xochitl with 'systemctl enable xochitl --now'"
+        echo -n "and "
+    fi
+    echo "to disable / uninstall any launchers like draft, oxide or remux before"
     echo "rebooting your tablet to complete the uninstallation"
 }
--- a/package/rm2fb/package
+++ b/package/rm2fb/package
@@ -56,7 +56,7 @@ configure() {
 
 preremove() {
     echo -n "make sure "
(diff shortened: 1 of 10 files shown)


raisjn:
the major focus of this PR is rM2 compatibility. rm2fb is a new package and all the other updated packages have had changes related to rM2 compatibility.

package/rm2fb/package
```diff
@@ -0,0 +1,56 @@
...
+    install -D -m 755 "$srcdir"/rm2fb-server "$pkgdir"/opt/bin/
+    install -D -m 755 "$srcdir"/rm2fb-client "$pkgdir"/opt/bin/
+    install -D -m 644 "$srcdir"/rm2fb.conf "$pkgdir"/etc/systemd/system.conf.d/01-rm2fb.conf
+    install -D -m 644 "$srcdir"/rm2fb.service "$pkgdir"/lib/systemd/system/rm2fb.service
+}
+
+configure() {
+    systemctl daemon-reload
+    systemctl enable rm2fb --now
+    systemctl restart xochitl
```
Eeems: hmm, I missed this when rm2fb was initially started. We should only restart it if it's running. Since configure runs on upgrade, we shouldn't be restarting the service if they've disabled it for something else.

package/rm2fb/rm2fb.service
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+[Unit]
+Description=Remarkable2 Framebuffer Server
+Before=xochitl.service
```
Eeems, line 6: This doesn't seem like it'll okay well with launchers that intend for tarnish to stay disabled

Eeems (changes requested):

package/rm2fb/rm2fb.service
raisjn: i don't think `Before=` actually causes xochitl to start, it is just a hint how to enable both services if they are both started at same time. (i think)

`Requires=` will cause xochitl (or the listed service) to start (if i understand systemd properly), but it will start them both at same time (hence the Before/After keywords)

package/rm2fb/package
raisjn: good point. should the fix be put into stable? (i don't think its high pri, since no one will be upgrading the package until we push an update)

package/rm2fb/rm2fb.service
Eeems: If it's a hint, likely it should be marked as before the launchers as well?

package/rm2fb/package
Eeems: I'd prefer to have it in place before going to stable. While it's not likely going to cause issues yet, it does have various unintended side effects that someone might encounter if they are re-enabling entware or reinstalling the package.

Eeems (dismissed):

raisjn:
tested on rM2:

* rm2fb
  * reinstalled
* chessmarkable
* plato
  * disabling inputs works properly
* oxide
  * can launch apps, correctly redraws screen on app exit
* retris
* harmony
* mines
* nao
* remux

all worked as expected with touch and stylus inputs

LinusCDE (dismissed):

Eeems (dismissed):

LinusCDE (approved):

PR #166, opened 2020-12-28
Update plato to 0.9.1-12 (rM2 bugfix)

LinusCDE:
Just noticed, that plato might have a bug on the rM 2, since the `/dev/input/` paths were not changed depending on the model:

- De-selecting "Input by -> Pen" will actually disable the power button (Pen will continue to work).
- De-selecting "Input by -> Touch" will actually disable the Pen (touch will continue to work).

The above bugs should be present on 0.9.1-11 but not on this version (0.9.1-12). The package is available at https://rmtestrepo.cosmos-ink.net/ for faster testing. The new version will behave the same on the rM 1 (verified it quickly).

As per suggesion from @Eeems in discord, I would recommend that I would be nice, if it could get squeezed into the stable window. (I'm not sure how many actually use this feature if any.)

### DIFF
--- a/package/plato/package
+++ b/package/plato/package
@@ -5,15 +5,15 @@
 pkgnames=(plato)
 pkgdesc="Document reader"
 url=https://github.com/LinusCDE/plato
-pkgver=0.9.1-11
-timestamp=2020-12-25T19:13Z
+pkgver=0.9.1-12
+timestamp=2020-12-28T17:53Z
 section=readers
 maintainer="Linus K. <linus@cosmos-ink.net>"
 license=AGPL-3.0-or-later
 
 image=rust:v1.2.1
-source=(https://github.com/LinusCDE/plato/archive/0.9.1-rm-release-5.zip)
-sha256sums=(e4203ffa79630259de495ce2076cce2e857f352333450a24588328ddcd76de6b)
+source=(https://github.com/LinusCDE/plato/archive/0.9.1-rm-release-6.zip)
+sha256sums=(f452189141844beb4ba15439ee440ea861ccfde283e9363ea92518296ea58417)
 
 build() {
     # Get needed build packages


SyntaxBlitz:
Just tested this on my rm2, running through oxide. Looks like it works! Your description of both the old and new behavior on `-11` and `-12` match up with my testing, though the power button seems to freeze (and maybe crash?) the app when I press it on either build.

Eeems:
@LinusCDE Since the crash exists on the previous build I'm approving it. Is the power button trying to just do a suspend?

Eeems (approved):

LinusCDE:
> @LinusCDE Since the crash exists on the previous build I'm approving it. Is the power button trying to just do a suspend?

> *[...]* the power button seems to freeze (and maybe crash?) the app when I press it on either build.

Yes. It does a suspend with it's own lockscreen. Seems to be an additional bug on the rM 2. Maybe there is some fighting with a launcher, who gets to suspend (i.e. the button press leaks through). On my rM 1 Oxide replaces the suspendscreen without a problem.

Eeems:
> > @LinusCDE Since the crash exists on the previous build I'm approving it. Is the power button trying to just do a suspend?
> 
> > _[...]_ the power button seems to freeze (and maybe crash?) the app when I press it on either build.
> 
> Yes. It does a suspend with it's own lockscreen. Seems to be an additional bug on the rM 2. Maybe there is some fighting with a launcher, who gets to suspend (i.e. the button press leaks through). On my rM 1 Oxide replaces the suspendscreen without a problem.

Oxide doesn't pass through the button press ever. The issue on the rM2 is likely the same one we've encountered for oxide where wifi doesn't turn back on properly, and it wasn't waiting long enough to properly draw the custom lockscreen: https://github.com/Eeems/oxide/pull/114

LinusCDE:
Could be. Though if Oxide never passed the button through, I'm wondering why it would crash since it would be just a freeze for plato.

[This](https://github.com/LinusCDE/plato/blob/master/scripts/suspend.sh) is the script that would get executed should the button actually get pressed for plato.

SyntaxBlitz:
Yeah, it wasn't a crash, just a freeze. Indeed with Eeems/oxide#114 installed, Plato behaves just fine when I press the power button now. I think it's just the Oxide bug I was noticing.

PR #167, opened 2020-12-28
Create dotnet-sdk package

Eeems:
- [x] dotnet-sdk
- [x] dotnet-runtime
- [x] runtime-aspnetcore


Eeems:
It would be great to get #137 merged so that we can use `provides=()` for this.

Eeems:
I also had to change how we are creating the package archives to work around https://dev.archive.openwrt.org/ticket/12496

Eeems:
Borrowed a lot from https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=dotnet-core-bin

package/dotnet/dotnet-profile
```diff
@@ -0,0 +1 @@
+export DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1
```
LinusCDE: This does not seem to get applied. Neither when relogging nor rebooting. Needed to run this manually to create a new test project.

package/dotnet/package
```diff
@@ -0,0 +1,120 @@
...
+
+dotnet-host() {
+    pkgdesc="A generic driver for the .NET Core Command Line Interface"
+    depends=(dotnet-profile)
+
+    package() {
+        install -dm 755 "$pkgdir"/opt/usr/{bin,lib,share/{dotnet,licenses/dotnet-host}}
+        cp -dr --no-preserve='ownership' "$srcdir"/{dotnet,host} "$pkgdir"/opt/usr/share/dotnet/
+        cp -dr --no-preserve='ownership' "$srcdir"/{LICENSE,ThirdPartyNotices}.txt "$pkgdir"/opt/usr/share/licenses/dotnet-host
+        ln -sf /opt/usr/share/dotnet/dotnet "$pkgdir"/opt/usr/bin/dotnet
```
LinusCDE: Maybe use `/opt/bin` instead? Our bootstrap still does not add `/opt/usr/bin/` to the path. So most users will not find the dotnet command.

LinusCDE (changes requested):
Created a project with `/opt/usr/bin/dotnet new console -o myApp` which needed the command from the `dotnet-profile` file to be manually run before it worked.

Afterwards, it complained about a missing lib:
```
reMarkable: ~myApp/ /opt/usr/bin/dotnet run
A fatal error occurred. The required library libhostfxr.so could not be found.
If this is a self-contained application, that library should exist in [/home/root/myApp/bin/Debug/netcoreapp3.1/].
If this is a framework-dependent application, install the runtime in the global location [/usr/share/dotnet] or use the DOTNET_ROOT environment variable to specify the runtime location or register the runtime location in [/etc/dotnet/install_location].

The .NET Core runtime can be found at:
  - https://aka.ms/dotnet-core-applaunch?missing_runtime=true&arch=arm&rid=codex.2.5.2-arm
```

Running `export DOTNET_ROOT=/opt/usr/share/dotnet/` fixed that for me. Maybe this env should also be added by default.

After that I could run a hello world.

package/dotnet/dotnet-profile
Eeems: It looks like `/etc/profile` is only sourcing `/etc/profile.d/*.sh` files, whoops. I'll fix that.

LinusCDE (approved):
Works now like a charm. :+1: I did also install `aspnet-runtime` but haven't tested whether any features of it actually work since I'm not familiar with it.

(Sorry for the late re-review. Got caught up in responding to an email).

Eeems:
That should be fine to start I think.

raisjn:
just FYI, this package brings the repo from around 80MB to 200MB (the dotnet packages are pretty large)

Eeems:
Yes they are :/ not sure if there is a way around it. We might want to look at if the opkg file format allows for other compression formats than tar.gz

LinusCDE:
I personally don't mind the size. Alternatively we may be able to make the dotnet-sdk into an installer that just downloads the stuff straight from source when installing similar to some oracle java packages on ubuntu.

Eeems:
> I personally don't mind the size. Alternatively we may be able to make the dotnet-sdk into an installer that just downloads the stuff straight from source when installing similar to some oracle java packages on ubuntu.

Perhaps, but that's more likely to run into issues in the future if they completely change things about where stuff is hosted, so we might have a period of time where the install doesn't work. This way it will always work as long as we host the repo. I'll look into if we can save some space by stripping the binaries (assuming they aren't already).

PR #168, opened 2020-12-29
Update rm2fb

Eeems:
Only restart xochitl if it's running.
Reset crashnum before restarting xochtil.
Modify preremove message based on xochitl's state.


package/rm2fb/package
```diff
@@ -46,11 +46,20 @@ package() {
 configure() {
     systemctl daemon-reload
     systemctl enable rm2fb --now
-    systemctl restart xochitl
+    # Restart xochitl if it's running
+    if systemctl --quiet is-active xochitl; then
```
raisjn, line 50: the exit code is 0 if xochitl is active, so should this be a negated condition?

raisjn (changes requested):
nice usage of systemctl!

i think need to bump version of rm2fb, too

package/rm2fb/package
Eeems: Oh right, forgot about that.

Eeems: Hmm, after a quick test, this actually works properly and will be truthy if systemctl is running.

raisjn (dismissed):

raisjn (approved):

Issue #85, opened 2020-09-29
Add builddepends metadata field to recipes


matteodelabre:
_(As proposed by @Eeems on Discord.)_

Some recipes require specific packages that are not provided by the Toltec Docker images. Currently, these are [calculator](https://github.com/toltec-dev/toltec/blob/9221d4f350fc8c782d8e0f80845f6846cf96d459/package/calculator/package#L24) and [plato](https://github.com/toltec-dev/toltec/blob/a9a45e8e7ddd1e986554334faa4ea88e98b9ad6e/package/plato/package#L16). We should add a `builddepends` field, similar to Arch’s makedepends, which contain be a list of Debian packages required for the `build()` function.

Eeems:
Preferably it would also cache the image with the builddeps installed after the first run so that subsequent builds don't need to reinstall them.

matteodelabre:
I’m starting to implement this. To make things simpler, I’ll omit the `optdepends` and `checkdepends` fields from Arch’s PKGBUILDs, which could still be implemented at a later time.

* **Specifying dependencies.** The `depends` and `makedepends` fields will support entries of the form: `type:name<=>version`, where
  - `type` is one of `apt` (for installing dependencies from Debian, can only be specified in the `makedepends` field) or `opkg` (for installing dependencies from Entware or Toltec),
  - `name` is the package name to install,
  - `<=>` is an optional version comparator from `<`, `<=`, `=`, `=>` or `>`,
  - `version` is an optional version constraint.
* **Building packages.** Before invoking the `prepare()` and `build()` functions, the build script will make sure that all the required Debian, Entware and Toltec packages are installed in the Docker container’s $SYSROOT directory. If a required Toltec package is not yet available, the build of that package is triggered first.
* **Support for split packages.** Only dependencies declared at the top-level of a split package recipe will be enforced before starting the build. A split recipe cannot declare a top-level makedepends on one of the packages it generates.

matteodelabre:
Also, regarding https://github.com/toltec-dev/toltec/pull/170#discussion_r549931040, I’m thinking that the `prepare()` function should be moved to run inside the Docker container. Thoughts?

Eeems:
Perhaps, this will allow us to have access to more tools out of the gate.

PR #170, opened 2020-12-30
Add evtest 1.34-1

LinusCDE:


package/evtest/package
```diff
@@ -0,0 +1,32 @@
...
+maintainer="Linus K. <linus@cosmos-ink.net>"
+license=GPL-2.0-only
+
+image=base:v1.2.1
+source=(https://gitlab.freedesktop.org/libevdev/evtest/-/archive/evtest-1.34/evtest-evtest-1.34.zip)
+sha256sums=(62f7e34c5bab91b5015de5b056d79051c677c5bd5702facb2885f8e4ba0df84c)
+
+build() {
+    apt-get update
+    apt-get install -y autoconf
```
Eeems, line 20: In the absence of `builddepends` (#85) this should be in `prepare`
```suggestion
prepare() {
    apt-get update
    apt-get install -y autoconf
}
build() {
```

Eeems (changes requested):

package/evtest/package
Eeems: Oh right, `prepare` doesn't run in the docker container, and `build` does. So it has to be in build. Ignore this.

Eeems (approved):

PR #171, opened 2020-12-30
[rmkit] update remux to 0.1.4-1

raisjn:
* use genie gesture engine inside remux

to test this update, use remux and swipe up the sides of the screen to verify it still launches

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -9,11 +9,11 @@ license=MIT
 
 image=python:v1.1
 source=(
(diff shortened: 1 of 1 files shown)


Eeems:
Is it suppose to add a new package for genie as well? Or is it embedded?

raisjn:
it's embedded, but genie has a binary that exposes that engine (and unpackaged still)

Eeems (changes requested):
So it launches if I swipe up the right side of the screen. Is it also suppose to do so on the left side?

When I click the suspend button it appears to go back to xochitl and not suspend. Upon pressing any button the device powers off.

Eeems (approved):

Issue #174, opened 2021-01-01
Installation problem: Toltec-bootstrap is not in the stable branch


jonahweissman:
I tried to install toltec with the bootstrap script, but the script failed when it tried to run `opkg install toltec-bootstrap`. From what I can tell, `toltec-bootstrap` is not available on the stable branch, only on testing. I was able to work around this issue by installing with
```
$ toltec_branch=testing bash bootstrap
```
And then changing the line
```
src/gz toltec https://toltec-dev.org/testing
```
in `/opt/etc/opkg.conf` to
```
src/gz toltec https://toltec-dev.org/stable
```
But it would probably be easier for new users if they didn't have to do this.

I'm assuming that `toltec-bootstrap` is not on the stable branch so that it doesn't appear in the package list that users read. Is there a better way to do this?

LinusCDE:
Boostrap forcing the default use of stable and then having these problems should be addressed indeed. We should treat any toltec repo url in `opkg.conf` as the choice of the user no matter what `toltec_branch` is selected (or just if the env isn't specified).

Those things are definitely things we need to iron out before the official [Community Rollout](https://github.com/toltec-dev/toltec/milestone/1).

Any ideas @Eeems ?

@matteodelabre We should probably make the stable branch the default one. That would make versioning the readme also easiyer since it would reflect the state of it's own branch. Even if this might induce a bit more trouble for devs creating their first PR, basing and merging their changes into stable if not careful.

Eeems:
toltec-bootstrap was just merged in to testing the other day and will be in stable as part of the next stable merge window.

We should make sure that the readme points people at the stable version of the script instead of the testing one. That was an oversight of the PR to merge it in.

I do agree though that we should make stable the default branch, although I'm not sure if GitHub likes to play nice with opening PRs by default etc.

LinusCDE:
> I do agree though that we should make stable the default branch, although I'm not sure if GitHub likes to play nice with opening PRs by default etc.

Good point. Also not sure on that one. Switching to stable would probably cause a lot of wrong PRs even from ourselves. :grin:

Eeems:
We could easily add a script to tell us to switch the branch.

matteodelabre:
There is a longstanding request for GitHub to make the default PR branch and the default view branch two separate settings (https://github.com/isaacs/github/issues/1401). Setting the default branch to `testing` makes things easier for devs. Setting it to `stable` makes things easier for users. The second makes more sense for several reasons:

[...]

PR #176, opened 2021-01-02
Patch bootstrap

LinusCDE:
We should maybe collect some more fixes here first for the bootstrap script.

The first commit **should** address issue #174 for the most part.

Read each commits description for the individual changes and feel free to add some to this branch.

### DIFF
--- a/scripts/bootstrap/bootstrap
+++ b/scripts/bootstrap/bootstrap
@@ -31,7 +31,7 @@ set -eE
 get-branch() {
     local opkg_conf=$entware_path/etc/opkg.conf
     # shellcheck disable=SC2002
-    branch=$(cat "$opkg_conf" | grep src/gz | while read -r line; do
+    branch=$(cat "$opkg_conf" | grep "^src/gz" | while read -r line; do 
         # Is this the toltec repo?
         if [[ "$(echo "$line" | awk '{print $2}')" == "toltec" ]]; then
             local url
--- a/scripts/bootstrap/bootstrap
+++ b/scripts/bootstrap/bootstrap
@@ -21,6 +21,7 @@ set -eE
 [[ -z $wget_path ]] && wget_path=/home/root/.local/bin/wget
 
 # Path for the systemd unit that mounts Entware to /opt
+[[ -z $old_systemd_mount_path ]] && old_systemd_mount_path=/etc/systemd/system/opt.mount
 [[ -z $systemd_mount_path ]] && systemd_mount_path=/lib/systemd/system/opt.mount
 
 # Path where the actual Entware distribution resides (will be mounted to /opt)
@@ -124,6 +125,18 @@ wget-bootstrap() {
 entware-mount() {
     mkdir -p /opt
     mkdir -p "$entware_path"
+
+    if [[ ! -f $old_systemd_mount_path ]] && [[ ! -f $systemd_mount_path ]]; then
+        # The user was probably using https://github.com/Evidlo/remarkable_entware
+        # before switching to toltec. Removing old .mount file to prevent duplicate files
+        log "Updating opt.mount location"
+
+        if systemctl -q --is-enabled opt.mount; then
+          # Remove path to old file if still symlinked
+          systemctl disable opt.mount
+        fi
+        rm $old_systemd_mount_path
+    fi
 
     # Create systemd mount unit to mount over /opt on reboot
     cat > "$systemd_mount_path" << UNIT
--- a/scripts/bootstrap/bootstrap
+++ b/scripts/bootstrap/bootstrap
@@ -29,21 +29,21 @@ set -eE
 
 # Find out used branch or get default one
 get-branch() {
-  local opkg_conf=$entware_path/etc/opkg.conf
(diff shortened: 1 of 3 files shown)


scripts/bootstrap/bootstrap
```diff
@@ -26,8 +26,26 @@ set -eE
 # Path where the actual Entware distribution resides (will be mounted to /opt)
 [[ -z $entware_path ]] && entware_path=/home/root/.entware
 
+# Find out used branch or get default one
+get-branch() {
```
LinusCDE, line 31: Is there a more clean way to do this? Didn't found a good way around defining this function in the env defaults section.

LinusCDE:
The changes here should be fine, but I kept it as a draft for now to allow other changes here before settling on applying them.

TODO:
- [x] Update hash when done with changes.

scripts/bootstrap/bootstrap
Eeems: You can probably parse out the branch with something like:
```bash
get-branch(){
  local opkg_conf=$entware_path/etc/opkg.conf
  branch=$(cat "$opkg_conf" | grep src/gz | while read -r line; do
    if [[ "$(echo "$line" | awk '{print $2}')" == "toltec" ]];then
      local url=$(echo "$line" | awk '{print $3}')
      if ! echo $url | grep -q https://toltec-dev.org/; then
        continue
      fi
      echo "$url" | sed 's|https://toltec-dev.org/||'
      return
    fi
  done)
  [[ "$branch" != "" ]] && echo "$branch" || echo stable
    
}
```
This will allow for more than just stable and testing if we ever add a new branch. You could also add some extra logic to it to detect that there is a toltec repo added with a completely wrong URL if you'd like.

scripts/bootstrap/bootstrap
```diff
@@ -21,13 +21,32 @@ set -eE
...
+              # Found branch specified in opkg.conf
+              echo $toltec_branch
+              return
+          fi
+        done
+    fi
+
+    # No opkg.conf or toltec repo found. Using default branch.
+    echo stable
+}
```
Eeems:
```suggestion  
get-branch() {
  local opkg_conf=$entware_path/etc/opkg.conf
  branch=$(cat "$opkg_conf" | grep src/gz | while read -r line; do
    # Is this the toltec repo?
    if [[ "$(echo "$line" | awk '{print $2}')" == "toltec" ]];then
      local url=$(echo "$line" | awk '{print $3}')
      if ! echo $url | grep -q https://toltec-dev.org/; then
        # If it's the wrong domain skip, we don't know how to get the branch
        continue
      fi
      echo "$url" | sed 's|https://toltec-dev.org/||'
      return
    fi
  done)
  # Default to stable if there was no result
  [[ "$branch" != "" ]] && echo "$branch" || echo stable
}
```

Eeems (changes requested):

Eeems:
@LinusCDE Fixed all the formatting warnings and updated the hash. Care to review?

scripts/bootstrap/bootstrap
```diff
@@ -21,13 +21,35 @@ set -eE
...
 [[ -z $systemd_mount_path ]] && systemd_mount_path=/lib/systemd/system/opt.mount
 
 # Path where the actual Entware distribution resides (will be mounted to /opt)
 [[ -z $entware_path ]] && entware_path=/home/root/.entware
 
+# Find out used branch or get default one
+get-branch() {
+    local opkg_conf=$entware_path/etc/opkg.conf
+    # shellcheck disable=SC2002
+    branch=$(cat "$opkg_conf" | grep src/gz | while read -r line; do
```
LinusCDE:
```suggestion
    branch=$(cat "$opkg_conf" | grep "^src/gz" | while read -r line; do 
```

The bootstrap script tought i was on stable because of a commented out line saying so. This ignores comments by checking that it **starts** with "src/gz".

Otherwise it is fine.

LinusCDE:
Thanks for the fix. Just had to regoogle how to make it an actual suggestion. xD

LinusCDE:
Thanks. Also just saw the dangling space. :sweat_smile:

LinusCDE:
Should we now just let this wait for a few days for other potential changes or merge now right away?

Eeems:
Since it's going into testing, no reason not to merge early. I'll let you update the sha256sum.

Eeems (approved):

raisjn:
is this ready for review? (ready to be merged?)

Eeems:
From my perspective it's ready to merge. I was leaving that for @LinusCDE since he hadn't moved it out of draft status yet.

LinusCDE:
> is this ready for review? (ready to be merged?)

> From my perspective it's ready to merge. I was leaving that for @LinusCDE since he hadn't moved it out of draft status yet.

Yes, its basically ready. I wasn't sure whether anyone wanted to make some adjustments here based on #174 . But it thinks it's okey to merge now. Anyone feel free to review or just merge.

Eeems:
I think we can resolve that in a new PR. I'll merge.

PR #177, opened 2021-01-03
Update chessmarkable to 0.6.0-1

LinusCDE:
Adds the first draft (and probably final) to rotate the board in PvP mode when selected in the main menu.

I don't see any reason why this wouldn't work specifically on the rM 2 (libremarkable is updated but should've stayed the same in terms of features).

(https://github.com/LinusCDE/chessmarkable/releases/tag/0.6.0-1)


Eeems:
Seems to work fine for me on my rM1. Do you want someone to test on the rM2 first?

Eeems (approved):

PR #180, opened 2021-01-04
Strip files by default

Eeems:


scripts/package-build
```diff
@@ -148,6 +148,16 @@ for pkgname in "${pkgnames[@]}"; do
...
             package
 
+            section "Stripping $pkgname"
+            docker container run --interactive --rm \
+                --mount type=bind,src="$(realpath "$pkgdir")",dst=/pkg \
+                "$(image-name base)" /bin/bash \
+                << "SCRIPT"
+find /pkg -print0 -type f | xargs --null arm-linux-gnueabihf-strip --strip-all &> /dev/null || true
+SCRIPT
+            find "$pkgdir" -print0 -type f | xargs --null strip --strip-all &> /dev/null || true
```
raisjn, line 158: why is strip happening twice, once in docker and once not?

Eeems: One is using `arm-linux-gnueabihf-strip` to strip files in that format, the other uses `strip` to strip any x86 files also going along. This is necessary as some files are not in a format that `arm-linux-gnueabihf-strip` understands, but still can be stripped.

raisjn (approved):

PR #76, opened 2020-09-27
Update packaging documentation

LinusCDE:
This pr is mainly to track and discuss updates on the documentation until it is ready for release. Feel free to commit any changes you want onto this branch.

PR #181, opened 2021-01-04
[wireguard][1.0.20201221] - New Package

jonahweissman:
Thanks for creating this project, it's very cool!

I've written a recipe to package WireGuard as a kernel module, and I've confirmed that it works on my reMarkable 2 using the [WireGuard demo server](https://www.wireguard.com/quickstart/). I do have a few questions:

1. Does it make sense to have two packages `wireguard-rm1` and `wireguard-rm2`? (I can easily add `wireguard-rm1` if so). I haven't tested this package on a reMarkable 1 yet, but it's likely that it won't work because kernel modules are supposed to be compiled against the version of the kernel that is currently in use. I see that there's been some talk about the best way to do this https://github.com/toltec-dev/toltec/pull/136. In my opinion, it would be ideal if users could just run `opkg install wireguard`, and based on their channel or something, they would get the right version for their device.
2. Is there a better way to have build time dependencies? The way I have it set up right now means that the version of the kernel is not reflected in the package version. Also, it's time consuming to download the entire kernel source every time the package is built.
3. Should I create a new package category for network tools? I believe entware has a separate category for this, but I'm not sure if it would be better to wait until there are other packages that would also belong to that category.
4. I'm a little unclear about what the `timestamp` field is for or exactly what it refers to. I saw the documentation in https://github.com/toltec-dev/toltec/pull/76, but I'm not sure if that's still current.


Eeems:
Let me try my best to answer these questions for you:

1. I think we are currently constrained to two packages right now, but if we implement some of the changes in #137 it might allow us to have both provide `wireguard`. That said it might be better to have the install strips figure out the right files to simplify the end user experience. I think others will need to weigh in.
2. Build time dependencies require you to `apt-get install` right now. You can see https://github.com/toltec-dev/toltec/blob/testing/package/plato/package#L19-L26 for an example. In the future #85 will be implemented to make this easier.
3. I'm not against it, but for now the `utils` section works until we have more packages that need to be moved out.
4. It's the date of the original release of the source software. So it should be set to the kernel source release date.

jonahweissman:
@Eeems Thank you for your answers! I think you're right that including the files for both and only installing the right ones for the device the script is running on would be easiest for users. I'll try implementing that.

LinusCDE:
If I get the time, I'll most likely take the time to make it work on the rM 1 in some days. I love wireguard! Great work!!!

Eeems:
We are also having some more discussions about rm1 vs rm2 packages in Discord to provide a better longer term solution.

package/wireguard-rm2/package
```diff
@@ -0,0 +1,50 @@
...
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(wireguard-rm2)
+pkgdesc="Fast, modern, secure VPN tunnel"
+url=https://www.wireguard.com
+pkgver=1.0.20201221
+timestamp=2020-12-21T11:54Z
+section=utils
+maintainer="Jonah Weissman <jonahrweissman+toltec@gmail.com>"
```
Eeems: @jonahweissman https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#maintainer You okay with being the one in charge of reviewing alongside one of us with merge power? Or would you prefer for one of us to own that responsibility for now?

jonahweissman: Yeah! I'm okay with reviewing, if that's okay with you. Now that I've spent all this time getting it to work in the first place, it would feel good to be able to continue to use my skills to help out.

Eeems: If you aren't already I'd highly recommend joining the community discord and poking Ameb for the developer role. We do a lot of talking in there.

jonahweissman:
I've modified the script so that it builds and packages kernel modules for both reMarkable 1 and 2. Since the two devices look for kernel modules in different folders (based on their kernel version), I just had the script install both sets of kernel modules. The `.ko` files are pretty small, so I don't think it's a big issue, but I could install only the necessary files if that'd be better. I've confirmed that this works on the reMarkable 2. Could someone do the same for reMarkable 1? I've been testing by connecting to the [demo server](https://www.wireguard.com/quickstart).

I am now cloning the entire 1.9 GB remarkable linux repo, which slows down the build of the repo a fair bit. Not sure if this is a problem.

package/wireguard/package
```diff
@@ -0,0 +1,71 @@
...
+
+image=base:v1.2.1
+source=(https://git.zx2c4.com/wireguard-linux-compat/snapshot/wireguard-linux-compat-"$_wireguardver".zip)
+sha256sums=(c74cedb10c9b830d937b5575e4b7d2625a8534d2c06a80368e5530db23e69d3b)
+
+build() {
+    apt-get update
+    apt-get install -y bc lzop git
+    git init
+    git remote add -f origin "$_kernelrepo"
```
Eeems: Probably better to mimic what was done here: https://github.com/toltec-dev/toltec/blob/a3db67b0c74e0d99edbdeb5fdb06aea45b21ae87/package/fbink/package#L18

Eeems (changes requested):

Eeems:
I'm getting errors when I try to use it on the rM1.
```shell
reMarkable: ~/ cd tmp/
reMarkable: ~/tmp/ opkg install wireguard_1.0.20201221-1_armv7-3.2.ipk 
Installing wireguard (1.0.20201221-1) to root...
Installing libnl-tiny (2020-08-05-c291088f-1) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/libnl-tiny_2020-08-05-c291088f-1_armv7-3.2.ipk
Installing ip-full (4.4.0-10a) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/ip-full_4.4.0-10a_armv7-3.2.ipk
Installing wireguard-tools (1.0.20200827-2) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/wireguard-tools_1.0.20200827-2_armv7-3.2.ipk
Configuring libnl-tiny.
Configuring ip-full.
Configuring wireguard-tools.
Configuring wireguard.
reMarkable: ~/tmp/ ip link add dev wg0 type wireguard
RTNETLINK answers: Operation not supported
reMarkable: ~/tmp/ nano client.sh
reMarkable: ~/tmp/ chmod +x client.sh 
reMarkable: ~/tmp/ ./client.sh 
RTNETLINK answers: Operation not supported
reMarkable: ~/tmp/ modprobe wireguard
modprobe: ERROR: could not insert 'wireguard': Exec format error
reMarkable: ~/tmp/ opkg remove wireguard
Removing package wireguard from root...
reMarkable: ~/tmp/ opkg install wireguard_1.0.20201221-1_armv7-3.2.ipk 
Installing wireguard (1.0.20201221-1) to root...
Configuring wireguard.
reMarkable: ~/tmp/ rmmod wireguard-tools
rmmod: ERROR: Module wireguard_tools is not currently loaded
reMarkable: ~/tmp/ rmmod wireguard      
rmmod: ERROR: Module wireguard is not currently loaded
reMarkable: ~/tmp/ rmmod ip-full
rmmod: ERROR: Module ip_full is not currently loaded
reMarkable: ~/tmp/ opkg install wireguard_1.0.20201221-1_armv7-3.2.ipk 
Package wireguard (1.0.20201221-1) installed in root is up to date.
reMarkable: ~/tmp/ ./client.sh 
RTNETLINK answers: Operation not supported
reMarkable: ~/tmp/
```

Should there be a preremove or postremove step to `rmmod` anything? Or preupgrade/postupgrade  to handle removing the old version on upgrade?

package/wireguard/package
jonahweissman: That's a good idea to do `--depth=1`, thanks!

Is it okay to track a branch rather than a specific revision? I was worried that it might make the build less reproducible.

The reason I didn't just do a git clone was that I wanted the files to appear directly in `/src`, so that I wouldn't have to make as many modifications to my existing recipe, but I realize now that I could do `git clone URL $(pwd)`.

Eeems: I'd suggest using a specific commit to ensure that the build is reproducible.

jonahweissman:
> ```shell
> reMarkable: ~/tmp/ ip link add dev wg0 type wireguard
> RTNETLINK answers: Operation not supported
> ```

This is the same error that I get when the kernel module isn't installed. I'm assuming you're using the stock kernel?

Could you try running `uname -r`? I believe that the kernel looks in `/lib/modules/$(uname -r)`, so if I got the name wrong, it wouldn't be able to find the modules. 

> Should there be a preremove or postremove step to `rmmod` anything? Or preupgrade/postupgrade to handle removing the old version on upgrade?

That would certainly make sense. When I discovered that it automatically found the modules without having to `insmod` or `modprobe` them, I sort of forgot about the case of upgrading. You're probably right, I'll double-check that it's necessary and then figure something out.

Eeems:
Yes I'm on the stock kernel: `4.9.84-zero-gravitas`
image

Looks like it's in the right place, and finding the module, but it's not liking it:
```shell
reMarkable: ~/tmp/ modprobe wireguard
modprobe: ERROR: could not insert 'wireguard': Exec format error
```

jonahweissman:
> Looks like it's in the right place, and finding the module, but it's not liking it:

Hmm, this is mysterious. I'm assuming that the general build process is correct (since it works for reMarkable 2) and that I'm just compiling against the wrong version/configuration of the kernel, but I may be wrong. My instinct is that the config is somehow different from the config that was actually used to compile the kernel that runs on the device. On reMarkable 2, the current config is stored at `/proc/config.gz`, but I don't think finding the config is that easy on reMarkable 1. It might be possible by running [a script from the kernel source](https://superuser.com/questions/287371/obtain-kernel-config-from-currently-running-linux-system/1127603#1127603).

(Part of the reason I think there might be something wrong with the config is that I get this warning when I make it:
```
warning: (SOC_IMX6SX && SOC_IMX6UL && SOC_IMX7D && SOC_IMX6SLL) selects KEYBOARD_SNVS_PWRKEY which has unmet direct dependencies (!UML && INPUT && INPUT_KEYBOARD && (SOC_IMX6SX || SOC_IMX6UL || SOC_IMX7) && OF)
```
)

I'm a little busy this week, so I probably won't get a chance to take a good look at this until the weekend.

Eeems:
Can confirm it's not as easy on the rM1 
image

raisjn:
cc @ddvk for help with kernel config

ddvk:
- rm1 and 2 have different kernel versions. the module should match the kernel version. what is dmesg showing after failed module

koenvervloesem:
I don't have a reMarkable 1, but on some systems you have to load the `configs` module first with `modprobe configs` before you can access `/proc/config.gz`, so it's worth a shot.

Eeems:
`[ 5022.738434] wireguard: module has no symbols (stripped?)`
Maybe we need to add an option to avoid stripping symbols for specific packages.

Eeems:
Okay, new errors:
`[ 6270.915263] udp_tunnel: version magic '4.9.84-zero-gravitas-g1774e2a6a091 ARMv7 p2v8 ' should be '4.9.84-zero-gravitas ARMv7 p2v8 '`

ddvk:
that can be fixe locally with

touch .scmversion

Eeems:
> 
> that can be fixe locally with touch .scmversion

@ddvk In what directory?

ddvk:
kernel root

scripts/package-build
```diff
@@ -148,14 +148,16 @@ for pkgname in "${pkgnames[@]}"; do
             # Run packaging script
             package
 
-            section "Stripping $pkgname"
-            docker container run --interactive --rm \
-                --mount type=bind,src="$(realpath "$pkgdir")",dst=/pkg \
-                "$(image-name "$image")" /bin/bash \
-                << "SCRIPT"
+            if ! printf '%s\n' "${flags[@]}" | grep -qP '^nostrip$'; then
```
matteodelabre: I suggest using [the `has-element` helper](https://github.com/toltec-dev/toltec/blob/5a206fef9b369597eb348ed8dcd3d3ea6f9466f4/scripts/package-lib#L182) for this test.

Eeems:
> 
> kernel root

`/lib/modules/4.9.84-zero-gravitas/`? Didn't help

ddvk:
no,before compiling the module

Eeems:
> 
> no,before compiling the module

Ah, that's what you meant.

Eeems:
Okay, that seems to have fixed it. Having some trouble connecting to the demo server though.

`Invalid argument: endpoint` when trying to run `wg set wg0 private-key <(echo "$privatekey") peer "$server_pubkey" allowed-ips 0.0.0.0/0 endpoint` in the sample client.sh

jonahweissman:
> Should there be a preremove or postremove step to rmmod anything? Or preupgrade/postupgrade to handle removing the old version on upgrade?

I just tested it to see if this would be necessary, and, at least on reMarkable 2, it's not:

```shell
reMarkable: ~/ opkg remove wireguard
Removing package wireguard from root...
reMarkable: ~/ opkg install .cache/opkg/wireguard_1.0.20201112-1_armv7-3.2.ipk 
Installing wireguard (1.0.20201112-1) to root...
Configuring wireguard.
reMarkable: ~/ modinfo wireguard | grep version
version:        1.0.20201112
srcversion:     01C1BF40CC89DEB605DD211
vermagic:       4.14.78 SMP preempt mod_unload modversions ARMv7 p2v8 
reMarkable: ~/ opkg install .cache/opkg/wireguard_1.0.20201221-1_armv7-3.2.ipk 
Upgrading wireguard on root from 1.0.20201112-1 to 1.0.20201221-1...
Configuring wireguard.
reMarkable: ~/ modinfo wireguard | grep version
version:        1.0.20201221
srcversion:     252D23B121293A056D7EF6B
vermagic:       4.14.78 SMP preempt mod_unload modversions ARMv7 p2v8 
reMarkable: ~/ 
```

jonahweissman:
> `Invalid argument: endpoint` when trying to run `wg set wg0 private-key <(echo "$privatekey") peer "$server_pubkey" allowed-ips 0.0.0.0/0 endpoint` in the sample client.sh

That error makes it seem like the line might be getting truncated at 100 characters. The full line is `wg set wg0 private-key <(echo "$privatekey") peer "$server_pubkey" allowed-ips 0.0.0.0/0 endpoint "demo.wireguard.com:$server_port" persistent-keepalive 25`. I tried running `wg set wg0 endpoint` and got the same error. Could you try checking `client.sh` to make sure that the whole line is intact?

package/wireguard/package
jonahweissman: Is this better?
https://github.com/toltec-dev/toltec/blob/27e2c318250d535b7d1de2644ce4f297405321dc/package/wireguard/package#L43-L44

It reduces the size of `.git/` from 1.9GB to 350MB

Eeems:
Huh, it looks like it was truncating the script when I downloaded it. Tested again and it works: 
image

As for removing it, does the module stay loaded when you do? And on an upgrade would it fail to load properly if the other module is still loaded?

jonahweissman:
> As for removing it, does the module stay loaded when you do? And on an upgrade would it fail to load properly if the other module is still loaded?

It does seem to stay loaded after it is uninstalled, but the new version shows up correctly when it is installed:
```shell
reMarkable: ~/ opkg install .cache/opkg/wireguard_1.0.20201112-1_armv7-3.2.ipk 
Installing wireguard (1.0.20201112-1) to root...
Configuring wireguard.
reMarkable: ~/ opkg remove wireguard
Removing package wireguard from root...
reMarkable: ~/ ./client.sh 
reMarkable: ~/ ping 192.168.4.1
PING 192.168.4.1 (192.168.4.1): 56 data bytes
64 bytes from 192.168.4.1: seq=0 ttl=64 time=95.160 ms
^C
--- 192.168.4.1 ping statistics ---
2 packets transmitted, 1 packets received, 50% packet loss
round-trip min/avg/max = 95.160/95.160/95.160 ms
reMarkable: ~/ modinfo wireguard | grep version:
modinfo: ERROR: could not get modinfo from 'wireguard': No such file or directory
reMarkable: ~/ opkg install .cache/opkg/wireguard_1.0.20201221-1_armv7-3.2.ipk 
Installing wireguard (1.0.20201221-1) to root...
Configuring wireguard.
reMarkable: ~/ modinfo wireguard | grep version:
version:        1.0.20201221
srcversion:     252D23B121293A056D7EF6B
reMarkable: ~/ 
```
From what I can tell by looking at `dmesg` output, the module does not get loaded until it's needed, but when it is loaded, it loads from the file. I think we should do a `modprobe -r wireguard` as a postremove because it doesn't make sense to have the module still loaded if the user thinks it is gone. It probably makes sense to also do this as a postupgrade. I'm guessing it would only matter for someone who has open wireguard connections while they are removing or upgrading, but I don't think there's any downside to making sure the correct version is running at all times.

package/wireguard/package
```diff
@@ -71,3 +71,11 @@ package() {
 configure() {
     depmod -a
 }
+
+postremove() {
+    modprobe -r wireguard
```
Eeems: Do we care if this fails due to the module being in use?

jonahweissman: Good point! I think it's fine if removal fails if the module is in use, as long as the error displayed makes this clear. However, it would probably be better if this were a preremove script, instead, so that the files are not removed in case of failure. (I'm assuming that a preremove script failing would stop removal.)

jonahweissman: I tried to test what would happen if I run `modprobe -r wireguard` while pinging a host through a wireguard interface. The packets just go unanswered, and `modprobe` does not throw any module in use errors. My guess is that the module is only in use while it is creating interfaces.

jonahweissman: (The reason I originally had the module removal in a postremove script was that it would prevent the following scenario:
- preremove script unloads the module
- some other script loads the module
- the module files are removed
which would incorrectly leave the module loaded. This seems very unlikely, preremove is probably fine)

Eeems: > I think it's fine if removal fails if the module is in use, as long as the error displayed makes this clear. 

In which case we should probably output a slightly nicer error about what's happening, and include the source error message.

> (I'm assuming that a preremove script failing would stop removal.)

I think there is a bug where it doesn't?

> I tried to test what would happen if I run modprobe -r wireguard while pinging a host through a wireguard interface. The packets just go unanswered, and modprobe does not throw any module in use errors. My guess is that the module is only in use while it is creating interfaces.

image
Looks like it fails silently :(

package/wireguard/package
Eeems: Looks good to me!

package/wireguard/package
jonahweissman: Did you get the "module in use" error by just having another terminal with `ping` going? I couldn't reproduce your output.

I was hoping that `modprobe -r` would have different exit codes for different errors, so that we could have it only fail if the module is in use, but [it doesn't look like it](https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/tools/modprobe.c#n954). Looking at the source code, I'm not sure at all why modprobe is failing silently. Oh well, I'll keep looking :)

Eeems: I ran client.sh and left everything configured.

jonahweissman: Hmm, I still can't reproduce that behavior. I can get a "module in use" error if I try to remove `udp_tunnel` because it's a dependency of wireguard. But I can't seem to get a "module in use" error for wireguard itself, no matter what I try. (My `modprobe --version` is
```
kmod version 25
-XZ +ZLIB -EXPERIMENTAL
```
Perhaps reMarkable 1 has a different version? Maybe this is a matter of differently configured kernels?)

I think our best option is to replace the current postremove `modprobe -r` with a preremove and preupgrade of 
```
rmmod wireguard
rmmod udp_tunnel || true
rmmod ip6_udp_tunnel || true
```
(The `|| true` is because we don't care if the dependencies fail to remove due to being used by other modules.)

Does this seem reasonable? Is rmmod's message of "Module wireguard is in use" clear enough? This feels like it would be a common problem for packaging kernel modules, but maybe everyone else just uses DKMS.

Eeems: rM1 has the same version and flags of `modprobe`.

Usually I'd say that it's a good enough message, but since this is a package and it's an error expect might happen we should give a more verbose explanation of what's going on. Maybe something like:
```
Failed to unload existing wireguard module as it is in use.
Likely you have a wireguard connection configured and will need to manually remove it before running the following:

rmmod wireguard udp_tunnel ip6_udp_tunnel
depmod -a

If it continues to error you may want to reboot your device.

```

jonahweissman: I just pushed a version with a more detailed error message. Sorry it's taken me so long to get back to you. I really appreciate you taking the time to think through edge cases and make this package better.

In the error message, I recommended using `opkg remove` (which will call the `modprobe -r` in the preremove script) instead of `rmmod`, but if you've gotten it to work with the latter but not the former, I'm happy to recommend `rmmod` instead. (I just thought it would be simpler for the user.) 

I put depmod in a postremove, to get wireguard out of the modules.dep file. It's called in the configure script, so I don't think that calling it manually would help with removing the module after a module in use error.

jonahweissman: @Eeems No rush on reviewing this, but I wanted to make sure you didn't miss it :)

Eeems: image
Hmm, I guess maybe we shouldn't try to remove and just let the user know they'll have to reboot for it to be fully removed?

Eeems (approved):

package/wireguard/package
jonahweissman: I think that's a good idea. Let me know if the postremove/postupgrade message I added could use more details/explanation.

PR #182, opened 2021-01-04
Create zoneinfo-utils package

### LINKED ISSUE #179: Add timezone info package
The reMarkable doesn't come with the full set of timezone information. So timezones listed by `timedatectl` as on the device are not actually selectable. We should add a package that contains the full set and likely install it by default.

Eeems:
Fix #179


package/tzdata/package
```diff
@@ -0,0 +1,92 @@
...
+
+    # install license
+    install -Dm644 LICENSE "${pkgdir}"/opt/usr/share/licenses/tzdata/LICENSE
+
+    popd
+
+    # cleanup
+    rm "${pkgdir}/etc/localtime"
+    rmdir "${pkgdir}/etc"
+    mv "${pkgdir}"/usr/{bin,lib,sbin,share/man} "${pkgdir}"/opt/usr
```
LinusCDE: `/opt/usr/bin` is still no in our path at the moment. (We should really fix that.)

LinusCDE (changes requested):
`/opt/usr/bin/tzselect` runs well and shows my proper time. I still need to figure out, why setting the TZ env and relogging doesn't make it permanent.

Other then `tzselect` and `zdump` not being in the path, it seems great.

Eeems:
> `/opt/usr/bin/tzselect` runs well and shows my proper time. I still need to figure out, why setting the TZ env and relogging doesn't make it permanent.
> 
> Other then `tzselect` and `zdump` not being in the path, it seems great.

Have you tried `timedatectl`?

I guess I can open up a PR to get `/opt/usr/bin` added to the path. I wonder if we should just have a package for it that drops a file in `/etc/profile.d`

Eeems:
I wouldn't mind a quick test by someone with an rM2. @danshick or @raisjn?

package/tzdata/package
Eeems: I'll just move them to `/opt/bin` for now

danshick:
> I wouldn't mind a quick test by someone with an rM2. @danshick or @raisjn?

`tzselect` is working for me but `zdump` gives me

```
-sh: /opt/bin/zdump: cannot execute binary file: Exec format error
```

and file seems to indicate it is x86-64

```
reMarkable: ~/ file /opt/bin/zdump
/opt/bin/zdump: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=8cb1be21845976a1dfb7965b0e65d3e125ac680e, stripped
```

I cloned toltec, checked out Eeems-patch-4, manually built with `make tzdata` and copied the ipk to my remarkable to install with opkg. Maybe I'm doing something wrong because this should be broken for rm1 too.

Eeems:
I can confirm that I'm also getting this on my rM1. Very odd as to why it's a x86-64 binary instead of arm. Time to poke around in the Makefile.

Eeems:
Found the problem. I had to coax the Makefile to actually use the cross compile toolchain.

Eeems:
@danshick and @LinusCDE could you test again?

Eeems:
Actually scratch that. It builds on my machine, but it doesn't like building as part of the actions. Time to look into it further.

Eeems:
That should fix it.

danshick:
> That should fix it.

Works here.

LinusCDE (dismissed):
> tzselect is working for me but zdump gives me
>
> ```
> -sh: /opt/bin/zdump: cannot execute binary file: Exec format error
> ```

Thanks for catching that! I totally missed that and assumed that all binaries would be equally working. I will never assume that in the future, I guess.

Everything works now. :shipit:

LinusCDE:
Though, what reason does 4f9220be have in here?

Eeems:
Trying to sort out why it was giving me errors past builds about the control folder not existing.

LinusCDE (commented):
The package seems to work as supposed, but `tzselect` cant find my timezone. Seems remarkable only includes a few ones or it got somehow removed during earlier testing.

```
reMarkable: /home/root/ tzselect 
Please identify a location so that time zone rules can be set correctly.
Please select a continent, ocean, "coord", or "TZ".
...
#? 7
...
#? 15
Please select one of the following timezones.
1) Swiss time
2) Germany (most areas)
#? 2
/opt/bin/tzselect: line 493: /usr/share/zoneinfo/Europe/Berlin: No such file or directory
/opt/bin/tzselect: time zone files are not set up correctly
```

If this is an intended problem, I probably wont mind, but seems to be a problem worth looking at.

Eeems:
@LinusCDE See #213

LinusCDE:
Faking the change the mentioned issue makes tzselect fail immediately:
```
reMarkable: /home/root/ mount --bind /opt/share/zoneinfo/ /usr/share/zoneinfo
reMarkable: /home/root/ tzselect 
/opt/bin/tzselect: line 179: /usr/share/zoneinfo/iso3166.tab: No such file or directory
/opt/bin/tzselect: time zone files are not set up correctly
```

Eeems:
@LinusCDE Do you have any of the zoneinfo-* packages installed?

Yes. I have `zoneinfo-europe` and `zoneinfo-asia` according to `opkg list-installed`. Trying to remove any of the packages fails since they are required by the base package `entware-opt`.

The zones are there, but the error above seems to miss some metadata file.

[tree-of-opt_share_zoneinfo.txt](https://github.com/toltec-dev/toltec/files/5804527/tree-of-opt_share_zoneinfo.txt)

Eeems:
https://github.com/toltec-dev/toltec/pull/182/commits/4cded4fcc460a15e99ee09077cafe4a4edb82995#diff-fbdcf7783cf5416bc6d0ff5665f9f2f095c97799c80203d3c74bb0d711d63b81L68-L73 might be what is needed.

matteodelabre:
> > Do you have any of the zoneinfo-* packages installed?
> 
> Yes. I have `zoneinfo-europe` and `zoneinfo-asia` according to `opkg list-installed`. Trying to remove any of the packages fails since they are required by the base package `entware-opt`.
> 
> The zones are there, but the error above seems to miss some metadata file.
> 
> [tree-of-opt_share_zoneinfo.txt](https://github.com/toltec-dev/toltec/files/5804527/tree-of-opt_share_zoneinfo.txt)

You may also need `zoneinfo-core`, which is also required to make `timedatectl` work.

LinusCDE:
Still doesn't seem to work.

@Eeems  I tried the lines you mentioned in a bash script with some fixes and they didn't seem to work (still have mounted `/opt/share/zoneinfo` to `/usr/share/zoneinfo`).

Maybe add the commands to `package()` to have the files included as well? I'll give this tomorrow a more thorough testing (probably just installing rm2fb to roll my device back :laughing: and do a fresh toltec/entware install from no files as well.)

Eeems:
@LinusCDE with `zoneinf-core` installed, what is in your `/usr/share/zoneinfo`?

LinusCDE:
Seems to be fairly similar.

[zones.txt](https://github.com/toltec-dev/toltec/files/5808001/zones.txt)
 (is also the tree of `/usr/share/zoneinfo` since bind-mounted anyways).

I'll try to do a rollback now though.

matteodelabre:
> Seems to be fairly similar.
> 
> [zones.txt](https://github.com/toltec-dev/toltec/files/5808001/zones.txt)
>  (is also the tree of `/usr/share/zoneinfo` since bind-mounted anyways).

Notice the added `zone.tab` at the end of the list. Not sure why tzselect wants a file called `iso3166.tab` though.

LinusCDE:
Still have struggle to make it work.

Steps I did:
- Force device to roll back (to 2.4...) and update back to the latest version
- Removing /home/root/.entware to have a totally clean install as far as toltec and mods are concerned
- Running the current bootstrap from #213
- Installed toltec sucessfully and also added the new /usr/share/zoneinfo mount that persisted successfully over the next reboot.
- Installed the most likely latest build of this packacke from my rest repo
- Running `tzselect`. Got the same error as before
- Installing `zoneinfo-core` and rerunning `tzselect` same result

<details>
<summary>Log of adding package on a fresh rootfs + bootstrap from #213 install</summary>

```
reMarkable: /home/root/ opkg list-installed
ca-certificates - 20200601-1
entware-opt - 227000-3
entware-release - 1.0-2
entware-upgrade - 1.0-1
findutils - 4.7.0-2
grep - 3.6-1
libc - 2.27-11
libgcc - 8.4.0-11
libopenssl - 1.1.1h-1
libpcre - 8.44-2
libpthread - 2.27-11
librt - 2.27-11
libssp - 8.4.0-11
libstdcpp - 8.4.0-11
locales - 2.27-9
opkg - 2020-11-24-61b3c62d-1
terminfo - 6.2-1
toltec-bootstrap - 0.0.0-1
wget - 1.20.3-4
zlib - 1.2.11-3
zoneinfo-asia - 2020d-1
zoneinfo-europe - 2020d-1
reMarkable: /home/root/ opkg install zoneinfo-utils
Unknown package 'zoneinfo-utils'.
Collected errors:
 * opkg_install_cmd: Cannot install package zoneinfo-utils.
reMarkable: /home/root/ nano /opt/etc/opkg.conf
reMarkable: /home/root/ opkg update
Downloading https://bin.entware.net/armv7sf-k3.2/Packages.gz
Updated list of available packages in /opt/var/opkg-lists/entware
Downloading https://toltec-dev.org/stable/Packages.gz
Updated list of available packages in /opt/var/opkg-lists/toltec
Downloading https://rmtestrepo.cosmos-ink.net/Packages.gz
Updated list of available packages in /opt/var/opkg-lists/rmtestrepo
reMarkable: /home/root/ opkg install zoneinfo-utils
Installing zoneinfo-utils (2020f-1) to root...
Downloading https://rmtestrepo.cosmos-ink.net/zoneinfo-utils_2020f-1_armv7-3.2.ipk
Configuring zoneinfo-utils.
reMarkable: /home/root/ tzselect
/opt/bin/tzselect: line 179: /usr/share/zoneinfo/iso3166.tab: No such file or directory
/opt/bin/tzselect: time zone files are not set up correctly
reMarkable: /home/root/ opkg install zoneinfo-core
Installing zoneinfo-core (2020d-1) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/zoneinfo-core_2020d-1_armv7-3.2.ipk
Configuring zoneinfo-core.
reMarkable: /home/root/ tzselect
/opt/bin/tzselect: line 179: /usr/share/zoneinfo/iso3166.tab: No such file or directory
/opt/bin/tzselect: time zone files are not set up correctly
reMarkable: /home/root/ df -h
Filesystem                Size      Used Available Use% Mounted on
/dev/root               223.0M    174.4M     32.8M  84% /
devtmpfs                 88.6M         0     88.6M   0% /dev
tmpfs                   249.1M         0    249.1M   0% /dev/shm
tmpfs                   249.1M    400.0K    248.7M   0% /run
tmpfs                   249.1M         0    249.1M   0% /sys/fs/cgroup
tmpfs                   249.1M      4.0K    249.1M   0% /tmp
tmpfs                   249.1M         0    249.1M   0% /var/volatile
/dev/mmcblk1p1           19.0M      3.7M     15.3M  20% /var/lib/uboot
/dev/mmcblk1p7            6.5G      5.2G    953.1M  85% /home
/dev/mmcblk1p7            6.5G      5.2G    953.1M  85% /opt
/dev/mmcblk1p7            6.5G      5.2G    953.1M  85% /usr/share/zoneinfo
reMarkable: /home/root/
```

</details>

So then there is probably some missing change in this package? Not sure.

scripts/package-build
```diff
@@ -44,10 +44,10 @@ readonly recipedir="$1"
 load-recipe-header "$1"
 shift
 
-readonly workdir="$1"
-[[ -d $workdir ]] && files="$(ls -qA -- "$workdir")" && [[ -n $files ]] \
-    && error "Working directory '$workdir' exists and is not empty"
-mkdir -p "$workdir"
+[[ -d "$1" ]] && files="$(ls -qA -- "$1")" && [[ -n $files ]] \
```
matteodelabre: Nit: The `[[ … ]]` Bash syntax does not need variable quoting.
```suggestion
[[ -d $1 ]] && files="$(ls -qA -- "$1")" && [[ -n $files ]] \
```

package/zoneinfo-utils/package
```diff
@@ -0,0 +1,76 @@
...
+pkgver=2020f-1
+timestamp=2020-05-04T06:16Z
+section=utils
+maintainer="Eeems <eeems@eeems.email>"
+license="custom: public domain"
+depends=()
+image=base:v1.2.1
+
+_tzcode=2020f
+_tzdata=2020f
```
matteodelabre, line 19: Suggestion for reducing duplication.
```suggestion
_tzcode=2020f
_tzdata="$_tzcode"
pkgver="$_tzcode-1"
timestamp=2020-05-04T06:16Z
section=utils
maintainer="Eeems <eeems@eeems.email>"
license="custom: public domain"
depends=()
image=base:v1.2.1

```

matteodelabre: Suggestion for reducing duplication of version numbers:
```suggestion
pkgnames=(zoneinfo-utils)
pkgdesc="Utilities for interacting with zoneinfo files"
url=https://www.iana.org/time-zones
_tzver=2020f
pkgver="$_tzver"-1
timestamp=2020-05-04T06:16Z
section=utils
maintainer="Eeems <eeems@eeems.email>"
license="custom: public domain"
depends=()
image=base:v1.2.1
```
And then use `$_tzver` instead of `$_tzcode` and `$_tzdata` below.

package/zoneinfo-utils/package
```diff
@@ -0,0 +1,76 @@
...
+
+# shellcheck disable=SC2206
+source=(
+    https://www.iana.org/time-zones/repository/releases/tzcode${_tzcode}.tar.gz{,.asc}
+    https://www.iana.org/time-zones/repository/releases/tzdata${_tzdata}.tar.gz{,.asc}
+)
```
matteodelabre: No need to disable the shellcheck warning.
```suggestion
source=(
    "https://www.iana.org/time-zones/repository/releases/tzcode${_tzcode}.tar.gz"{,.asc}
    "https://www.iana.org/time-zones/repository/releases/tzdata${_tzdata}.tar.gz"{,.asc}
)
```

package/zoneinfo-utils/package
```diff
@@ -0,0 +1,76 @@
...
+    121131918c3ae6dc5d40f0eb87563a2be920b71a76e2392c09519a5e4a666881
+    SKIP
+)
+
+timezones=(
+    'africa' 'antarctica' 'asia' 'australasia'
+    'europe' 'northamerica' 'southamerica'
+    'etcetera' 'backward' 'factory'
+)
+
```
matteodelabre: It looks like this is not used anywhere anymore.
```suggestion
```

matteodelabre:
Note that Entware packages are still on version `2020d`, I guess we would need to follow their versions to avoid any compatibility issues. Or we could replace their packages with ours.

package/zoneinfo-utils/package
matteodelabre: Also, I don’t think the `.asc` files are used in this recipe.

package/zoneinfo-utils/package
```diff
@@ -0,0 +1,76 @@
...
+    local zic=.x86/zic
+    # install tzcode stuff
+    make DESTDIR="${pkgdir}" zic="$zic" install
+    # install license
+    install -Dm644 LICENSE "${pkgdir}"/opt/usr/share/licenses/tzdata/LICENSE
+
+    popd > /dev/null
+
+    mv "${pkgdir}"/usr/{lib,share/man} "${pkgdir}"/opt/usr
+    mv "${pkgdir}"/usr/{s,}bin "${pkgdir}"/opt
```
matteodelabre, line 61: To make `tzselect` work, we can simply install the `iso3166.tab` and `zone1970.tab` files which are missing from Entware’s `zoneinfo-core` package (see @LinusCDE comments above).
```suggestion
    mv "${pkgdir}"/usr/{s,}bin "${pkgdir}"/opt
    
    install -D -m 644 -t "$pkgdir"/opt/share/zoneinfo "$srcdir"/iso3166.tab
    install -D -m 644 -t "$pkgdir"/opt/share/zoneinfo "$srcdir"/zone1970.tab
```

matteodelabre:
@LinusCDE Can you try [this version of the package](https://github.com/toltec-dev/toltec/files/5808548/zoneinfo-utils.zip) which includes `iso3166.tab` and `zone1970.tab`? I believe it should fix the issue.

Eeems:
The lines I linked included code to install the files that are missing right now, including mentioning that zones.tab is deprecated. Perhaps we should split this out into newer versions of zoneinfo-* with the latest tzdata package.

matteodelabre:
> Perhaps we should split this out into newer versions of zoneinfo-* with the latest tzdata package.

That could work. However, our package would probably be overridden by Entware’s as soon as they update their version. We could ask upstream to include those two files in `zoneinfo-core` (relevant line: <https://github.com/openwrt/packages/blob/002fcac835a128a283716f2a7a16a238da3c1477/utils/zoneinfo/Makefile#L144>).

Eeems:
> 
> 
> > Perhaps we should split this out into newer versions of zoneinfo-* with the latest tzdata package.
> 
> That could work. However, our package would probably be overridden by Entware’s as soon as they update their version. We could ask upstream to include those two files in `zoneinfo-core` (relevant line: https://github.com/openwrt/packages/blob/002fcac835a128a283716f2a7a16a238da3c1477/utils/zoneinfo/Makefile#L144).

That might be the better solution. Do you want to handle that request?

matteodelabre:
I’d prefer if someone else could handle that, but otherwise I’ll do it.

Eeems:
I'll try to find some time to look into how to make the request.

LinusCDE:
> @LinusCDE Can you try [this version of the package](https://github.com/toltec-dev/toltec/files/5808548/zoneinfo-utils.zip) which includes `iso3166.tab` and `zone1970.tab`? I believe it should fix the issue.

That ipkg file works flawlessly! The time is correct and when I set the TZ to the suggested one, I also get the correct date from `date` displayed. :+1: 

Also retested the current state of this package, but that was still missing the file as expected.

Eeems:
Sounds like either a version mismatch, or files not being included by the entware package.

package/zoneinfo-utils/package
Eeems: Fair enough, I think they are used automatically in PKGBUILD file that I pulled from.

LinusCDE (changes requested):
I think you forgot to specify `$_tzcode`.

```shell
╭─linus@HomeLX /tmp/tol2/toltec  ‹Eeems-patch-4›
╰─$ make zoneinfo-utils
./scripts/package-build package/"zoneinfo-utils" build/package/"zoneinfo-utils"
==> Fetching source files
tzcode.tar.gz
curl: (22) The requested URL returned error: 404 
make: *** [Makefile:55: zoneinfo-utils] Fehler 22
```

Eeems:
> I think you forgot to specify `$_tzcode`.
> 
> ```shell
> ╭─linus@HomeLX /tmp/tol2/toltec  ‹Eeems-patch-4›
> ╰─$ make zoneinfo-utils
> ./scripts/package-build package/"zoneinfo-utils" build/package/"zoneinfo-utils"
> ==> Fetching source files
> tzcode.tar.gz
> curl: (22) The requested URL returned error: 404 
> make: *** [Makefile:55: zoneinfo-utils] Fehler 22
> ```

I committed a suggestion from @matteodelabre that removed it.

Eeems:
Should be fixed now.

LinusCDE (dismissed):
Works. I didn't retest on rM 1, but seeing that the package now contains the `iso3166.tab`, I don't see any reason why it shouldn't work.

Tested on rM 2 (toltec/entware installed with bootstrap from #213) and it worked flawless as well.

:shipit:

Eeems:
Hmm, should we install this by default in bootstrap?

matteodelabre:
I see that this package and a few others install man pages to the device. I’m not sure we should do that since space is a bit scarce.

matteodelabre:
> Hmm, should we install this by default in bootstrap?

What would be the motivations for installing it by default?

Eeems:
> I see that this package and a few others install man pages to the device. I’m not sure we should do that since space is a bit scarce.

Man is available in entware, and other packages there do install manpages. Perhaps we should split them out into separate `*-manpages` packages?

package/zoneinfo-utils/package
```diff
@@ -0,0 +1,68 @@
...
+
+pkgnames=(zoneinfo-utils)
+pkgdesc="Utilities for interacting with zoneinfo files"
+url=https://www.iana.org/time-zones
+_tzver=2020f
+pkgver="$_tzver"-1
+timestamp=2020-05-04T06:16Z
+section=utils
+maintainer="Eeems <eeems@eeems.email>"
+license="custom: public domain"
```
matteodelabre, line 15: This value of the license field conflicts with [the format required in the docs](https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#license). I guess we should update the docs with some allowed exceptions such as this one.

matteodelabre:
> Man is available in entware, and other packages there do install manpages. Perhaps we should split them out into separate `*-manpages` packages?

Good point. I think this can be done in a separate PR.

package/zoneinfo-utils/package
Eeems: Opened #236

Eeems:
Opened #235

Eeems:
> What would be the motivations for installing it by default?

I'm not sure what I was thinking before. timedatectl doesn't require these files right?

matteodelabre:
I believe it uses `zone.tab` which is already in `zoneinfo-core`. We should however install `zoneinfo-core` by default ideally, which is not the case currently.

Eeems:
Doesn't entware install it by default?

matteodelabre:
For some reason it installs `zoneinfo-asia` and `zoneinfo-europe` but not `zoneinfo-core`. (See [the definition of `entware-opt`](https://github.com/Entware/rtndev/blob/cd7604c890bb6b00b2e7174bb4bd27862a3f7179/entware-opt/Makefile#L22).)

Eeems:
Very odd. I'll open an issue to address it.

LinusCDE:
Retested on rM 1. Works as well.

Eeems:
Can I get a approval so I can merge?

matteodelabre (approved):

PR #184, opened 2021-01-04
Add restream to toltec

danshick:
Adds restream helper binary to toltec. restream will need to update the script to stop assuming that this binary is in the home directory, and rely on it being in the path. Additionally, another PR (in progress) will add support for rm2fb users.

### DIFF
--- a/package/restream/package
+++ b/package/restream/package
@@ -10,7 +10,7 @@ timestamp=2021-01-01T13:55:28Z
 section=utils
 maintainer="Dan Shick <dan.shick@gmail.com>"
 license=MIT
-
+conflicts=(rm2fb)
 image=rust:v1.2.1
 source=(https://github.com/rien/reStream/archive/c41b87778953513cd52d9c58b3cc5ce52825700e.zip)
 sha256sums=(89da0932f17a546f194b816eb28a83f84f2484a1508d545bfce2115908248fc3)
--- a/package/restream/package
+++ b/package/restream/package
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(restream)
+pkgdesc="A binary framebuffer capture tool for the reStream script"
+url=https://github.com/rien/reStream
+pkgver=0.0.0-c41b877
+timestamp=2021-01-01T14:55:28+01:00
+section=utils
+maintainer="Dan Shick <dan.shick@gmail.com>"
+license=MIT
+
+image=rust:v1.2.1
+source=(https://github.com/rien/reStream/archive/c41b87778953513cd52d9c58b3cc5ce52825700e.zip)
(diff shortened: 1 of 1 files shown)


package/restream/package
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(restream)
+pkgdesc="A binary framebuffer capture tool for the reStream script"
+url=https://github.com/rien/reStream
+pkgver=0.0.0-c41b877
+timestamp=2021-01-01T14:55:28+01:00
```
Eeems: This should be set to the date of https://github.com/rien/reStream/commit/c41b87778953513cd52d9c58b3cc5ce52825700e with no timezone modifier.

Eeems (changes requested):
You should also add a postinstall script that just lets the user know that they have to install the PC side for this to work. Maybe give a link as well to the repository.

package/restream/package
```diff
@@ -0,0 +1,27 @@
...
+section=utils
+maintainer="Dan Shick <dan.shick@gmail.com>"
```
Eeems, line 11: https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#maintainer You okay with being the one in charge of reviewing alongside one of us with merge power? Or would you prefer for one of us to own that responsibility for now?

danshick: I'm happy to review updates for this package.

package/restream/package
danshick: That is the commit date according to `git log --date=iso-strict`. I'll fix the TZ, but I avoided changing it because I know I'll get that math exactly wrong :)

package/restream/package
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(restream)
+pkgdesc="A binary framebuffer capture tool for the reStream script"
+url=https://github.com/rien/reStream
+pkgver=0.0.0-c41b877
```
raisjn: https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#pkgver

package/restream/package
Eeems: Setting the timestamp is the one thing I really dislike doing in the packages myself due to this :P

package/restream/package
danshick: Done

package/restream/package
danshick: Done, used [this](https://dencode.com/en/date/iso8601).

Eeems (dismissed):
Works fine on the rM1.

danshick:
Added a conflict for rm2fb. I'll remove it and PR the change once reStream support is added.

Eeems (approved):

LinusCDE (commented):
Works fine. :+1: (Not that it matters now) xD

PR #187, opened 2021-01-05
Use systemctl is-enabled in function is-enabled

koenvervloesem:
This PR changes the function `is-enabled` in `scripts/install-lib` to use the built-in `systemctl is-enabled` command instead of piping the output of `systemctl list-unit-files` to `awk`.


Eeems:
I feel like there was a reason we were using list-unit-files instead of is-enabled here, but I could be very wrong.
@matteodelabre?

scripts/install-lib
```diff
@@ -21,7 +21,7 @@ is-enabled() {
     # Note that we must invert the condition so that the exit code indicates
     # a failure when the unit exists and is disabled or when the unit does
     # not exist
-    ! systemctl list-unit-files "$1" | awk "/$1/{exit \$2 == \"enabled\"}"
+    ! systemctl --quiet is-enabled "$1"
```
LinusCDE:
```suggestion
    systemctl --quiet is-enabled "$1" 2>/dev/null
```

The original command didn't print any error for missing service files. The inversion is also no more necessary.

LinusCDE (dismissed):
Thanks for catching that. This is a lot more efficient indeed!

LinusCDE:
> I feel like there was a reason we were using list-unit-files instead of is-enabled here, but I could be very wrong.
> @matteodelabre?

I personally don't see a problem there. The function is documented what it is supposed to do.

```bash
# Check whether a systemd unit exists and is enabled
#
# Arguments:
#
# $1 - Full name of the systemd unit, e.g. "xochitl.service"
#
# Exit code:
#
# 0 if the unit exists and is enabled, 1 otherwise
is-enabled() { ... }
```

The only change now would be that checking e.g. for "tarnish" instead of "tarnish.service" would works. But I don't see a problem there. If anything this seems like an improvement to me. Could be wrong though.

LinusCDE:
Maybe an edit to the documentation, that not just the full name is required would be appropriate. But whether we want this behaviour is also a point to discuss.

scripts/install-lib
koenvervloesem: Yes, I was wondering about suppressing the error message, I'll add the redirect of stderr.

But the inversion is still necessary:

```console
koan@tux:~$ sudo systemctl disable ssh.service
Synchronizing state of ssh.service with SysV service script with /lib/systemd/systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install disable ssh
Removed /etc/systemd/system/sshd.service.
Removed /etc/systemd/system/multi-user.target.wants/ssh.service.
koan@tux:~$ systemctl --quiet is-enabled ssh.service; echo $?
1
koan@tux:~$ sudo systemctl enable ssh.service
Synchronizing state of ssh.service with SysV service script with /lib/systemd/systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install enable ssh
Created symlink /etc/systemd/system/sshd.service → /lib/systemd/system/ssh.service.
Created symlink /etc/systemd/system/multi-user.target.wants/ssh.service → /lib/systemd/system/ssh.service.
koan@tux:~$ systemctl --quiet is-enabled ssh.service; echo $?
0
```

From the `systemctl` man page:

           Checks whether any of the specified unit files are enabled (as with enable). Returns an
           exit code of 0 if at least one is enabled, non-zero otherwise.

LinusCDE:
All in all, the changes seem to boil down to these test cases as far as I can see (when suggestions are applied to fix bugs):

In my example, `tarnish.service` is currently enabled and `remux.service` is present but disabled.

 Value of `$1`            | Old exit code | New exit code
-------------------- | ------------- | ---------------
 tarnish.service         | 0                  |  0
 tarnish                    | 1                   | **0**
 tarnishfoo.service    | 1                   |  1
 remux.service          | 1                   | 1

koenvervloesem:
So because the exit code is 0 if **at least one** service is enabled, it's probably better (less error-prone) if the function `is-enabled` requires/documents the use of the full name, isn't it?

LinusCDE:
> So because the exit code is 0 if **at least one** service is enabled, it's probably better (less error-prone) if the full name is required, no?

I'm not sure what systemd decides in the case of duplicate services files. From testing it seems, that if `$1` has no ending (`.service`, `.mount`, `.socket`, ...), systemd puts a `.service` at the end by default.

Returning exit code 1 if a file didn't contain a dot (`.`) would most likely make the behaviour match again.

scripts/install-lib
koenvervloesem: Sorry, you're right about the inversion, I forgot the result of the if :)

LinusCDE (commented):
Seems good so far! :+1: 

I'll wait a bit though for any further discussion regarding the optional `.service` files from other persons participating / mentioned in this pr.

matteodelabre:
> I feel like there was a reason we were using list-unit-files instead of is-enabled here, but I could be very wrong.
> @matteodelabre?

I think I simply missed `systemctl is-enabled` while reading the man page. Anyway, this is clearly a welcome improvement, thanks @koenvervloesem!

In addition to the differences highlighted by @LinusCDE, we should also be aware that `systemctl is-enabled` also reports success for units in the following states: `enabled-runtime`, `alias`, `static`, `indirect`, `generated` and `transient`, whereas the existing `is-enabled()` function only reports success for `enabled` units.

Overall, I don’t believe that making the function more permissive is an issue given how it is currently used, i.e. for checking if a launcher is enabled or if conflicting units are enabled.

koenvervloesem:
This looks like an example 'in the wild' of the use of systemd unit names without the full name:

https://github.com/toltec-dev/toltec/blob/testing/package/rm2fb/package#L61

Apart from the typo (`xochtil` instead of `xochitl`), it would probably be better to make use of the `is-enabled` function here too, as well as using the full name `xochitl.service`.

LinusCDE:
> > I feel like there was a reason we were using list-unit-files instead of is-enabled here, but I could be very wrong.
> > @matteodelabre?
> 
> I think I simply missed `systemctl is-enabled` while reading the man page. Anyway, this is clearly a welcome improvement, thanks @koenvervloesem!
> 
> In addition to the differences highlighted by @LinusCDE, we should also be aware that `systemctl is-enabled` also reports success for units in the following states: `enabled-runtime`, `alias`, `static`, `indirect`, `generated` and `transient`, whereas the existing `is-enabled()` function only reports success for `enabled` units.
> 
> Overall, I don’t believe that making the function more permissive is an issue given how it is currently used, i.e. for checking if a launcher is enabled or if conflicting units are enabled.

Nice. Feel free to approve this then if there is no more to discuss.

matteodelabre:
> This looks like an example 'in the wild' of the use of systemd unit names without the full name:
> 
> https://github.com/toltec-dev/toltec/blob/testing/package/rm2fb/package#L61
> 
> Apart from the typo (`xochtil` instead of `xochitl`), it would probably be better to make use of the `is-enabled` function here too, and using the full name `xochitl.service`.

Nice catch! Would you be willing to make this change too?

koenvervloesem:
Sure!

This looks wrong too:

https://github.com/toltec-dev/toltec/blob/testing/scripts/bootstrap/bootstrap#L137

`systemctl -q --is-enabled opt.mount` would have `systemctl` complain about an unknown option `--is-enabled` because it doesn't need the dashes. Is the `install-lib` script also imported in the bootstrap script so the `is-enabled` function can be used there?

matteodelabre:
> This looks wrong too:
> 
> https://github.com/toltec-dev/toltec/blob/testing/scripts/bootstrap/bootstrap#L137
>
> `systemctl -q --is-enabled opt.mount` would have `systemctl` complain about an unknown option `--is-enabled` because it doesn't need the dashes.

Indeed. Thanks for reporting!

> Is the `install-lib` script also imported in the bootstrap script so the `is-enabled` function can be used there?

Unfortunately not, since the bootstrap script is meant to be fully standalone to simplify the installation procedure.

koenvervloesem:
Ok, that's it then. My first minor contribution to toltec, as a professional nitpicker ;)

koenvervloesem:
Sorry, forgot to increase the pkgver of rm2fb, @Eeems

Eeems:
@koenvervloesem You'll also need to update the sha256sum in the readme to match the bootstrap script.

matteodelabre:
I took the liberty to update the docblock above `is-enabled` to reflect the updated semantics due to this changeset.

koenvervloesem:
Sha256sum updated

matteodelabre (approved):

matteodelabre:
Your first contribution is much appreciated @koenvervloesem! Looking forward to the next ones :-)

PR #188, opened 2021-01-05
Fix spelling mistake

Eeems:


Eeems:
I see that #187 fixes this

PR #189, opened 2021-01-05
Prevent unwanted PRs against stable

matteodelabre:
(Part of fixing #174)

* Add a workflow check that fails if a PR targets the `stable` branch and does not have the `merge` label.
* Warn future contributors that they should target the `testing` branch instead of the default `stable` branch.

### DIFF
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -1,0 +1,13 @@
+name: pr-labels
+on:
+    pull_request:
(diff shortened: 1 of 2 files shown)


raisjn (approved):
lgtm - does it mean that we manually add "merge" label to the PR when creating it for stable merges?

matteodelabre:
Yup. This is to guard us from accidentally merging to stable, since we’ll switch the default branch to stable as per #174.

Eeems:
Looks good to me!

PR #190, opened 2021-01-05
[genie][1.0] add new package for genie gesture engine

raisjn:
genie is a gesture based launcher - it recognizes gestures and runs commands

### DIFF
--- a/package/rmkit/genie.service
+++ b/package/rmkit/genie.service
@@ -1,0 +1,17 @@
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+[Unit]
+Description=Gesture launcher
+After=opt.mount
+StartLimitInterval=30
+StartLimitBurst=5
+
+[Service]
+ExecStart=/opt/bin/genie /opt/etc/genie.conf
+Restart=on-failure
+RestartSec=5
+Environment="HOME=/home/root"
+
+[Install]
+WantedBy=multi-user.target
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -38,7 +38,7 @@ genie() {
 
     configure() {
         systemctl daemon-reload
-        systemctl enable genie --now
+        echo "Run 'systemctl enable genie --now' to enable genie"
     }
 
     preremove() {
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -2,7 +2,7 @@
 # Copyright (c) 2020 The Toltec Contributors
 # SPDX-License-Identifier: MIT
 
-pkgnames=(harmony mines nao remux simple)
+pkgnames=(genie harmony mines nao remux simple)
 timestamp=2021-01-05T14:03-08:00
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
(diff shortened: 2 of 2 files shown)


package/rmkit/package
```diff
@@ -11,17 +11,46 @@ image=python:v1.1
...
+
+    package() {
+        install -D -m 755 "$srcdir"/src/build/genie "$pkgdir"/opt/bin/genie
+        install -D -m 644 "$srcdir"/genie.service "$pkgdir"/lib/systemd/system/genie.service
+        install -D -m 644 "$srcdir"/src/genie/example.conf "$pkgdir"/opt/etc/genie.example.conf
+    }
+
+    configure() {
+        systemctl daemon-reload
+        systemctl enable genie --now
```
Eeems: I thought we were avoiding doing this for users for most things?

package/rmkit/package
```diff
@@ -11,17 +11,46 @@ image=python:v1.1
...
+
+    package() {
+        install -D -m 755 "$srcdir"/src/build/genie "$pkgdir"/opt/bin/genie
+        install -D -m 644 "$srcdir"/genie.service "$pkgdir"/lib/systemd/system/genie.service
+        install -D -m 644 "$srcdir"/src/genie/example.conf "$pkgdir"/opt/etc/genie.example.conf
+    }
+
+    configure() {
+        systemctl daemon-reload
+        echo "Run 'systemctl enable genie --now' to enable genie"
```
Eeems, line 41: I think you can use how-to-enable
```suggestion
       how-to-enable genie
```

danshick:
Just installed, I'll test drive for the next day or two.

Eeems:
I think @PeterGrace also has installed and tested on the rM2.

danshick:
So far genie is working well.

Eeems (approved):

PeterGrace:
> I think @PeterGrace also has installed and tested on the rM2.

Yes indeed -- this is my configuration for the reMarkable 2, specifically so that the full button features of oxide are retained even though there aren't physicalized buttons on the reMarkable 2.  One item of note: see how the paths on the command have to be fully qualified; I'm not sure if there's a genie package/config issue here but the exec'd command doesn't have /opt/bin in the path.  That could be a different issue (for instance, does the toltec installer include /opt/bin in the shell rc file?  At  any rate, the below config does work but beware: that "1 finger tap" for oxide menu also triggers if, like me, you're a lefty and lay your hand on the screen while writing!

```
gesture=tap
fingers=1
command=/opt/bin/rot apps call leftHeld
duration=3

gesture=tap
fingers=2
command=/opt/bin/rot screen call screenshot
duration=3

gesture=tap
fingers=3
command=/opt/bin/rot apps get applications | /opt/bin/jq '."codes.eeems.erode"' -cr | sed 's|/codes/eeems/oxide1/||' | xargs -I {} /opt/bin/rot --object Application:{} apps call launch
duration=3
```

package/rmkit/genie.service
```diff
@@ -0,0 +1,17 @@
...
+Description=Gesture launcher
+After=opt.mount
+StartLimitInterval=30
+StartLimitBurst=5
+
+[Service]
+ExecStart=/opt/bin/genie /opt/etc/genie.conf
+Restart=on-failure
+RestartSec=5
+Environment="HOME=/home/root"
```
Eeems, line 14:
```suggestion
Environment="HOME=/home/root"
Environment="PATH=/opt/bin:/opt/sbin:/opt/usr/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin"
```

raisjn:
> At any rate, the below config does work but beware: that "1 finger tap" for oxide menu also triggers if, like me, you're a lefty and lay your hand on the screen while writing!

nice report! i guess we can make genie work better by not invoking any gestures if pen is on or near screen so that this doesn't happen as frequently

PR #191, opened 2021-01-05
[rmkit] update remux to 0.1.5-1

raisjn:
* launch remux on 3 finger tap (much more useful on rM2)
* remux launches when SIGWINCH signal is sent


matteodelabre:
Looks good. Could you please update the source timestamp in the recipe?

matteodelabre (approved):
Looks good to me

PR #193, opened 2021-01-06
[docs] webpage built from rst

raisjn:
Add an index page built from rst. Uses hand drawn logo from matteo


raisjn:
Screenshot_2021-01-08_12-06-50
Screenshot_2021-01-08_12-07-23

web/index.rst
```diff
@@ -0,0 +1,88 @@
...
+
+  No, it is a community project
+
+* Where can I get help?
+
+  `please open an issue on github <#>`_
+
+* Will this brick my remarkable?
+
+  No
```
Eeems:
```suggestion
  No, but `standard disclaimers apply <https://github.com/toltec-dev/toltec/blob/stable/LICENSE>`_
```

raisjn:
moving this out of WIP. do people have more content / suggestions? if so, please feel free to update the branch (or add comments)

can see the latest site here: https://opkg.rmkit.dev/ (was auto generated from rmkit fork of toltec)

LinusCDE:
I quite like that color!

I toyed around a bit with having icons float space like, but think it ultimatively would just look like a cheap among us knock off and would increase the homepages complexity (would need a js to coordinate having like 2-3 random icons floating and a way of getting the icons from the packages `/opt/etc/draft/icons` folder).


https://user-images.githubusercontent.com/22298664/104127387-20f8aa80-5362-11eb-99cf-1e1bd1e13bc5.mp4

---

The index.html should probably get placed into `build/`.

raisjn:
I like the floating icons! We can hardcode a few to start with (instead of adding tooling) and later figure that out

The next two things are:

* finalize content
* add in links (we don't have an opkg quickstart guide, for example)

LinusCDE:
> I like the floating icons! We can hardcode a few to start with (instead of adding tooling) and later figure that out
> 
> The next two things are:
> 
>     * finalize content
> 
>     * add in links (we don't have an opkg quickstart guide, for example)

To have decent flexibility, I would like to automate the creation of animations with JS though. But I can't find a way to make rst2html include it.

Eeems:
Should we maybe use [rst2html5](https://pypi.org/project/rst2html5/)?

It will allow you to include a script with 
image

okayzed:
rst2html5 sounds good to me, just pushed it

matteodelabre:
Made some changes to the web build process:

* Move the workflow for pushing to the remote server into `stable`
* Make the build script more consistent with other scripts in the repo
* Automatically compute the bootstrap script hash displayed in `index.rst`

scripts/web-build
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/bash
```
raisjn: should be /usr/bin/env bash or /bin/bash

matteodelabre: Thanks!

Eeems:
When I'm trying to generate this, the stylesheet isn't being added to the html.

image
I'm also seeing some strange stuff happening to the text. I blame running this on Windows.

Eeems (approved):
Okay, building on a linux host works properly.

matteodelabre:
> When I'm trying to generate this, the stylesheet isn't being added to the html.

I added the `embed-stylesheet: no` flag to docutils, so the stylesheet should be linked in the header of the HTML file. Do you see a link in the source? Is the `toltec.css` file in the same folder as the HTML file you’re viewing?

> image
> I'm also seeing some strange stuff happening to the text. I blame running this on Windows.

Looks like an encoding issue. Can you try again with the latest commit?

Eeems:
> I added the `embed-stylesheet: no` flag to docutils, so the stylesheet should be linked in the header of the HTML file. Do you see a link in the source? Is the `toltec.css` file in the same folder as the HTML file you’re viewing?

On windows, no it doesn't add the link to the html file. The css file does exist though.


> Looks like an encoding issue. Can you try again with the latest commit?

That fixed it!

raisjn:
for subsequent web updates, should we open PR on stable branch or does it go through testing -> stable?

> I added the embed-stylesheet: no flag to docutils, so the stylesheet should be linked in the header of the HTML file. Do you see a link in the source? Is the toltec.css file in the same folder as the HTML file you’re viewing?

i see it when i do a build locally

matteodelabre:
> for subsequent web updates, should we open PR on stable branch or does it go through testing -> stable?

I think the most straightforward approach is to push the web updates through stable merges. This way, all end-user-facing content is checked and approved at the same point in our process.

PR #195, opened 2021-01-06
Add keywriter package

matteodelabre:
The app can only be controlled by an USB OTG keyboard, which for now users have only managed to connect on rM1 (see
<https://github.com/dps/remarkable-keywriter/issues/14>).

The docs are a bit lacking, so if you want to test the app:

* the “home” button (middle one) opens a menu to select a file to edit in the `/home/root/edit` directory. By default `draft.md` is opened,
* the default view shows the content of the active file (or an empty screen if the file is empty) with any Markdown syntax properly rendered,
* switch between normal and edit mode by pressing \<Esc\>.

### DIFF
--- a/package/keywriter/keywriter.draft
+++ b/package/keywriter/keywriter.draft
@@ -1,0 +1,7 @@
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+name=keywriter
+desc=Markdown-enabled free writing app
+call=/opt/bin/keywriter
+term=:
--- a/package/keywriter/package
+++ b/package/keywriter/package
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(keywriter)
+pkgdesc="Markdown-enabled free writing app"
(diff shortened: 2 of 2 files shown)


LinusCDE (approved):
Works fine. :+1: (Worked even with my abomination of a ["otg"-adapter](https://i.imgur.com/S9EKJim.png)). :laughing: 

Just curious: How did you even manage to build the package? For me it failed due to a bug here: #198 . My guess would be that you either built it with v1.2.1 (since you seemed to add 1.2.2 to fix an encountered bug) or has some cached version. Or do we have usually have a "latest" of the toolchain anyway and my PR #198 not needed?

---

It would be really cool to have an icon for this as well. Just a idea for @dps:
screenshot3191

matteodelabre:
Thanks for testing!

> Just curious: How did you even manage to build the package? For me it failed due to a bug here: #198 . My guess would be that you either built it with v1.2.1 (since you seemed to add 1.2.2 to fix an encountered bug) or has some cached version. Or do we have usually have a "latest" of the toolchain anyway and my PR #198 not needed?

I built it yesterday evening and I messed up the `latest` tag today, just before you tested it on your side apparently. Sorry about that. There is indeed usually a `latest` tag, but it’s best to set an explicit version to avoid builds suddenly breaking, so your PR is still valid.

PR #197, opened 2021-01-07
Add koreader nightly with rM2 support

raisjn:
instead of updating koreader package,  add a package for koreader-nightly that conflicts with the main koreader package.


raisjn:
matteo added rm2 support to main koreader

PR #198, opened 2021-01-07
Fix base image for stripping not found

LinusCDE:
When trying to build the keywriter package (#195) for testing on my server (since my internet at home for some reason takes ages to pull new images), I encountered the following error:

grafik

The error seems to be that the image was specified without an version and the default ("latest") is not used by us and was not found.

@Eeems @matteodelabre This for now just assumed a version, but we should consider using taking the version part of the already specified image or something else to keep that hidden part in the code somehow up to date.

I'll use this version for building keywriter for now, but I guess this should not be the final fix.

### DIFF
--- a/scripts/package-build
+++ b/scripts/package-build
@@ -151,7 +151,7 @@ for pkgname in "${pkgnames[@]}"; do
             section "Stripping $pkgname"
             docker container run --interactive --rm \
                 --mount type=bind,src="$(realpath "$pkgdir")",dst=/pkg \
-                "$(image-name base:v1.2.2)" /bin/bash \
+                "$(image-name "$image")" /bin/bash \
                 << "SCRIPT"
 find /pkg -print0 -type f | xargs --null arm-linux-gnueabihf-strip --strip-all &> /dev/null || true
 SCRIPT
--- a/scripts/package-build
+++ b/scripts/package-build
@@ -153,7 +153,7 @@ for pkgname in "${pkgnames[@]}"; do
                 --mount type=bind,src="$(realpath "$pkgdir")",dst=/pkg \
                 "$(image-name "$image")" /bin/bash \
                 << "SCRIPT"
-find /pkg -print0 -type f | xargs --null arm-linux-gnueabihf-strip --strip-all &> /dev/null || true
+find /pkg -print0 -type f | xargs --null "${CROSS_COMPILE}strip" --strip-all &> /dev/null || true
 SCRIPT
             find "$pkgdir" -print0 -type f | xargs --null strip --strip-all &> /dev/null || true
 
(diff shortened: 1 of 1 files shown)


matteodelabre:
Since all images contain the build tools, we could simply use the image specified in each recipe (i.e. `"$image"` instead of `base:v1.2.2`), just like in the line for building the package:

https://github.com/toltec-dev/toltec/blob/7569ea6f48c94097237637a2197b5ad6715af1ac/scripts/package-build#L115

matteodelabre:
By the way, we can replace `arm-linux-gnueabihf-strip` with `${CROSS_COMPILE}strip` to make it easier to change the cross-compiler name in the future. The CROSS_COMPILE var [is defined for all toolchain images.](https://github.com/toltec-dev/toolchain/blob/6e2825fae706083821b6fc330beb313bb5d72fa3/base/Dockerfile#L69)

LinusCDE:
Thanks, was searching for the button. Best hidden button on GitHub. :laughing:

For reviewing, I guess just building any package that has at least one binary should do.

scripts/package-build
```diff
@@ -151,7 +151,7 @@ for pkgname in "${pkgnames[@]}"; do
             section "Stripping $pkgname"
             docker container run --interactive --rm \
                 --mount type=bind,src="$(realpath "$pkgdir")",dst=/pkg \
-                "$(image-name base)" /bin/bash \
+                "$(image-name "$image")" /bin/bash \
                 << "SCRIPT"
 find /pkg -print0 -type f | xargs --null arm-linux-gnueabihf-strip --strip-all &> /dev/null || true
```
matteodelabre: I suggest we do the change now instead of opening a separate PR.

```suggestion
find /pkg -print0 -type f | xargs --null "${CROSS_COMPILE}strip" --strip-all &> /dev/null || true
```

matteodelabre (changes requested):

scripts/package-build
LinusCDE: Thanks missed that.

--- 
Btw do we also have declarations of $CC, $STRIP, $LD and such? I think they could be helpful.

Eeems (approved):
Looks good to me

matteodelabre (approved):
Tested by compiling `draft`, can confirm that it works correctly.

matteodelabre:
> Btw do we also have declarations of $CC, $STRIP, $LD and such? I think they could be helpful.

@LinusCDE Good idea. They are not defined in the toolchain currently. Could you open an issue on the toolchain repo about that, please?

LinusCDE:
Done: https://github.com/toltec-dev/toolchain/pull/5

PR #199, opened 2021-01-07
Oxide v2.1

Eeems:
- Upgrade Oxide to v2.1
- Add automatic conversion of draft configuration files to oxide application registration files
- Add flag to disable draft -> oxide conversion for specific packages
- Added oxide application registration file to KOReader to resolve fbdepth issues
- Added code to automatically ask tarnish (Oxide system service) to reload application registration on install/remove of packages that contain a registration file.

### DIFF
--- a/package/koreader/KOReader.oxide
+++ b/package/koreader/KOReader.oxide
@@ -1,6 +1,6 @@
 {
   "displayName": "KOReader",
-  "description": "ebook reader",
+  "description": "An ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats",
   "bin": "/opt/bin/koreader",
   "icon": "/opt/etc/draft/icons/koreader.png",
   "type": "foreground",
--- a/package/koreader/KOReader.oxide
+++ b/package/koreader/KOReader.oxide
@@ -4,7 +4,7 @@
   "bin": "/opt/bin/koreader",
   "icon": "/opt/etc/draft/icons/koreader.png",
   "type": "foreground",
-  "on": {
+  "events": {
     "pause": "/opt/koreader/fbdepth -d 16 -r 1",
     "stop": "/opt/koreader/fbdepth -d 16 -r 1",
     "resume": "/opt/koreader/fbdepth -d 8 -r 1"
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -16,7 +16,7 @@ _srcver="v${pkgver%-*}"
 source=(
     "https://build.koreader.rocks/download/stable/$_srcver/koreader-remarkable-$_srcver.zip"
     koreader.draft
-    koreader.oxide
+    KOReader.oxide
     rm2-support.patch
 )
 sha256sums=(
@@ -40,7 +40,7 @@ package() {
     ln -s /opt/bin/fbdepth "$pkgdir"/opt/koreader/fbdepth
 
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
-    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$recipedir"/koreader.oxide
+    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$recipedir"/KOReader.oxide
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
     install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
 }
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -5,7 +5,7 @@
 pkgnames=(koreader)
 pkgdesc="An ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats"
 url=https://github.com/koreader/koreader
-pkgver=2020.12-1
+pkgver=2020.12-2
 timestamp=2020-10-10T20:13Z
 section=readers
 maintainer="raisjn <of.raisjn@gmail.com>"
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -41,14 +41,9 @@ package() {
 }
 
 configure() {
-    if systemctl --quiet is-active tarnish; then
-        rot apps call reload
-    fi
+    reload-oxide-apps
 }
+
 postremove() {
-    # Check to see if tarnish is running and rot is available
-    # shellcheck disable=SC2009
-    if systemctl list-units --full -all | grep -Fq 'tarnish.service' || [[ "$(ps | grep tarnish | grep -v grep)" != "" ]]; then
-        echo "You may need to manually remove the application entry for KOReader in Oxide."
-    fi
+    reload-oxide-apps
 }
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -18,13 +18,20 @@ source=(
     koreader.png
     koreader.draft
     koreader.oxide
+    rm2-support.patch
 )
 sha256sums=(
     38db8f3895472828d7c1d63fc54db426135a8dbf8425633e299a1a9abfdb69b2
     SKIP
     SKIP
     SKIP
+    SKIP
 )
+
+prepare() {
+    patch -p1 -d"$srcdir" < "$srcdir"/rm2-support.patch
+    rm "$srcdir"/rm2-support.patch
+}
 
 package() {
     install -d "$pkgdir"/opt/koreader
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -13,8 +13,18 @@ license=AGPL-3.0-or-later
 depends=(fbink fbdepth)
 
 _srcver="v${pkgver%-*}"
-source=("https://build.koreader.rocks/download/stable/$_srcver/koreader-remarkable-$_srcver.zip")
-sha256sums=(38db8f3895472828d7c1d63fc54db426135a8dbf8425633e299a1a9abfdb69b2)
+source=(
+    "https://build.koreader.rocks/download/stable/$_srcver/koreader-remarkable-$_srcver.zip"
+    koreader.png
+    koreader.draft
+    koreader.oxide
+)
+sha256sums=(
+    38db8f3895472828d7c1d63fc54db426135a8dbf8425633e299a1a9abfdb69b2
+    SKIP
+    SKIP
+    SKIP
+)
 
 package() {
     install -d "$pkgdir"/opt/koreader
@@ -25,12 +35,15 @@ package() {
     ln -s /opt/bin/fbdepth "$pkgdir"/opt/koreader/fbdepth
 
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
+    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$recipedir"/koreder.oxide
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
     install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
 }
 
 configure() {
-    echo "KOReader has an outstanding bug where it doesn't resize the screen properly when returning to Oxide. See Oxide's release notes for a workaround."
+    if systemctl --quiet is-active tarnish; then
+        rot apps call reload
+    fi
 }
 postremove() {
     # Check to see if tarnish is running and rot is available
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -5,7 +5,7 @@
 pkgnames=(koreader)
 pkgdesc="An ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats"
 url=https://github.com/koreader/koreader
-pkgver=2020.12-2
+pkgver=2020.12-3
 timestamp=2020-10-10T20:13Z
 section=readers
 maintainer="raisjn <of.raisjn@gmail.com>"
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -44,11 +44,3 @@ package() {
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
     install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
 }
-
-configure() {
-    reload-oxide-apps
-}
-
-postremove() {
-    reload-oxide-apps
-}
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -40,7 +40,7 @@ package() {
     ln -s /opt/bin/fbdepth "$pkgdir"/opt/koreader/fbdepth
 
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
-    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$recipedir"/koreder.oxide
+    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$recipedir"/koreader.oxide
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
     install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
 }
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -39,8 +39,8 @@ package() {
     ln -s /opt/bin/fbink "$pkgdir"/opt/koreader/fbink
     ln -s /opt/bin/fbdepth "$pkgdir"/opt/koreader/fbdepth
 
-    install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
-    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$recipedir"/KOReader.oxide
+    install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$srcdir"/koreader.draft
+    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$srcdir"/KOReader.oxide
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
     install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
 }
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -15,14 +15,12 @@ depends=(fbink fbdepth)
 _srcver="v${pkgver%-*}"
 source=(
     "https://build.koreader.rocks/download/stable/$_srcver/koreader-remarkable-$_srcver.zip"
-    koreader.png
     koreader.draft
     koreader.oxide
     rm2-support.patch
 )
 sha256sums=(
     38db8f3895472828d7c1d63fc54db426135a8dbf8425633e299a1a9abfdb69b2
-    SKIP
     SKIP
     SKIP
     SKIP
--- a/package/oxide/package
+++ b/package/oxide/package
@@ -8,8 +8,8 @@ timestamp=2021-01-07T03:28Z
 maintainer="Eeems <eeems@eeems.email>"
 license=MIT
 
-source=(https://github.com/Eeems/oxide/files/5783885/oxide.zip)
-sha256sums=(2e46ea132061424a98207f99d875119914498b384bc1222a8fbd586a75595c65)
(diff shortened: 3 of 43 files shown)


scripts/install-lib
```diff
@@ -55,3 +69,21 @@ how-to-enable() {
 
     echo "$ systemctl enable --now ${1/.service/}"
 }
+
+# Reload application registrations in oxide
+#
+# Output:
+#
+# Reload status message
+reload-oxide-apps() {
```
Eeems, line 81: The one concern I have with this on every install/upgrade/remove of an application, is that if you install multiple applications, you only need to call it once, instead of once per application. So extra processing happens. Not really sure how to work around that though.

Eeems: Well now that I've tested, it's rather interesting watching them disappear one by one and then reappear all at once. Doesn't seem to be too big of a deal.

raisjn:
asi mentioned offline, I'm not a fan of creating new oxide files for every package. it will mean redundant work.

i think only make the koreader.oxide file while we figure out the path forward. even better is to just add new oxide specific options to draft format and use those

Eeems:
> asi mentioned offline, I'm not a fan of creating new oxide files for every package. it will mean redundant work.
> 
> i think only make the koreader.oxide file while we figure out the path forward. even better is to just add new oxide specific options to draft format and use those

Currently working on adding automatic generation of oxide files if it's not specified. Without these files, the experience for end users is not that great. That said, I do agree that maintaining both isn't ideal.

scripts/package-build
```diff
@@ -123,6 +123,38 @@ elif [[ $(type -t build) == "function" ]]; then
     error "A 'build()' function was defined but the 'image' variable is missing"
 fi
 
+function convert-draft-apps() {
```
Eeems: @matteodelabre Would it make more sense to move this to package-lib?

Eeems: I think I've answered this question as "Yes"

raisjn:
from a zoomed out perspective, i have some qualms with the idea of creating new .oxide format when we have an acceptable one for draft. 

my understanding is that this path was taken because:

1) reloading the apps at display time is too slow (250 - 500ms) - potentially because of the dbus connection. instead of dynamically listing apps, this approach does it at package build time and install time. 
2) the draft file config does not support features you want to add - either `chroot` or handlers for on_pause, on_stop, etc


questions:

* why bump all those package versions?
* why is this not more general? for example, a package can drop a 'post-install' script in /opt/opkg/hooks/postinstall.d/. as it stands now, this is adding coupling between oxide and toltec
* is there a guarantee that every launchable app will always have a .draft file? 
* why JSON? i know you said JSON is valid YAML but why not start with YAML or TOML? (in my mind, draft spec > TOML > YAML > JSON here)


the main benefit (so far) is that KOReader gets a few new fields for on_pause, on_resume, on_kill. i still think there's a shorter path (that doesn't involve making a launcher spec):
 
1) update draft spec to add a few new fields (so new file format is unecessary)
2) speed up app importing code and run it when the oxide UI displays

raisjn:
alternative path:

* add just one .oxide file for koreader with your custom changes
* modify opkg to call post-install hooks, where you read all draft files and the koreader.oxide files and refresh app list. 

changes from what's here:

* only one .oxide file
* no change to toltec build that are specific to oxide
* no generation from draft -> .oxide required

LinusCDE:
> why JSON? i know you said JSON is valid YAML but why not start with YAML or TOML? (in my mind, draft spec > TOML > YAML > JSON here)
@raisjn 

Had the same question a while ago. I think it started with a discussion in #toltec [about here](https://discordapp.com/channels/385916768696139794/756853330382422037/793650461134684211) (or just search up "toml" in the toltec channel) when the koreader pr got finished. I also proposed the named formats [here](https://github.com/Eeems/oxide/issues/122). Json was chosen in part because *[QSettings can handle it](https://discordapp.com/channels/385916768696139794/756853330382422037/793654797726646272)* and because users are not supposed to edit them normally anyway (we had a conversation about what format if meant to be edited by users and what not there as well).

At that time, nobody else chimed in and I think changing it now would be too much hassle for too less gain imo (though I would still love toml or yaml more, but I'm biased here as well and don't really understand the internal consequences for oxide).

---

@Eeems should I wait with my review until the above things are settled first or do it today anyway?

Eeems:
> 
> questions:
> 
> * why bump all those package versions?

To generate a version of the package with the oxide file. Otherwise they won't have one and won't automatically register with oxide on install.

> * why is this not more general? for example, a package can drop a 'post-install' script in /opt/opkg/hooks/postinstall.d/. as it stands now, this is adding coupling between oxide and toltec

This is an option, but it does require someone to write such a script. As it stands my current solution requires jq and jo to be installed.

> * is there a guarantee that every launchable app will always have a .draft file? 

That's something we can decide on the package level. We've added it to a package where there isn't one in the source repo. 

> * why JSON? i know you said JSON is valid YAML but why not start with YAML or TOML? (in my mind, draft spec > TOML > YAML > JSON here)

JSPN parsing was already available to mean with no extra work required. It also provides a transparent upgrade path to YAML, as YAML is a superset of JSON. I also don't really like the TOML format.

> 
> 
> the main benefit (so far) is that KOReader gets a few new fields for on_pause, on_resume, on_kill. i still think there's a shorter path (that doesn't involve making a launcher spec):
>  
> 1) update draft spec to add a few new fields (so new file format is unecessary)
> 2) speed up app importing code and run it when the oxide UI displays
> 

Updating the draft spec requires me to write/maintain code that parses it, as well as push out a new release with those changes. It also doesn't support the types of data I'm trying to use (arrays, maps).

> alternative path:
> 
> * add just one .oxide file for koreader with your custom changes
> * modify opkg to call post-install hooks, where you read all draft files and the koreader.oxide files and refresh app list. 

There is no such hook as importing is done by the launcher and not the system service.

> 
> changes from what's here:
> 
> * only one .oxide file
> * no change to toltec build that are specific to oxide
> * no generation from draft -> .oxide required

But it requires changes to oxide, which as the package author I'm not really enthusiastic about.

---

> > why JSON? i know you said JSON is valid YAML but why not start with YAML or TOML? (in my mind, draft spec > TOML > YAML > JSON here)
> @raisjn 
> 
> Had the same question a while ago. I think it started with a discussion in #toltec [about here](https://discordapp.com/channels/385916768696139794/756853330382422037/793650461134684211) (or just search up "toml" in the toltec channel) when the koreader pr got finished. I also proposed the named formats [here](https://github.com/Eeems/oxide/issues/122). Json was chosen in part because *[QSettings can handle it](https://discordapp.com/channels/385916768696139794/756853330382422037/793654797726646272)* and because users are not supposed to edit them normally anyway (we had a conversation about what format if meant to be edited by users and what not there as well).
> 
> At that time, nobody else chimed in and I think changing it now would be too much hassle for too less gain imo (though I would still love toml or yaml more, but I'm biased here as well and don't really understand the internal consequences for oxide).

I actually didn't even end up using QSettings as that would have added unnecessary abstraction. But I'm still using the Qt JSON parsing. 

> 
> @Eeems should I wait with my review until the above things are settled first or do it today anyway?

I'd still like to have feedback on the changes themselves.

---

https://github.com/Eeems/oxide/wiki/Developing-Applications-for-Oxide
I've documented the file spec here as it's been implemented.

raisjn:
At a high level, if the .oxide JSON format is meant to replace draft files, I'm opposed to it. It sounds like linus gave feedback on it originally and it was inline with my feedback.

Once the proposed changes (new .oxide files that every package needs to create) hit toltec, I became more aware of them and it raises a few flags. 

If you mean .oxide to be internal to oxide only, then you can do what you please. If you want to create a new launcher format, I think it should be opened for discussion and not tied to oxide 2.1 release. 

As linus mentioned, there is some "sunk cost" here - which seems to say "its too much effort to change now that it's been done". I disagree with that reasoning if you are proposing a new launcher file format.

LinusCDE:
Seems that some packages are not building properly.

Since almost all the packages are affected, I decided to simply use `make repo` for that task and it failed on calculator (the first package).

screenshot3201.

Adding `jo` only for github seems to make local builds not be in sync with gh anymore.

Eeems:
> Seems that some packages are not building properly.

@LinusCDE You'll need to install `jo` on your machine. The PR workflow installs it here: https://github.com/toltec-dev/toltec/pull/199/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160R50

---

>
> At a high level, if the .oxide JSON format is meant to replace draft files, I'm opposed to it. It sounds like linus gave feedback on it originally and it was inline with my feedback.

I already dropped first class support of draft files in v2.0 and was requiring people to use `rot` to add applications properly. This was a continuation of that move after seeing how `rot` doesn't play nicely with package manager installs (See the struggle I had in #159). While it's good to have standards that support multiple applications, we don't have a proper standards body around application launcher formats yet, and the draft format just does not have the features that I need.

> 
> Once the proposed changes (new .oxide files that every package needs to create) hit toltec, I became more aware of them and it raises a few flags.
> 
> If you mean .oxide to be internal to oxide only, then you can do what you please. If you want to create a new launcher format, I think it should be opened for discussion and not tied to oxide 2.1 release.

I don't mean for oxide's format to be used by other applications. They can use them if they'd like to, but end goal would be to define a better standard with a proper "standards body" in place to keep everybody on the same page. Something that has proper channels for enhancing it instead of taking an existing format and just adding our own things that wont work in other launchers.

> 
> As linus mentioned, there is some "sunk cost" here - which seems to say "its too much effort to change now that it's been done". I disagree with that reasoning if you are proposing a new launcher file format.

I'm not proposing a new launcher format, I'm creating a new format for oxide only, and if other launchers want to borrow from it, or to use it that's fine. But I don't intend on keeping draft format support in oxide for much longer. Which is something I've thought I've been clear about since we first discussed the possibility of using the [XDG Desktop Entry format](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html). If I hadn't been clear enough, I apologize for that.

Let me see if I can articulate your primary issues with my format:

 1. > Not suported by other launchers.
    
    It's not meant to be supported by other launchers, they can keep using the draft format.
 2. > It uses JSON which people find hard to use/write.
    
    It's not meant for non-technical users to write. It's only meant for application developers, or package maintainers to use.
 3. > It requires toltec to be aware of the standard.
    
    Fair enough, lets look into alternatives around this. I do like the idea of installing post install hooks that all packages run. That way the repository doesn't directly need to be aware of the contents of oxide, and there is still a mechanism to convert draft files on install. This would also need a mechanism to remove the files on uninstall, and knowing which files need to be removed. It also requires something other than the code I've written in this package, as it depends upon `jq` and `jo`. That said, I could package a conversion tool into oxide that would be called. We would need to sort out how to build these hooks, and how to properly tie them into opkg.
 4. > You want there to be one format instead of multiple
    
    I agree, but the format needs to be good enough, and currently the "owner" of the format isn't very active and hasn't been doing any development on draft in the last 6 months. If/when we decide to create a single format, I think that should be owned by toltec and not by a single application. It should also provide a mechanism for installed applications to be notified of the new files so that they can consume them how they please. That way they can use their own formats if they want to, but they are in charge of the conversion.
 5. > You want there to be a smooth transition
    
    I think the current PR does provide that. Ideally having install/uninstall hooks would be even better, because it's not something that effects the entire repo with a redeploy of packages. I do think doing it that way will take longer, and I think we can target that for a couple weeks out as we sort through it.
 6. > You want the format to be easy to parse
    
    JSON is one of the most widely used/supported formats out there, which makes it very easy to parse in almost any programming language. There is already a library out there for whatever programming language you want to use, with few (if any) exceptions.

LinusCDE:
> >    Seems that some packages are not building properly.
>
>@LinusCDE You'll need to install jo on your machine. The PR workflow installs it here: https://github.com/toltec-dev/toltec/pull/199/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160R50

Will do. Though I also created a new issue (probably more a todo for me) for documenting the needed packages on the host (#210).

Eeems:
@raisjn See https://discord.com/channels/385916768696139794/756853330382422037/797899382455795804 for discussion on having a post install hook. I think that's likely a few weeks out so not ideal for the v2.1 release. I'm thinking provide the oxide file temporarily with my current setup until we can move over to a better solution?

scripts/package-build
matteodelabre: Sure, we ideally want to keep the `package-build` script as lightweight as possible.

matteodelabre:
Thanks for your work on this @Eeems, and thanks everyone for their comments on the topic. I’ll start with a general review of this PR’s intent and I’ll get into the change-level review a bit later.

First, something I hope we can all agree on: _Toltec needs to be independent of any specific launcher_ (or, more generally, of any specific package). This is to encourage the development of several concurrent launchers.

### On the new application entry format

Now, this goal is not currently met, since we’re distributing Draft-specific configuration files. But I think that adding a new launcher-specific format drives us further away from it. Making Toltec aware of an application entry format is okay to me, as long as it’s shared by all launchers.

A short-term solution for this would be to:

* move the new format specification outside of the Oxide project (I don’t mind incorporating it into Toltec, for example, but it could also be a separate project if you think more independence is needed),
* remove all Draft application entries from Toltec and replace them with the new format,
* temporarily generate Draft config files for compatibility from the new-format entries (i.e., the reverse of what you’re currently doing in this PR).

In the long term, we could continue by:

* proposing support for the new format in Draft,
* removing the Draft compatibility layer in Toltec.

### On the hook for registering new apps

As mentioned above, and in the spirit of keeping Toltec independent from specific packages, it would be best to have a way for packages to run a script when other packages are installed/removed (in the spirit of pacman hooks). The solution proposed here does not fit with that spirit.

Since implementing a general solution could take up to a few weeks, I don’t mind merging it as a _temporary_ workaround. We should open a separate issue to track this.

Eeems:
Thanks @matteodelabre, as I've mentioned, I have no intention of having the format that oxide is using become the defacto standard for other launchers, especially since it has oxide specific things in it. I'm perfectly happy to start defining a new shared standard between all the applications that oxide can move towards supporting. Especially if that new standard has a mechanism for making changes to it that all launchers will support, or providing a mechansim for launcher specific information to be in it.

raisjn:
/opt/usr/share/applications is being used to hold oxide specific files (with .oxide extension). Can this be changed to an oxide specific directory?

Eeems:
> 
> 
> /opt/usr/share/applications is being used to hold oxide specific files (with .oxide extension). Can this be changed to an oxide specific directory?

Since /usr/share/applications is part of the xdg standard for application registration. I'd rather keep using it as it's the correct location. They have an extension of `.oxide` which is good enough to keep other applications from using them unintentionally.

raisjn (changes requested):
until we discuss further, let's keep the shared directory clear of oxide files, esp since they are short term solution

Eeems:
That would require another release of Oxide as this is not configurable. It should be fairly easy to clean them up in future releases in such a way that is low effort when we decide on a standard. Oxide doesn't have the same mandate to be backwards compatible. It will provide migration paths, but will not support old configuration formats.

raisjn:
it doesn't make sense to me to add more work by doing this PR and then undoing it instead of changing a path now. currently no one has .oxide files installed in their rM tablet. why create this format and put it in a shared directory and incur tech debt / more work? i get that no one wants to do more work and that changing this path will be extra work for you, but it's less now than doing it later.

Eeems:
> it doesn't make sense to me to add more work by doing this PR and then undoing it instead of changing a path now. currently no one has .oxide files installed in their rM tablet. why create this format and put it in a shared directory and incur tech debt / more wor

There are already users who have installed 2.1 manually, or through the packages provided in the oxide release.

scripts/package-build
Eeems: I think I'll remove this and users can just use the less ideal "Import draft apps" menu option for everything other than KOReader.

package/koreader/package
```diff
@@ -36,18 +40,7 @@ package() {
     ln -s /opt/bin/fbdepth "$pkgdir"/opt/koreader/fbdepth
 
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$recipedir"/koreader.draft
+    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$recipedir"/KOReader.oxide
     install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
     install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
```
matteodelabre: You should use `$srcdir` here, since you added the files to the `source` array.
```suggestion
    install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$srcdir"/koreader.draft
    install -D -m 644 -t "$pkgdir"/opt/usr/share/applications/ "$srcdir"/KOReader.oxide
    install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons/ "$srcdir"/resources/koreader.png
    install -D -m 755 -t "$pkgdir"/opt/bin/ "$recipedir"/koreader
```

package/oxide/package
```diff
@@ -2,47 +2,50 @@
...
+        install -D -m 644 -t "$pkgdir"/opt/usr/share/applications "$srcdir"/opt/usr/share/applications/codes.eeems.fret.oxide
     }
 }
 
 oxide() {
     pkgdesc="Launcher application"
     url=https://github.com/Eeems/oxide/tree/master/applications/launcher
     section=launchers
-    depends=(erode fret tarnish xochitl rot)
+    depends=("erode  (= $pkgver)" "fret (= $pkgver)" "tarnish (= $pkgver)" "rot (= $pkgver)" "decay (= $pkgver)")
```
matteodelabre, line 42: What is the reasoning behind moving the `xochitl` dependency to `tarnish`?

package/oxide/package
```diff
@@ -69,25 +72,39 @@ tarnish() {
...
-        if systemctl list-units --full -all | grep -Fq 'tarnish.service'; then
-            echo "Disabling $pkgname"
-            systemctl disable --now tarnish
+        if is-active tarnish; then
+            echo "Stopping tarnish"
+            systemctl stop tarnish
+        fi
+        if is-enabled tarnish; then
+            echo "Disabling tarnish"
+            systemctl disable tarnish
```
matteodelabre, line 92: `systemctl disable --now tarnish` will stop the unit even if it was not enabled. So I’m not sure whether separating the two actions is actually needed here.

package/koreader/KOReader.oxide
```diff
@@ -0,0 +1,12 @@
+{
+  "displayName": "KOReader",
+  "description": "ebook reader",
```
matteodelabre: Would be best to sync this with the package description.
```suggestion
  "description": "An ebook reader application supporting PDF, DjVu, EPUB, FB2 and many more formats",
```

matteodelabre (changes requested):
I’m ok with adding `.oxide` files to the shared applications directory, since they are clearly identified with their extension. This is only temporary until a proper standard is agreed upon though. @Eeems committed to providing a transition path to the new standard once it’s available.

Here is my review on the actual changes.

package/koreader/KOReader.oxide
Eeems: Should this also be updated in the draft file? I copied from there.

package/oxide/package
Eeems: The oxide package doesn't actually depend on xochitl, tarnish does. Since in theory the oxide package could be provided by other packages in the future (alternative launcher screens) it's better to reflect that. Although perhaps rot should also be moved to tarnish.

package/oxide/package
Eeems: Because it might not be enabled so just doing a `is-enabled` check beforehand could miss that `is-active` is true.

package/koreader/KOReader.oxide
matteodelabre: Hah, never noticed that. I guess it’s as good a time as any to fix this. Thanks! :)

package/oxide/package
matteodelabre: Okay, thanks!

package/oxide/package
matteodelabre: I see the reasoning. This should also be applied to other packages that declare systemd services then, and maybe factored out int `install-lib`. But let’s keep this PR focused on Oxide.

matteodelabre (approved):

raisjn (approved):
flags is cool, nice addition!

what is xochitl.oxide about?

Eeems:
> 
> flags is cool, nice addition!

Thanks, I pulled it in from another PR that's been in limbo.

> 
> what is xochitl.oxide about?

Since the xochitl package provides xochitl.draft I thought it best to leave it in for now.

PR #200, opened 2021-01-07
Default image if missing

Eeems:
(Follow-up to #198)

### DIFF
--- a/scripts/package-build
+++ b/scripts/package-build
@@ -151,7 +151,7 @@ for pkgname in "${pkgnames[@]}"; do
             section "Stripping $pkgname"
             docker container run --interactive --rm \
                 --mount type=bind,src="$(realpath "$pkgdir")",dst=/pkg \
-                "$(image-name "$image")" /bin/bash \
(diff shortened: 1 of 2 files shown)


scripts/package-build
```diff
@@ -151,7 +151,7 @@ for pkgname in "${pkgnames[@]}"; do
             section "Stripping $pkgname"
             docker container run --interactive --rm \
                 --mount type=bind,src="$(realpath "$pkgdir")",dst=/pkg \
-                "$(image-name "$image")" /bin/bash \
+                "$(image-name "$([[ -n $image ]] && echo "base:v1.2.2" || echo "$image")")" /bin/bash \
```
matteodelabre: If I may suggest a terser version using Bash’s default value expansion:

```suggestion
                "$(image-name "${image:-base:v1.2.2}")" /bin/bash \
```

matteodelabre (dismissed):
Looking good. Sorry that we missed that case in the other PR!

matteodelabre (approved):

PR #205, opened 2021-01-09
Backport KOReader support for reMarkable 2

matteodelabre:
This patch contains the changes from koreader/koreader#6992, koreader/koreader#7066, and koreader/koreader#7065, which add support for reMarkable 2. They are expected to land in the future 2021.01 release.

### DIFF
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -24,6 +24,7 @@ sha256sums=(
 
 prepare() {
     patch -p1 -d"$srcdir" < "$srcdir"/rm2-support.patch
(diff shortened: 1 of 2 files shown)


matteodelabre:
Confirmed that it does not break support for rM1. Needs testing with rM2.

raisjn (dismissed):
tested on rM2, looks good. i noticed the rm2-support.patch file stays in /opt/koreader (which is fine)

matteodelabre:
Good catch. Easy enough to fix. Thanks for testing!

raisjn (approved):

PR #206, opened 2021-01-09
[build] only keep new packages, useful for PRs

raisjn:
this will keep PR artifact size down

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -49,7 +49,7 @@ jobs:
             - name: Install missing tools
               run: sudo apt-get install bsdtar tree
             - name: Build packages
-              run: make repo
+              run: make repo-new
             - name: Save the build output
               uses: actions/upload-artifact@v2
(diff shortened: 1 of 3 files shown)


raisjn:
notice that the artifact is only 21kb here (because there are no new packages in this PR)

i tested in my fork by bumping a sub-package in rmkit and the artifact contained only that sub-package. 

i feel that the deletion of the file could be handled better (or use a web request that just asks for the size of the file instead of downloading the whole thing when in $onlynew mode)

matteodelabre (dismissed):
This looks good. I added some small formatting fixes. You’re right that the deletion part could be improved, but as you may have seen while working with the `repo-build` script, we’re reaching the limits of expressivity of the Bash language IMO. I’m currently in the process of rewriting it to Python, which should make it a bit more readable and allow us to improve the deletion logic. In the meantime I think we can merge this.

matteodelabre (approved):

PR #208, opened 2021-01-10
[issue->pr] how-to-enable says to disable shutdown.target

### LINKED ISSUE #207: how-to-enable says to disable shutdown.target
image

We should probably filter out anything that isn't a *.service file.

LinusCDE:
Continuation of issue #207.

Fix by ignoring all conflicts that are not ending in ".service".

Should we filter in the `get-conflicts()` function instead? The docs of the function don't deny it returning non-service files.

### DIFF
--- a/scripts/install-lib
+++ b/scripts/install-lib
@@ -48,7 +48,7 @@ get-conflicts() {
 # Commands to run to achieve the desired result
 how-to-enable() {
     for conflict in $(get-conflicts "$1"); do
-        if is-enabled "$conflict"; then
+        if is-enabled "$conflict" && echo "$conflict" | grep -q '\.service$'; then
             echo "$ systemctl disable --now ${conflict/.service/}"
         fi
     done
--- a/scripts/install-lib
+++ b/scripts/install-lib
@@ -34,7 +34,7 @@ is-enabled() {
 # List of conflicting units
 get-conflicts() {
     systemctl show "$1" | awk -F'=' '/^Conflicts=/{print $2}' \
-        | awk 'BEGIN { ORS=" " }; { for(i = 1; i <= NF; i++) if($i != "shutdown.target") print $i }'
+        |  sed 's|\bshutdown.target\b||'
 }
 
 # Print instructions about how to enable a given systemd service and disable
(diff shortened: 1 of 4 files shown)


Eeems:
@matteodelabre Should probably weigh in on this. I'm not completely sure how we want to handle them either.

matteodelabre:
Good catch. Looking in the systemd manual, it seems that `shutdown.target` is the only conflict which is automatically added to the list of conflicts of all services by systemd itself (see <https://www.freedesktop.org/software/systemd/man/systemd.service.html#Automatic%20Dependencies>). So I think we can simply filter out `shutdown.target` from `get-conflicts()`’s output.

LinusCDE:
> Good catch. Looking in the systemd manual, it seems that `shutdown.target` is the only conflict which is automatically added to the list of conflicts of all services by systemd itself (see https://www.freedesktop.org/software/systemd/man/systemd.service.html#Automatic%20Dependencies). So I think we can simply filter out `shutdown.target` from `get-conflicts()`’s output.

Done. Probably just review and merge now.

scripts/install-lib
```diff
@@ -33,7 +33,8 @@ is-enabled() {
 #
 # List of conflicting units
 get-conflicts() {
-    systemctl show "$1" | awk -F'=' '/^Conflicts=/{print $2}'
+    systemctl show "$1" | awk -F'=' '/^Conflicts=/{print $2}' \
+        | awk 'BEGIN { ORS=" " }; { for(i = 1; i <= NF; i++) if($i != "shutdown.target") print $i }'
```
Eeems:
```suggestion
        |  sed 's|\bshutdown.target\b||'
```

LinusCDE: Didn't think that sed could do it in the sense of not interrupting a space separated list. If that works as well, I'm all for the change. Please review this as I'm blindly committing that from my phone (pc is shut down for today).

LinusCDE: Just testee it and, as guessed, it keeps both spaces where the shutdown.target was supposed to be.

Not the cleanest, but our purpose of looping over it doesn't care about it.

Basically the change would turn `a b shutdown.target c` into `a b  c` (two spaces between b and c) instead of `a b c` with awk.

matteodelabre (dismissed):
Looks good. I added a note about why we need to filter out `shutdown.target`.

matteodelabre (approved):
Just realized that we also need to bump the versions of the packages using this function.

PR #204, opened 2021-01-09
Stable merge window for week 02 of 2021

matteodelabre:
Tooling changes:

* Make `stable` the default branch.
* Strip binaries by default.
* Use `systemctl is-enabled` instead of `systemctl list-unit-files` to see whether a unit is enabled or not.

~~Installer changes:~~

* ~~Improve migration path from remarkable\_entware regarding the location of the `opt.mount` configuration file.~~
* ~~Keep existing branch configuration if an install of Toltec already exists.~~

Updated packages (not listing the packages that were only updated because of the strip changes):

* chessmarkable 0.5.1-1 → 0.6.0-2
* koreader 2020.10-2 → 2020.12-2
* oxide 2.0.1\~beta-3 → 2.0.3\~beta-1
* fret 2.0.1\~beta-3 → 2.0.3\~beta-1
* tarnish 2.0.1\~beta-3 → 2.0.3\~beta-1
* erode 2.0.1\~beta-3 → 2.0.3\~beta-1
* rot 2.0.1\~beta-3 → 2.0.3\~beta-1
* rm2fb 1.0.0-3 → 1.0.0-5
* remux 0.1.4-1 → 0.1.5-1

New packages:

* keywriter 0.1.0-1
* restream 0.0.0-1
* rm2-suspend-fix 0.0.0-1
* genie 0.1.0-1

PR #213, opened 2021-01-12
Add timezone mount to bootstrap

Eeems:
- [x] Update checksum
- [x] Test on device

### DIFF
--- a/package/toltec-bootstrap/entware-reenable
+++ b/package/toltec-bootstrap/entware-reenable
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # Path for the systemd unit that mounts Entware to /opt
-systemd_mount_path=/lib/systemd/system/opt.mount
+systemd_opt_mount_path=/lib/systemd/system/opt.mount
 
 # Path for the systemd unit that mounts /opt/share/zoneinfo to /usr/share/zoneinfo
 systemd_timezone_mount_path=/lib/systemd/system/timezone.mount
@@ -13,7 +13,7 @@ entware-mount() {
     mkdir -p "$entware_path"
 
     # Create systemd mount unit to mount over /opt on reboot
-    cat > "$systemd_mount_path" << UNIT
+    cat > "$systemd_opt_mount_path" << UNIT
 [Unit]
 Description=Bind mount Entware over /opt
 DefaultDependencies=no
--- a/package/toltec-bootstrap/entware-reenable
+++ b/package/toltec-bootstrap/entware-reenable
@@ -2,6 +2,9 @@
 # Path for the systemd unit that mounts Entware to /opt
 systemd_mount_path=/lib/systemd/system/opt.mount
 
+# Path for the systemd unit that mounts /opt/share/zoneinfo to /usr/share/zoneinfo
+systemd_timezone_mount_path=/lib/systemd/system/timezone.mount
+
 # Path where the actual Entware distribution resides (will be mounted to /opt)
 entware_path=/home/root/.entware
 
@@ -31,6 +34,42 @@ UNIT
     systemctl enable --now opt.mount 2> /dev/null
 }
 
+# Mount /opt/share/zoneinfo to /usr/share/zoneinfo
+timezone-mount() {
+    mkdir -p "$entware_path"/share/zoneinfo
+
+    if [[ ! -f $systemd_timezone_mount_path ]]; then
+        log "Updating timezone.mount location"
+
+        if systemctl -q is-enabled timezone.mount; then
+            # Remove path to old file if still symlinked
+            systemctl disable timezone.mount
+        fi
+    fi
+
+    # Create systemd mount unit to mount over /opt on reboot
+    cat > "$systemd_timezone_mount_path" << UNIT
+[Unit]
+Description=Bind mount /opt/share/zoneinfo over /usr/share/zoneinfo
+DefaultDependencies=no
+Conflicts=umount.target
+Before=local-fs.target umount.target
+After=opt.mount
+
+[Mount]
+What=/opt/share/zoneinfo
+Where=/usr/share/zoneinfo
+Type=none
+Options=bind
+
+[Install]
+WantedBy=local-fs.target
+UNIT
+
+    systemctl daemon-reload
+    systemctl enable --now timezone.mount 2> /dev/null
+}
+
 reinstall-root() {
     opkg update
     # Get the list of installed packages with files on root
@@ -48,6 +87,7 @@ reinstall-root() {
 
 main() {
     entware-mount
+    timezone-mount
     reinstall-root
 }
 
--- a/package/toltec-bootstrap/entware-reenable
+++ b/package/toltec-bootstrap/entware-reenable
@@ -3,7 +3,7 @@
 systemd_opt_mount_path=/lib/systemd/system/opt.mount
 
 # Path for the systemd unit that mounts /opt/share/zoneinfo to /usr/share/zoneinfo
-systemd_timezone_mount_path=/lib/systemd/system/timezone.mount
+systemd_timezone_mount_path=/lib/systemd/system/usr-share-zoneinfo.mount
 
 # Path where the actual Entware distribution resides (will be mounted to /opt)
 entware_path=/home/root/.entware
@@ -31,7 +31,7 @@ WantedBy=local-fs.target
 UNIT
 
     systemctl daemon-reload
-    systemctl enable --now opt.mount 2> /dev/null
+    systemctl enable --now "$(basename "$systemd_timezone_mount_path")" 2> /dev/null
 }
 
 # Mount /opt/share/zoneinfo to /usr/share/zoneinfo
@@ -57,7 +57,7 @@ WantedBy=local-fs.target
 UNIT
 
     systemctl daemon-reload
-    systemctl enable --now timezone.mount 2> /dev/null
+    systemctl enable --now "$(basename "$systemd_timezone_mount_path")" 2> /dev/null
 }
 
 reinstall-root() {
--- a/package/toltec-bootstrap/entware-reenable
+++ b/package/toltec-bootstrap/entware-reenable
@@ -31,7 +31,7 @@ WantedBy=local-fs.target
 UNIT
 
     systemctl daemon-reload
-    systemctl enable --now "$(basename "$systemd_timezone_mount_path")" 2> /dev/null
+    systemctl enable --now "$(basename "$systemd_opt_mount_path")" 2> /dev/null
 }
 
 # Mount /opt/share/zoneinfo to /usr/share/zoneinfo
--- a/package/toltec-bootstrap/entware-reenable
+++ b/package/toltec-bootstrap/entware-reenable
@@ -38,15 +38,6 @@ UNIT
 timezone-mount() {
     mkdir -p "$entware_path"/share/zoneinfo
 
-    if [[ ! -f $systemd_timezone_mount_path ]]; then
-        log "Updating timezone.mount location"
-
-        if systemctl -q is-enabled timezone.mount; then
-            # Remove path to old file if still symlinked
-            systemctl disable timezone.mount
-        fi
-    fi
-
     # Create systemd mount unit to mount over /opt on reboot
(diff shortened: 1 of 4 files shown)


Eeems:
@matteodelabre looks like the label check hasn't properly reported it's status?

matteodelabre:
Looks like you made your PR against `stable`, so it’s using the workflow files from that branch, which do not include the `pr-labels` workflow yet.

matteodelabre:
Fixing this requires rebasing your branch on `testing`, so I’ll let you do that.

Eeems:
Looks like I'll have to fix this up and recreate it. We should probably merge stable into testing every so often to make it easier to rebase when we accidentally do this.

Eeems:
@matteodelabre Okay, fixed

package/toltec-bootstrap/entware-reenable
```diff
@@ -2,6 +2,9 @@
 # Path for the systemd unit that mounts Entware to /opt
 systemd_mount_path=/lib/systemd/system/opt.mount
```
matteodelabre, line 3: I suggest making the variable naming more consistent:

```suggestion
systemd_opt_mount_path=/lib/systemd/system/opt.mount
```

matteodelabre: I suggest making the variable naming more consistent:

```suggestion
systemd_opt_mount_path=/lib/systemd/system/opt.mount
```

package/toltec-bootstrap/entware-reenable
```diff
@@ -31,6 +34,42 @@ UNIT
...
+    mkdir -p "$entware_path"/share/zoneinfo
+
+    if [[ ! -f $systemd_timezone_mount_path ]]; then
+        log "Updating timezone.mount location"
+
+        if systemctl -q is-enabled timezone.mount; then
+            # Remove path to old file if still symlinked
+            systemctl disable timezone.mount
+        fi
+    fi
```
matteodelabre: Why is this needed?

package/toltec-bootstrap/entware-reenable
```diff
@@ -31,6 +34,42 @@ UNIT
...
+    cat > "$systemd_timezone_mount_path" << UNIT
+[Unit]
+Description=Bind mount /opt/share/zoneinfo over /usr/share/zoneinfo
+DefaultDependencies=no
+Conflicts=umount.target
+Before=local-fs.target umount.target
+After=opt.mount
+
+[Mount]
+What=/opt/share/zoneinfo
```
matteodelabre: Is the extra level of indirection really needed here? If not, we could also remove the `After=opt.mount` dependency above.

```suggestion
What=/home/root/.entware/share/zoneinfo
```

scripts/bootstrap/bootstrap
```diff
@@ -345,11 +384,13 @@ main() {
         else
             log "Re-enabling existing Entware install"
             entware-mount
+            timezone-mount
         fi
     else
         log "Creating $entware_path and mounting to /opt"
         entware-mount
         entware-install
```
matteodelabre, line 378: Let’s keep the verbosity level consistent.

```suggestion
        entware-install
        log "Mounting timezone info over /usr/share/timezone"
```

matteodelabre (changes requested):
The duplication between the bootstrap script and the toltec-bootstrap package is something we will need to fix as soon as possible. But this is not the focus of this PR.

package/toltec-bootstrap/entware-reenable
Eeems: In case we change the location. I copied from the entware-mount method.

package/toltec-bootstrap/entware-reenable
Eeems: Fair enough

package/toltec-bootstrap/entware-reenable
matteodelabre: Assuming we changed the location in the future, this would not remove the old file (as opposed to what is done in `entware-mount`). I propose we wait until we actually change the location to add this block, then we’ll know what the old name is and will be able to clean it.

Eeems: Sounds good

matteodelabre:
Thanks @Eeems! I wanted to test your changes but I ran into [this error](https://github.com/toltec-dev/toltec/pull/204#issuecomment-758865591) which seems to have been introduced before your PR.

matteodelabre:
Turns out `systemd` refuses to enable the bind mount unless it’s named after the target path (so `usr-share-zoneinfo.mount`). I changed that and also made sure that the new unit is cleaned up in `error-cleanup()`.

matteodelabre (dismissed):
Tested and confirmed to work on my rM. Waiting on @Eeems to review my changes before merging.

package/toltec-bootstrap/entware-reenable
```diff
@@ -28,7 +31,33 @@ WantedBy=local-fs.target
 UNIT
 
     systemctl daemon-reload
-    systemctl enable --now opt.mount 2> /dev/null
+    systemctl enable --now "$(basename "$systemd_timezone_mount_path")" 2> /dev/null
```
Eeems:
```suggestion
    systemctl enable --now "$(basename "$systemd_opt_mount_path")" 2> /dev/null
```

matteodelabre:
Well spotted. Not sure how I missed that during testing, I must have used a version where I hadn’t made the change yet.

matteodelabre (dismissed):
Confirmed to work on my rM1.

Eeems:
@raisjn could you test on your rM2?

Eeems:
Test scenarios:

- Test upgrading the toltec-boostrap package
- Run bootstrap on existing install
- Run entware-reenable on existing install where it's disabled
- Run entware-reenable on existing install where it's already enabled
- Run bootstrap on device without entware/toltec

LinusCDE:
on rM 2 in order of testing:

 - Run bootstrap on device without entware/toltec » :heavy_check_mark: 
 - Run bootstrap on existing install » :heavy_check_mark: 
 - Run entware-reenable on existing install where it's disabled » :x: see logs below
 - Run entware-reenable on existing install where it's already enabled » :x: see same logs below

```shell
reMarkable: ~/ bash eetest 
Downloading https://bin.entware.net/armv7sf-k3.2/Packages.gz
Updated list of available packages in /opt/var/opkg-lists/entware
Downloading https://toltec-dev.org/testing/Packages.gz
Updated list of available packages in /opt/var/opkg-lists/toltec
eetest: command substitution: line 75: syntax error near unexpected token `|'
eetest: command substitution: line 75: `| uniq)'
```

 - Test upgrading the toltec-boostrap package » (:heavy_check_mark:) Worked but also triggered a re-installation of entware for me.

LinusCDE:
We also may want to trigger the timezone check when updating the `toltec-bootstrap` package. Haven't checked whether this is already done but seemed not to for me when `opkg update/upgrade`ing from stable to testing now at least.

matteodelabre:
> * Run entware-reenable on existing install where it's disabled » x see logs below
> * Run entware-reenable on existing install where it's already enabled » x see same logs below
> 
> ```console
> reMarkable: ~/ bash eetest 
> Downloading https://bin.entware.net/armv7sf-k3.2/Packages.gz
> Updated list of available packages in /opt/var/opkg-lists/entware
> Downloading https://toltec-dev.org/testing/Packages.gz
> Updated list of available packages in /opt/var/opkg-lists/toltec
> eetest: command substitution: line 75: syntax error near unexpected token `|'
> eetest: command substitution: line 75: `| uniq)'
> ```

Interesting. I re-ran entware-reenable on my rM1 in both scenarios and I don’t get any error. Can you confirm that we have the same `entware-reenable` script? Mine has a SHA-256 sum of `701173003cda882067d97d2f7f9fb8a18d2c17ac888e9646dee5acda3f34ae67`. Can you also provide us with the Bash version that’s running on the rM2? On rM1, I have 4.4.12(1).

> * Test upgrading the toltec-boostrap package » (heavy_check_mark) Worked but also triggered a re-installation of entware for me.

I don’t think just upgrading toltec-bootstrap should trigger the reenable script, since the package itself doesn’t contain any install script. Can you confirm @Eeems?

Eeems:
> I don’t think just upgrading toltec-bootstrap should trigger the reenable script, since the package itself doesn’t contain any install script. Can you confirm @Eeems?

Correct, this shouldn't happen.



> on rM 2 in order of testing:
> 
>     * Run bootstrap on device without entware/toltec » heavy_check_mark
> 
>     * Run bootstrap on existing install » heavy_check_mark
> 
>     * Run entware-reenable on existing install where it's disabled » x see logs below
> 
>     * Run entware-reenable on existing install where it's already enabled » x see same logs below
> 
> 
> ```shell
> reMarkable: ~/ bash eetest 
> Downloading https://bin.entware.net/armv7sf-k3.2/Packages.gz
> Updated list of available packages in /opt/var/opkg-lists/entware
> Downloading https://toltec-dev.org/testing/Packages.gz
> Updated list of available packages in /opt/var/opkg-lists/toltec
> eetest: command substitution: line 75: syntax error near unexpected token `|'
> eetest: command substitution: line 75: `| uniq)'
> ```
> 
>     * Test upgrading the toltec-boostrap package » (heavy_check_mark) Worked but also triggered a re-installation of entware for me.

Line 75 is `echo -e "${packages[@]}\n${available_packages[@]}" | sed 's/ /\0/g' | sort | uniq -d | awk '{print $1}' | xargs -r opkg install --force-reinstall`, which is a line I didn't touch. It also looks good to me. Perhaps `$available_packages` and `$packages` are empty?

LinusCDE:
> Interesting. I re-ran entware-reenable on my rM1 in both scenarios and I don’t get any error. Can you confirm that we have the same `entware-reenable` script? Mine has a SHA-256 sum of `701173003cda882067d97d2f7f9fb8a18d2c17ac888e9646dee5acda3f34ae67`. Can you also provide us with the Bash version that’s running on the rM2? On rM1, I have 4.4.12(1).
> 

> Line 75 is `echo -e "${packages[@]}\n${available_packages[@]}" | sed 's/ /\0/g' | sort | uniq -d | awk '{print $1}' | xargs -r opkg install --force-reinstall`, which is a line I didn't touch. It also looks good to me. Perhaps `$available_packages` and `$packages` are empty?

Got the problem. My fault. I took the latest file from here, but copy&pasted the content into nano which made a long line wrap. If I fix that the file matches with your hash, Mattéo.

That was my change. Makes sense that the command didn't work:
```diff
72,73c72
<     available_packages=$(gunzip -c /opt/var/opkg-lists/* | grep -E '(Package:)|(Version:)' | sed 's|Package: ||' | tr '\n' ' ' | sed -r 's| Version: ([^ ]+) | \1\n|g' | sort 
< | uniq)
---
>     available_packages=$(gunzip -c /opt/var/opkg-lists/* | grep -E '(Package:)|(Version:)' | sed 's|Package: ||' | tr '\n' ' ' | sed -r 's| Version: ([^ ]+) | \1\n|g' | sort | uniq)
```

I re-tested:
 - Run entware-reenable on existing install where it's disabled
 - Run entware-reenable on existing install where it's already enabled

For my test, I had remux installed, disabled opt.mount and usr-share-zoneinfo.mount and removed these unit files as well. The first reenable didn't reinstall remux to add the file again. I added rm2fb and installed remux (which was listed as removed?). After that I ran the second test and it reinstalled remux and rm2fb.

I would say this is not an issue with this specific PR, but entware-reenable might not reinstall if the firmware gets updated and disables toltec/entware.

Eeems:
> I would say this is not an issue with this specific PR, but entware-reenable might not reinstall if the firmware gets updated and disables toltec/entware.

Should we try to fix it as part of this, or as part of a different PR?

LinusCDE:
> > I would say this is not an issue with this specific PR, but entware-reenable might not reinstall if the firmware gets updated and disables toltec/entware.
> 
> Should we try to fix it as part of this, or as part of a different PR?

Ok, scratch that as well. Redid the test with faking a fw update and now it worked fine. Not sure what I missed in the last test.

LinusCDE (approved):

PR #214, opened 2021-01-12
Fix regressions in bootstrap

LinusCDE:
See https://github.com/toltec-dev/toltec/pull/204#issuecomment-758865591

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To automatically install Opkg, Entware and Toltec, run the bootstrap script in a
 
 ```sh
 $ wget http://toltec-dev.org/bootstrap
-$ echo "e58e37dd65f180cd6079cb273c31a0bbfc30c10e6bbc5ba4c95bd515ffc586f7  bootstrap" | sha256sum -c
+$ echo "0762d8dbb49ea381b03324ccfb6e35d68fc66c58f2cee1f7b2c1f8ff95ccffef  bootstrap" | sha256sum -c
 $ bash bootstrap
 ```
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To automatically install Opkg, Entware and Toltec, run the bootstrap script in a
 
 ```sh
 $ wget http://toltec-dev.org/bootstrap
-$ echo "3eefd43f5c80ea4b7b8bb60475a5ebe23d895fc2eb12d3174d1b7594a6a0c6d5  bootstrap" | sha256sum -c
+$ echo "b028b7449c36216e82b6a0d3733ed870d8896d5ba407fa6778fcee3a2d6c8fd5  bootstrap" | sha256sum -c
 $ bash bootstrap
 ```
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To automatically install Opkg, Entware and Toltec, run the bootstrap script in a
 
 ```sh
 $ wget http://toltec-dev.org/bootstrap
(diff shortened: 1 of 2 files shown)


LinusCDE:
Regarding the comment from @matteodelabre  linked above:

> `cat: can't open '/home/root/.entware/etc/opkg.conf': No such file or directory`

The initial check I made included a check for the existance of the file and would return the assumed standard or set branch. I failed to notice the regression in @Eeems [suggested and included change](https://github.com/toltec-dev/toltec/pull/176#discussion_r550920101) that also checked the specified repo name to be "toltec" which I ignored.

> ```
> INFO:  Updating opt.mount location
> Failed to get unit file state for opt.mount: No such file or directory
> rm: can't remove '/etc/systemd/system/opt.mount': No such file or directory
> ```

This was caused by me messing up this if-statement: `if [[ ! -f $old_systemd_mount_path ]] && [[ ! -f $systemd_mount_path ]]; then` which migrated only when there was never anything to migrate. I guess that the first error came from disabling the opt.mount unit and the rm failed to find the file that the if should've ensured existed.

matteodelabre:
I think we can make `get-branch` use less subshells and pipes:

```sh
# Find out used branch or get default one
get-branch() {
    local opkg_conf="$entware_path/etc/opkg.conf"
    if [[ -f "$opkg_conf" ]]; then
        while read -r line; do
            # Is this the toltec repo?
            if [[ "$(echo "$line" | awk '{print $2}')" == "toltec" ]]; then
                local url
                url=$(echo "$line" | awk '{print $3}')
                if ! echo "$url" | grep -q https://toltec-dev.org/; then
                    # If it's the wrong domain skip, we don't know how to get the branch
                    continue
                fi
                echo "${url/'https://toltec-dev.org/'/}"
                return
            fi
        done < <(grep "^src/gz" "$opkg_conf")
    fi
    # Default to stable if there was no result
    echo stable
}
```

LinusCDE:
We can probably make it even shorter.

Not sure if the extend of regex hurts more then it helps, but having just one regex do the search and extraction.

```bash
# Find out used branch or get default one
get-branch() {
    local opkg_conf="$entware_path/etc/opkg.conf"
    local branch
    if [[ -f "$opkg_conf" ]]; then
        # Get the first (...) value (aka \1) of the pattern if matching
        branch=$(sed -E 's|^src/gz\s+toltec\s+https://toltec-dev.org/([[:alnum:]]*).*$|\1|;t;d' "$opkg_conf" | head -n1)
    fi
    # Default empty/missing value to "stable"
    [[ -n "$branch" ]] && echo $branch || echo stable
}
```

It seems to work in all scenarios the user could throw in the file like:
 - trailing slash or everything else (not fixed by previous solution)
   - `src/gz toltec https://toltec-dev.org/testing/`
   - `src/gz toltec https://toltec-dev.org/testing/ # comment, even if maybe not supported`
 - Commented out lines
 - (Multiple) Tabs or spaces
 - Same repo defined multiple times (first picked with `head -n1`)

matteodelabre:
Looks good. I did some further tests:

* opkg does not support comments at the end of the `src/gz` line (it ignores the line).
* opkg does not support multiple `src` declarations with the same name, it only considers the first.

Your implementation therefore works great with all opkg-supported cases. I think we can also simplify the last line to `echo "${branch:-stable}"`. I’ll update the PR to include that.

matteodelabre (approved):
Tested on my rM and confirmed to work.

PR #215, opened 2021-01-13
[build] only build new artifacts during PRs should work for stable too

raisjn:


matteodelabre (approved):
LGTM

raisjn:
this still needs testing (i guess i can put it in a PR on stable and see what happens)

matteodelabre:
I tested the command locally, replacing the placeholder alternatively with stable and testing, and it produces the expected result.

Eeems:
Will this be a problem with the rsync --delete changes?

PR #216, opened 2021-01-13
Add version link

raisjn:
Make the version number into a download link on the web listing page. This makes it easier to link other people to packages


matteodelabre (approved):
Good idea. Looks good to me.

Issue #211, opened 2021-01-10
Rewrite tooling scripts in Python


matteodelabre:
I’m currently rewriting the tooling scripts from Bash to Python.

Goals of the rewrite:

* Contain it to only the tooling scripts. Package recipes should be unaffected by the change.
* Make it easier to add features to the tooling scripts, which is quite a challenge in their current Bash form.
* Add more sanity checks when parsing and building recipes.

If you’d like to suggest a specific feature for the rewrite, feel free to comment below. I’d like to keep the rewrite as focused as possible so that we can quickly merge it and move to other more productive things.

Eeems:
#85 #210 #8 and perhaps poke at #45

matteodelabre:
In the interest of not blocking other tasks that depend on this, I’m planning on splitting this in three PRs:

- [x] Rewrite of `package-build` and `repo-build` (#218).
- [x] Rewrite of `repo-build-web` (#266).
- [ ] Rewrite of `repo-check`.

matteodelabre:
Closing this since almost everything has been rewritten to Python, except for the `repo-check` script, which is ununsed anyway since we haven’t really integrated reproducible builds into our workflow yet.

PR #218, opened 2021-01-13
Rewrite repo-build and package-build in Python

matteodelabre:
(See #211.)

The new build scripts are backward-compatible with existing recipes. Apart from some minor improvements on generated maintainer scripts, there are no changes in the final .ipk files.

* Rewrite to Python 3.8
    - Create `toltec` Python support library (in `scripts/toltec`)
    - Remove superseded `opkg` Python library (in `scripts/opkg`)
    - Add Python linting and formatting checks
    - Replace `scripts/repo-build` (Bash script) with `scripts/repo_build.py`
    - Replace `scripts/package-build` (Bash script) with `scripts/package_build.py`
    - Update Makefile and workflows to work with the new scripts
* Package changes
    - Update koreader to not use the `$recipedir` variable (not documented, not supported by the new scripts, and should not be used anymore)
    - Update zoneinfo-utils to fix use of an undefined variable
* Documentation changes
    - Add info on how required Python version and modules
    - Document custom fields and functions support
* New features
    - Improve support for custom fields and functions in recipes
    - Parameterize the <https://toltec-dev.org> value in workflows, using a `REMOTE_HTTP` secret.
    - Rename the `REMOTE` secret to `REMOTE_SSH` for clarity
* Additional checks
    - Check that all custom declarations start with `_` to guard against typos on optional standard declarations
    - Check that timestamp fields contain a valid ISO-8601 value
    - Check that pkgver fields contain a valid version number
    - Run all Bash scripts with `set -euo pipefail` flags enabled


.github/workflows/pr.yml
```diff
@@ -35,6 +35,9 @@ jobs:
                 chmod a+x shellcheck
                 sudo chown root:root shellcheck
                 sudo mv shellcheck "$install_dir"
+
+                # Install black
+                pip install black==20.8.b1
```
Eeems:
```suggestion
                pip install black==20.8b1
```

matteodelabre: Thanks!

package/koreader/package
```diff
@@ -18,12 +18,14 @@ source=(
     koreader.draft
     KOReader.oxide
     rm2-support.patch
+    koreader
```
raisjn, line 21: what is this doing?

scripts/toltec/bash.py
```diff
@@ -0,0 +1,379 @@
...
+            raise ValueError(
+                f"Unsupported type {type(value)} for variable \
+{name}"
+            )
+
+    return result
+
+
+def _parse_string(token: str) -> str:
+    """Remove escape sequences from a Bash string."""
```
raisjn, line 194: is this the only escape sequence that needs to be removed?

raisjn (commented):
how are you testing this? did you generate full repo and compare with the old repo?

matteodelabre:
> how are you testing this? did you generate full repo and compare with the old repo?

I generated the repo, installed the packages and tested them. I haven’t automatically checked that the `data.tar.gz` sub-archives match with the existing ones yet.

package/koreader/package
matteodelabre: `$recipedir` is not supported by the new implementation, since it was just an implementation detail of the Bash tooling. This change makes the `koreader` recipe use the correct mechanism for importing sources from the recipe directory.

scripts/toltec/recipe.py
```diff
@@ -0,0 +1,225 @@
...
+        """Get the control fields for this package."""
+        control = f"""Package: {self.name}
+Version: {self.pkgver}
+Maintainer: {self.parent.maintainer}
+Section: {self.section}
+Architecture: {self.arch}
+Description: {self.pkgdesc}
+HomePage: {self.url}
+License: {self.license}
+"""
```
LinusCDE:
```suggestion
        control = inspect.cleandoc(f"""
            Package: {self.name}
            Version: {self.pkgver}
            Maintainer: {self.parent.maintainer}
            Section: {self.section}
            Architecture: {self.arch}
            Description: {self.pkgdesc}
            HomePage: {self.url}
            License: {self.license}
            """)
```

(need to add `inspect` to the imports for it to work!)

Not sure if wanted, but can get indented better and still get almost the same result (no trailing newline) with `inspect.cleandoc`. Alternatively `textwrap.dedent` can be used as well but this wont remove the first newline here (maybe add lstrip() as well to remove it).

scripts/package_build.py
```diff
@@ -0,0 +1,39 @@
...
+"""Build packages from a given recipe."""
+
+import argparse
+import logging
+import sys
+import docker
+from toltec.builder import Builder
+from toltec.util import argparse_add_verbose, LOGGING_FORMAT
+
+parser = argparse.ArgumentParser(description=sys.modules[__name__].__doc__)
```
LinusCDE: This can get shortened

```suggestion
parser = argparse.ArgumentParser(description=__doc__)
```

(If you want to have a multiline docstring that is properly displayed in the help as well, specify `formatter_class=argparse.RawDescriptionHelpFormatter` as well (otherwise newlines get ignored).)

scripts/repo_build.py
```diff
@@ -0,0 +1,60 @@
...
+
+import argparse
+import logging
+import sys
+import docker
+from toltec.builder import Builder
+from toltec.repo import Repo
+from toltec.util import argparse_add_verbose, LOGGING_FORMAT
+
+parser = argparse.ArgumentParser(description=sys.modules[__name__].__doc__)
```
LinusCDE: This can get shortened (same as in `package_build.py`)

```suggestion
parser = argparse.ArgumentParser(description=__doc__)
```

LinusCDE (commented):
I have not checked out all the code yet. It seems really impressive so far!

The comments only only some minor things I found. I need to dig into the new code deeper and do some more

The data.tar.gz currently contains absolute paths only. I'll make a fix for that.
grafik

scripts/toltec/bash.py
matteodelabre: As far as I know, yes. The lexer already removes all escape sequences. But the Bash `declare` builtin seems to add an additional escape in front of dollar signs in strings, even where it’s not necessary. Unfortunately this is not documented anywhere I could find.

scripts/package_build.py
matteodelabre: Oh, nice, didn’t know this could be accessed directly. Thanks!

matteodelabre:
Thanks for your review @LinusCDE.

> The data.tar.gz currently contains absolute paths only. I'll make a fix for that.

It worked before and then I switched to using absolute paths, which triggers a special behavior in `tarfile`. Should be fixed now!

scripts/toltec/recipe.py
matteodelabre: Nice, I’ll use that, thanks! This can also be useful in `Builder._archive`.

LinusCDE:
Cool. Just first fixed the paths, then redid the adding of files manually, since it seemed to me they were empty. I was stupid (looked at a symlink) and the work was for nothing. xD

Ideally I would like to replace `str.removeprefix` with `os.path.relpath` in `_clean_info(...)`.

matteodelabre:
> Ideally I would like to replace `str.removeprefix` with `os.path.relpath` in `_clean_info(...)`.

That’s also what I did in my fix. :-)

.github/workflows/pr.yml
```diff
@@ -21,10 +25,14 @@ jobs:
...
-            - name: Setup Toltec environment
+            - name: Setup Python
+              uses: actions/setup-python@v2
+              with:
+                python-version: '3.x'
+            - name: Setup Toltec dependencies
               uses: ./.github/actions/setup
             - name: Build packages
-              run: remote_repo='https://toltec-dev.org/${{ github.base_ref }}' make repo-new
+              run: make repo-new FLAGS='--remote-repo https://toltec-dev.org/${{ github.base_ref }}'
```
raisjn: i know it's not related, but both linus and I have separate repos we use for testing (hosted on different URLs than toltec) - would it be possible to make the URL configurable here? (we can do in a separate commit, but it just came to mind while reviewing)

Makefile
```diff
@@ -40,35 +40,43 @@ help:
...
 	./scripts/repo-check build/repo
 
 $(RECIPES): %:
-	./scripts/package-build package/"${@}" build/package/"${@}"
+	./scripts/package_build.py $(FLAGS) "$(@)"
 
-$(RECIPES_PUSH): %:
-	ssh root@"${HOST}" mkdir -p .cache/opkg
-	scp build/package/"$(@:%-push=%)"/*/*.ipk root@"${HOST}":.cache/opkg
+push: %:
```
raisjn, line 64: how does this fit with local development? is your rM setup to point at .cache/toltec as a repository or do you manually install packages out of .cache/toltec?

scripts/repo_build.py
```diff
@@ -0,0 +1,57 @@
...
+parser.add_argument(
+    "-n",
+    "--no-fetch",
+    action="store_true",
+    help="do not fetch missing packages from the remote repository",
+)
+
+argparse_add_verbose(parser)
+
+group = parser.add_mutually_exclusive_group()
```
raisjn, line 23: cool feature

raisjn (commented):
overall, your python looks good! i don't have much to say about the code itself (i've read through it a few times), just comments+questions on usage and how things fit together

.github/workflows/pr.yml
matteodelabre: Good point. Would using a secret work for your use case?

raisjn: yes, definitely - but a secret with default to toltec?

matteodelabre: I don’t think that there’s a way to add a default value to a GitHub Action expression unfortunately. But it should be fairly easy to set that secret in the Toltec org to point to our server.

Makefile
matteodelabre: This pushes the whole repo, including the Packages index, so that you can add /home/root/.cache/toltec as a local repository in `opkg.conf`. The downside is that since it’s using scp instead of rsync, it pushes the whole set of packages everytime. Couldn’t get rsync to work in the PC→rM direction.

.github/workflows/pr.yml
matteodelabre: Added a new secret called `REMOTE_HTTP` which parameterizes `https://toltec-dev.org`. Renamed the existing `REMOTE` secret to `REMOTE_SSH` for clarity.

matteodelabre:
Thanks for your review @raisjn!

Makefile
raisjn: `rsync --rsync-path /opt/bin/rsync blah blah`

raisjn:
i'm running this on python 3.8.5 on my machine locally, misc notes:

* `from collections.abc import Iterable` causes an error about ABCMeta, using `from typing import Iterable` works
* keeping the package dir during build causes an error because there are mkdir lines that throw errors (src_dir). consider using makedirs() with exist_ok=True for most of the creates. 
* `AttributeError: 'str' object has no attribute 'removeprefix'` - https://www.python.org/dev/peps/pep-0616/ - looks like it is python 3.9 feature added a year ago?
* ~~the output is missing the tree (see details below) in rmkit package~~ nvm, this is because i deleted the lines that used removeprefix

Makefile
matteodelabre: Not sure how I missed that flag. Thanks for the heads-up!

matteodelabre:
> i'm running this on python 3.8.5 on my machine locally, misc notes:
> 
> * `from collections.abc import Iterable` causes an error about ABCMeta, using `from typing import Iterable` works
> * `AttributeError: 'str' object has no attribute 'removeprefix'` - https://www.python.org/dev/peps/pep-0616/ - looks like it is python 3.9 feature added a year ago?

I think we’re seeing differences between 3.8 and 3.9 (I used the latter for testing). Should be fairly easy to make the code 3.8-compatible.

matteodelabre:
Thanks for the report. Should be fixed now.

raisjn (dismissed):
shipit

matteodelabre:
Thanks for the review! I’ll do a final test of all packages to make sure everything works fine tomorrow, and then I’ll merge this.

matteodelabre:
I went through all the packages and checked for differences between the versions generated with the new scripts and the versions existing in stable. I found some bugs that I addressed in the second commit of this branch (which is the diff since you last reviewed @raisjn).

There remains no difference in the `data.tar.gz` archives that contain the installed files. There are some differences in the generated maintainer scripts in `control.tar.gz` since the generation was slightly improved with the rewrite. I have validated on my rM1 that those differences do not cause problems. I don’t believe that those differences could have an impact on rM2.

This is now awaiting a final review before merging.

raisjn (approved):
looks good to me, whats the roll-out plan? 

i tested by building the whole repository (repo-local), then pushed packages and tested installation and removal of various packages: rm2fb, oxide, remux, fingerterm, draft, etc

matteodelabre:
Could you clarify what you mean by roll-out plan in this context @raisjn?

matteodelabre:
Summary of our Discord discussion: @raisjn was “wondering if it makes sense to build a repo using the new build scripts while keeping the old repo, then have them run side by side and switch over down the line (so it gets live testing of packages). but after thinking about it, its much work”. @matteodelabre: “If we merge it into testing, we'll have a bit of time to look for remaining issues before it hits stable. Those remaining bugs should be fairly minor and affect relatively few users.”

PR #220, opened 2021-01-14
Update plato to 0.9.13-1

LinusCDE:
- Updated base version from 0.9.1 to 0.9.13
- Added "Ignored Buttons" to main menu

[(Changelog)](https://github.com/LinusCDE/plato/releases/tag/0.9.13-rm-release-7)

The version should most likely function the same on the rM 2 as the release before it. Probably just check that it runs any maybe whether buttons are disabled and the settings saved.

EDIT: One variation the rM 2 should have is that only the "Power" button should be shown in "Disabled Buttons".

### DIFF
--- a/package/plato/package
+++ b/package/plato/package
@@ -5,15 +5,15 @@
 pkgnames=(plato)
 pkgdesc="Document reader"
 url=https://github.com/LinusCDE/plato
-pkgver=0.9.1-13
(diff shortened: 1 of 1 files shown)


raisjn (approved):
lgtmacro

verified that the ignored button menu only contains 'power' as an option

PR #221, opened 2021-01-14
templatectl: Add/Remove items from templates.json via cli

PeterGrace:
This PR adds a package file for templatectl, which is a rust-based binary that interacts with the `/usr/share/remarkable/templates/templates.json` file.  It can add entries and remove entries from the file, sparing the end user the need to edit the json file manually.

Additionally, since it is a cli-based app and will be installable via toltec, we can then use it to package template files in toltec for installing to xochitl from the package repository.

Issue #109, opened 2020-10-06
Automate re-enabling of Toltec after system upgrades


raisjn:
when opkg/toltec is installed and there is a system upgrade, we need a way to re-enable opkg. the entware package comes with entware_reinstall which re-creates the opt mounting system service. i believe our bootstrap script (thanks linusCDE!) automatically re-enables the opkg as well. (is the bootstrap script saved locally for re-use after upgrade?)

currently, we do not have any hooks into the remarkable's upgrade process to automatically re-enable opkg. 

options:

* try to get hook in upgrade process
* add something to .bashrc (or similar) that notifies user when they ssh in that opkg isn't enabled and print way to re-enable it
* ???

matteodelabre:
The current bootstrap script [creates a script called `entware-reenable`](https://github.com/toltec-dev/toltec/blob/36bdcfdc7162054ff0f3553576c7fe3733b30ccb/scripts/bootstrap/bootstrap#L324) in the home folder, and shows “Use '/home/root/entware-reenable' to re-enable Toltec after a system update” to the user after install. Automatically detecting in .bashrc when it needs to be run seems like a good idea.

LinusCDE:
I think @Eeems had the idea to add a specific feature to reMarkables [update_engine](https://github.com/reMarkable/update_engine). One would need to:
 - Add the feature to their update_engine
 - Create a PR there
 - Hope that they merge it

Maybe creating an issue there would be a better first step, to ask if they would even want to allow such a feature.

matteodelabre:
Have we found a way to downgrade to a previous version to test the upgrade path? The only way I know of currently is flashing using uuuflash.

Eeems:
Flashing a previous version to the other root partition while live and then switching partitions.

matteodelabre:
Do you know if this is documented somewhere?

Eeems:
> 
> 
> Do you know if this is documented somewhere?

Not that I'm aware of.

matteodelabre:
Update: This is now easily doable thanks to ddvk’s [remarkable-update](https://github.com/ddvk/remarkable-update) project (works on rM1 and rM2).

Eeems:
https://github.com/Jayy001/codexctl implements full on-device upgrade/downgrade handling, and would be extendable to run post-upgrade scripts for us. At some point soon, I'll add codexctl as a toltec package for us when I sort out automating the build for the device binary.

Eeems:
[...]

PR #222, opened 2021-01-14
templatectl: Add/Remove items from templates.json via cli 

PeterGrace:
(Re-open of PR #221 with proper branch targeting.)

This PR adds a package file for templatectl, which is a rust-based binary that interacts with the `/usr/share/remarkable/templates/templates.json` file. It can add entries and remove entries from the file, sparing the end user the need to edit the json file manually.

Additionally, since it is a cli-based app and will be installable via toltec, we can then use it to package template files in toltec for installing to xochitl from the package repository.


LinusCDE (dismissed):
works like a charm :+1:

package/remarkable_math_graph/package
```diff
@@ -0,0 +1,45 @@
...
+build() {
+    export DEBIAN_FRONTEND=noninteractive
+    apt-get update -y
+    apt-get install -y \
+        --no-install-recommends \
+        -o Dpkg::Options::="--force-confdef" \
+        imagemagick librsvg2-bin
+	convert -scale 26.464% remarkable_math_graph_qcadexport.svg \
+        remarkable_math_graph.svg
+    convert remarkable_math_graph.svg remarkable_math_graph.png
```
Eeems:
```suggestion
	  convert -scale 26.464% remarkable_math_graph_qcadexport.svg \
          remarkable_math_graph.svg
      convert remarkable_math_graph.svg remarkable_math_graph.png
```

package/remarkable_math_graph/package
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(remarkable_math_graph)
```
Eeems: Maybe have template in the name?
```suggestion
pkgnames=(template-math-graph)
```

package/remarkable_math_graph/package
```diff
@@ -0,0 +1,45 @@
...
+        --no-install-recommends \
+        -o Dpkg::Options::="--force-confdef" \
+        imagemagick librsvg2-bin
+	convert -scale 26.464% remarkable_math_graph_qcadexport.svg \
+        remarkable_math_graph.svg
+    convert remarkable_math_graph.svg remarkable_math_graph.png
+}
+
+package() {
+    install -D -m 755 -t "$pkgdir"/usr/share/remarkable/templates \
```
Eeems: Perhaps these should be stored in /opt and templatectl handles copying them in?

Eeems (dismissed):

package/remarkable_math_graph/package
danshick: Agreed. Symlinks are even better.

PeterGrace: If we're going to be copying it to the right spot eventually, why would we first copy it to /opt and then obfuscate its installation by doing it inside of a program that the user is unaware of the intermediate steps for?

templatectl is meant to interact with templates.json only.  It's not meant as an installer for templates, but as a way to properly format xochitl's datafile that it uses to load templates.  I'd prefer to keep its use case succinct.

Eeems: Having them deploy to /usr requires them to be reinstalled on device upgrade. If we have a location in /opt, we could easily iterate through the items and reinstall them as part of the reenable script.

PeterGrace: OK -- then I'm in agreement with @danshick  then -- install into /opt/templates and the the enable scripting will ensure that a symbolic links exist?

Eeems: `/opt/share/templates` or similar would probably be better. The reeneable script could then loop through any package that has files in those locations and rerun the configure call to add the symlink and get `templatectl` to add them to the json file.
@matteodelabre thoughts?

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
...
+depends=(templatectl)
+
+image=qt:v1.2.2
+source=(https://github.com/PeterGrace/template-cartesian-graph/archive/v1.0.0.zip)
+sha256sums=(d3215546c6fd709bb5988b7042ce5fb990d9a030d783ed7505fcc2dbedbc0e0f)
+
+build() {
+    echo "no build steps at this time"
+}
+
```
matteodelabre: Since you’re not using the `build()` step, you can simply not define it and remove the `image` field. This will save the cost of a Docker container instantiation.

```suggestion
source=(https://github.com/PeterGrace/template-cartesian-graph/archive/v1.0.0.zip)
sha256sums=(d3215546c6fd709bb5988b7042ce5fb990d9a030d783ed7505fcc2dbedbc0e0f)

```

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
...
+url=https://github.com/PeterGrace/template-cartesian-graph
+pkgver=1.0.0-4
+timestamp=2021-01-14T17:46Z
+section=utils
+maintainer="Peter Grace <pete.grace@gmail.com>"
+license=MIT
+depends=(templatectl)
+
+image=qt:v1.2.2
+source=(https://github.com/PeterGrace/template-cartesian-graph/archive/v1.0.0.zip)
```
matteodelabre: You could use `$pkgver` here to avoid duplicating the version number.
```suggestion
source=("https://github.com/PeterGrace/template-cartesian-graph/archive/v${pkgver%-*}.zip")
```

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(template-cartesian-graph)
+pkgdesc="A Cartesian graph template for the reMarkable tablet"
+url=https://github.com/PeterGrace/template-cartesian-graph
+pkgver=1.0.0-4
+timestamp=2021-01-14T17:46Z
+section=utils
```
matteodelabre: I think it would be useful to create a section specific to template management.
```suggestion
section=templates
```

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
...
+            if [[ "$(readlink \"${pkgdir}${t_file}.${ext}\")" != "${pkgdir}${o_file}.${ext}" ]]; then
+                rm -f "$pkgdir"${t_file}.${ext}
+                ln -s "$pkgdir"${o_file}.${ext} ${t_file}.${ext}
+            fi
+        else
+            ln -s "$pkgdir"${o_file}.${ext} ${t_file}.${ext}
+        fi
+    done
+    templatectl add -n "Cartesian Graph" -f "template-cartesian-graph" \
+        -c "Custom" -c "Math"
```
matteodelabre: For clarity it would be best to use long flags here.
```suggestion
    templatectl add --name "Cartesian Graph" --filename "template-cartesian-graph" \
        --category "Custom" --category "Math"
```

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
...
+}
+
+preremove() {
+    templatectl remove -n "Cartesian Graph"
```
matteodelabre:
```suggestion
    templatectl remove --name "Cartesian Graph"
```

package/templatectl/package
```diff
@@ -0,0 +1,25 @@
...
+pkgdesc="Tool to add/remove templates for xochitl"
+url=https://github.com/PeterGrace/templatectl
+pkgver=0.1.0-1
+timestamp=2021-01-14T01:54Z
+section=utils
+maintainer="Peter Grace <pete.grace@gmail.com>"
+license=MIT
+
+image=rust:v1.2.2
+source=(https://github.com/PeterGrace/templatectl/archive/v0.1.0.zip)
```
matteodelabre:
```suggestion
source=("https://github.com/PeterGrace/templatectl/archive/v${pkgver%-*}.zip")
```

package/templatectl/package
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(templatectl)
+pkgdesc="Tool to add/remove templates for xochitl"
+url=https://github.com/PeterGrace/templatectl
+pkgver=0.1.0-1
+timestamp=2021-01-14T01:54Z
+section=utils
```
matteodelabre:
```suggestion
section=templates
```

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(template-cartesian-graph)
+pkgdesc="A Cartesian graph template for the reMarkable tablet"
```
matteodelabre: I suggest keeping the description simple. See the guidance at <https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#pkgdesc>.
```suggestion
pkgdesc="Cartesian graph template"
```

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
```
matteodelabre: Happy new year :-)
```suggestion
# Copyright (c) 2021 The Toltec Contributors
```

package/template-cartesian-graph/package
```diff
@@ -0,0 +1,49 @@
...
+    install -D -m 755 -t "$pkgdir"/opt/share/remarkable/templates \
+        "$srcdir"/template-cartesian-graph.svg
+}
+
+configure() {
+    o_file="/opt/share/remarkable/templates/template-cartesian-graph"
+    t_file="/usr/share/remarkable/templates/template-cartesian-graph"
+    for ext in png svg; do
+        if [[ -L "${pkgdir}${t_file}.${ext}" ]]; then
+            if [[ "$(readlink \"${pkgdir}${t_file}.${ext}\")" != "${pkgdir}${o_file}.${ext}" ]]; then
```
matteodelabre: A bit counterintuitive but I don’t think the inner double quotes need escaping here.
```suggestion
            if [[ "$(readlink "${pkgdir}${t_file}.${ext}")" != "${pkgdir}${o_file}.${ext}" ]]; then
```

matteodelabre (dismissed):
Thanks for your PR @PeterGrace! Here are a few comments.

package/templatectl/package
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
```
matteodelabre:
```suggestion
# Copyright (c) 2021 The Toltec Contributors
```

package/remarkable_math_graph/package
matteodelabre: @Eeems Maybe another use-case for a bind mount that we can restore in the reenable script?

Eeems: @matteodelabre I was thinking that as well.

package/template-cartesian-graph/package
Eeems: Which means we should update the doc as part of this PR.

matteodelabre: Relevant docs are in <https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#section> for reference.

Eeems:
Deferring system upgrade question until we handle #109

Eeems (commented):
@matteodelabre I think this is ready to merge, but I'd like you to double check my changes.

package/templatectl/package
```diff
@@ -0,0 +1,62 @@
...
+        "$srcdir"/target/armv7-unknown-linux-gnueabihf/release/templatectl
+    install -D -t "$pkgdir"/lib/systemd/system "$srcdir"/usr-share-remarkable-templates.mount
+}
+
+preinstall() {
+    local target_path=share/remarkable/templates
+    local factory_path="/usr/$target_path"
+    if [ ! -d /home/root/.entware/"$target_path" ]; then
+        mkdir -p /home/root/.entware/"$target_path"
+        cp -r /"$factory_path"/* /home/root/.entware/"$target_path"
```
matteodelabre: `$factory_path` already has a leading slash.
```suggestion
        cp -r "$factory_path"/* /home/root/.entware/"$target_path"
```

package/templatectl/package
```diff
@@ -0,0 +1,62 @@
...
+
+configure() {
+    systemctl daemon-reload
+    systemctl enable --now usr-share-remarkable-templates.mount
+    echo "If you perform a system upgrade that adds new templates"
+    echo "Manual intervention may be required to handle adding the new templates"
+}
+
+preremove() {
+    systemctl disable --now usr-share-remarkable-templates.mount
```
matteodelabre, line 54: @Eeems Is this different from what we discussed here: https://github.com/toltec-dev/toltec/pull/199#discussion_r556735501?

package/templatectl/package
```diff
@@ -0,0 +1,62 @@
...
+    local target_path=share/remarkable/templates
+    local factory_path="/usr/$target_path"
+    if [ ! -d /home/root/.entware/"$target_path" ]; then
+        mkdir -p /home/root/.entware/"$target_path"
+        cp -r /"$factory_path"/* /home/root/.entware/"$target_path"
+    fi
+}
+
+postupgrade() {
+    systemctl stop usr-share-remarkable-templates.mount
```
matteodelabre, line 43: Why do we need this?

scripts/package-build
```diff
@@ -228,7 +228,7 @@ SCRIPT
             for step in pre post; do
                 if [[ $(type -t "$step"upgrade) = function ]] \
                     || [[ $(type -t "$step"remove) = function ]] \
-                    || [ -d "$pkgdir"/opt/usr/share/applications ] && [ $step = post ]; then
+                    || { [ -d "$pkgdir"/opt/usr/share/applications ] && [ $step = post ]; }; then
```
matteodelabre, line 231: Well spotted. Sorry I missed that during my review.

matteodelabre (commented):
This looks good. Going to test this on my device and will report back with comments. Here are some notes on the diff:

package/templatectl/package
Eeems: Because on an upgrade the file may have changed, so we stop it in postupgrade, daemon-reload and then start in configure.

package/templatectl/package
Eeems: I'm not doing an is-active/is-enabled check beforehand and then logging what's happening here. I guess we should pick a standard that we'd prefer and document it. That or allow either.

scripts/package-build
Eeems: I had it in `()` originally, but shellcheck didn't like it, and I misunderstood how bash does it's boolean logic.

package/templatectl/package
matteodelabre: Probably factor out the logic for handling systemd services to `install-lib`.

matteodelabre: Opened #243 to track this.

package/templatectl/package
matteodelabre: Fair enough. This should also go in #243 since I don’t think we’re doing this for all packages that contain systemd services.

matteodelabre (dismissed):
Tested on rM1, works great! As might be expected, I had to restart xochitl to make the templates appear, this might confuse users as it’s not explicitly mentioned in the install output.

Eeems:
> Tested on rM1, works great! As might be expected, I had to restart xochitl to make the templates appear, this might confuse users as it’s not explicitly mentioned in the install output.

Perhaps that's something templatectl should be in charge of outputting if it detects that xochitl is runing?

package/templatectl/usr-share-remarkable-templates.mount
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Bind mount /opt/share/remarkable/templates over /usr/share/remarkable/templates
+DefaultDependencies=no
+Conflicts=umount.target
+Before=local-fs.target umount.target
```
matteodelabre, line 5: Just a nit: Why disable default dependencies and re-add them manually afterwards? If you just want to disable the `After=local-fs-pre.target` and `Before=local-fs.target` default dependencies, you can simply set the `nofail` mount option below.

Eeems: I stole this from the *.mount's in bootstrap/reenable.

matteodelabre: Okay, we’ll fix that at a later time then.

matteodelabre (approved):

PR #225, opened 2021-01-14
Workflow scripts improvements

matteodelabre:
* Send Discord notifications when new commits are added to the stable and testing branches and result in a successful CI build.
* Factor out the logic for synchronizing with the remote server to a `sync-repository` action.
* Factor out the logic for fetching build dependencies to a `setup` action.

For repository maintainers: I set up a repo at <https://github.com/toltec-dev/workflow-tests> for testing the updated `stable.yml` and `testing.yml` workflows. They deploy to <https://toltec-dev.org/workflow-tests-stable> and <https://toltec-dev.org/workflow-tests-testing>.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -10,19 +10,24 @@ jobs:
         steps:
             - name: Checkout the Git repository
               uses: actions/checkout@v2
-            - name: Install missing tools
-              run: |
-                sudo apt-get install bsdtar tree
+            - name: Setup Toltec dependencies
+              uses: ./.github/actions/setup
             - name: Build packages
               run: |
                 remote_repo="https://toltec-dev.org/testing" make repo
-            - name: Transfer packages and index
-              run: |
-                mkdir -p private
-                chmod 700 private
-                echo "${{ secrets.SSH_PRIVATE_KEY }}" > private/id_rsa
-                echo "${{ secrets.SSH_KNOWN_HOSTS }}" > private/known_hosts
-                chmod 600 private/*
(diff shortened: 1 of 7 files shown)


.github/workflows/testing.yml
```diff
@@ -10,19 +10,24 @@ jobs:
...
+                remote-path: ${{ secrets.REMOTE }}:/srv/toltec/testing
+            - name: Send notification to Discord
+              uses: ./.github/actions/discord-send
+              with:
+                channel: ${{ secrets.DISCORD_TESTING_CHANNEL_ID }}
+                token: ${{ secrets.DISCORD_TESTING_CHANNEL_TOKEN }}
+                title: New Toltec testing update available
+                link: https://toltec-dev.org/testing
+                color: 0xe3b341
+                message: ${{ github.event.head_commit.message }}
```
raisjn: what happens if the discord secrets are unset? (i have a repo which builds and pushes to my own URL and am wondering if the build will fail if the secrets are missing)

raisjn (dismissed):
looks good, nice refactoring

.github/workflows/testing.yml
matteodelabre: It does fail the build — see <https://github.com/toltec-dev/workflow-tests/runs/1708568012>. I’ll think about a workaround.

matteodelabre: Okay, adding the `continue-on-error` flag makes it ignore errors from the Discord step. (See <https://github.com/toltec-dev/workflow-tests/runs/1708690940>.)

raisjn (approved):

PR #226, opened 2021-01-14
Document new sections & move packages to new sections

### LINKED ISSUE #203: clean up utils section
the util and utils section are the largest sections and have some apps that can be re-sectioned. some ideas for sections:

* streaming
* runtimes (or libraries?)
* terminals
* input

raisjn:
add screensharing, templates, devel and admin sections


LinusCDE (dismissed):
Seems good. Not sure what issue/pr was responsible for re-categorizing packages into these sections but they should probably get merged at around the same time.

matteodelabre:
Do you think we could use this PR to actually move packages to their new sections @raisjn? So that we could close #203.

raisjn:
do the package versions need to be bumped or is it enough to just change the section in the package file? (basically: do we need new packages or is it ok just to update the web listing)

matteodelabre:
Would be best to update the revision number since this will change the `control` file inside the package, and therefore the package checksum.

raisjn:
upon reflection, i don't have available time to bump versions and change packages - i'll leave it to someone else to merge this PR or to expand upon it

matteodelabre:
Done. This will need a new review though.

LinusCDE (commented):
Should we maybe quote all sections? I've seen that that spellcheck enforced that only on "admin", but having some quoted and some not might confuse people.

matteodelabre:
Yeah, Shellcheck thought there was a possibility of a mistake since `admin` is also a command. I guess we can quote them all.

matteodelabre:
Done.

LinusCDE (dismissed):
Thanks. :shipit:

package/rmkit/package
```diff
@@ -85,7 +85,7 @@ nao() {
     pkgdesc="Nao Package Manager: opkg UI built with SAS"
     url="https://rmkit.dev/apps/nao"
     pkgver=0.1.0-4
```
Eeems:
```suggestion
    pkgver=0.1.0-5
```

package/rmkit/package
```diff
@@ -131,7 +131,7 @@ simple() {
     pkgdesc="Simple app script for writing scripted applications"
     url="https://rmkit.dev/apps/sas"
     pkgver=0.1.1-1
```
Eeems:
```suggestion
    pkgver=0.1.1-2
```

package/rm2-suspend-fix/package
```diff
@@ -7,7 +7,7 @@ pkgdesc="Fix issue where suspend breaks networking on the reMarkable 2"
 url=https://toltec-dev.org/
 pkgver=0.0.0-1
```
LinusCDE:
```suggestion
pkgver=0.0.0-2
```

LinusCDE (changes requested):
Good catch, Eeems. I hadn't checked on that. Found another one as well.

package/rm2-suspend-fix/package
Eeems: I really thought I had found them all when I was skimming through this morning haha.

LinusCDE (approved):

Eeems (approved):
lgtm

matteodelabre:
Thanks for your careful reviews. These things are hard to spot, and I’m still not sure what’s the best way to automatically detect whether a package has changes that require a version bump. Something we’ll need to figure out in #8.

PR #234, opened 2021-01-16
[rmkit] update rmkit apps

raisjn:
* rmkit: fix touch event handling while dragging
* harmony: faster drawing on pen down on rM2
* remux (0.1.7)
  * add "quick back" feature
  * add config file to configure gestures
  * make killing xochitl harder
* nao: use built in wget when available

everything in rmkit gets bumped because of the touch drag fix


LinusCDE (dismissed):
Runs on both devices.

(Btw, you may want to add an "opkg" badge for genie on your repo as well.)

Eeems (approved):

PR #241, opened 2021-01-17
Update Oxide to v2.1.2

Eeems:
Fix issue where the lockscreen doesn't work properly when you specify "No PIN".


LinusCDE:
Let me know when you want this reviewed again. Will probably not do this today anymore though.

Eeems:
@raisjn likely found the issue you were running into. I'll push out v2.1.2 later today and update this PR to use it.

LinusCDE:
I think the "suspend screen loop" issues should be solved first. I think the bug wasn't present before and I can experience it on both models (either on the launcher screen or in apps that don't have a suspend screen; didn't test xochitl yet).

Eeems:
> I think the "suspend screen loop" issues should be solved first. I think the bug wasn't present before and I can experience it on both models (either on the launcher screen or in apps that don't have a suspend screen; didn't test xochitl yet).

That's one of the things this is suppose to solve. See https://github.com/toltec-dev/toltec/pull/241#issuecomment-761877289

LinusCDE:
Oh. Missed that. Thanks for the heads up.

LinusCDE:
(Still suspend screen looping. Not sure if this was the version I already tested.)

Eeems:
While this doesn't fix the no pin situation, it does fix a different issues that users are having: https://github.com/Eeems/oxide/issues/172#issuecomment-789979351
So I think it's still a good idea to get it merged.

danshick:
I've been daily driving this commit long enough to call it stable enough for testing.

Edit: Any idea why checks are failing? Looks like maybe some kind of toolchain issue?

danshick:
Ah, yup. New Ubuntu.

danshick (approved):
LGTM

PR #244, opened 2021-01-17
Update fingerterm

### LINKED ISSUE #242: Update fingerterm to newest source?
Hi, I wasn't sure how else to do this since I couldn't find a similar request in the issues.

But  dixonary's fingerterm-reMarkable repo has had a lot of updates (toltec timestamp says 2017), namely being able to turn off the large feedback when using the keyboard by adding

```
showVisualKeyPressFeedback=false
keyPressFeedback=false
```

to `[ui]` in `/home/root/.config/FingerTerm/settings.ini`

Should I send a PR into testing with updated SHA's? Or is it best if we send in issues for the package devs/toltec maintainers to handle.

matteodelabre:
(Fixes #242)

This adds commit <https://github.com/dixonary/fingerterm-reMarkable/commit/02c17b5b485743c698e005ca89366c32b66aa044> which provides a new setting called 'showVisualKeyPressFeedback' for disabling the flash that happens after each tap on a key of the terminal keyboard.

### DIFF
--- a/package/fingerterm/package
+++ b/package/fingerterm/package
@@ -11,7 +11,7 @@ section="admin"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=GPL-2.0-or-later
 
-image=qt:v1.1
+image=qt:v1.2.2
 source=(
     https://github.com/dixonary/fingerterm-reMarkable/archive/02c17b5b485743c698e005ca89366c32b66aa044.zip
     fingerterm.png
--- a/package/fingerterm/package
+++ b/package/fingerterm/package
@@ -5,20 +5,20 @@
 pkgnames=(fingerterm)
 pkgdesc="Terminal emulator with an on-screen touch keyboard"
 url=https://github.com/dixonary/fingerterm-reMarkable
-pkgver=1.3.5-11
-timestamp=2017-12-08T15:40Z
+pkgver=1.3.5-12
+timestamp=2020-10-27T12:02Z
 section="admin"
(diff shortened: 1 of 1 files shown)


package/fingerterm/package
```diff
@@ -5,20 +5,20 @@
...
 url=https://github.com/dixonary/fingerterm-reMarkable
-pkgver=1.3.5-11
-timestamp=2017-12-08T15:40Z
+pkgver=1.3.5-12
+timestamp=2020-10-27T12:02Z
 section="admin"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=GPL-2.0-or-later
 
 image=qt:v1.1
```
LinusCDE:
```suggestion
image=qt:v1.2.2
```

Might upgrade this as well while at it. (Tested with both versions (and each on both models), so I'll can approve it no matter what version gets used.)

LinusCDE (dismissed):
Works on both rM1 and rM2. Tested with the new visual option. I wanted something like it for a while! :+1: 

I also re-tested with using qt:v1.2.2. Might throw this upgrade in there as well.

package/fingerterm/package
matteodelabre: Good point. Thanks for testing!

matteodelabre:
Thanks for your review @LinusCDE. You’ll need to re-approve since I committed your suggestion to the branch.

LinusCDE (approved):
Sure. Wasn't sure if this would dismiss it since I approved with that suggestion.

PR #247, opened 2021-01-19
Add ddvk-hacks

matteodelabre:
Patches Xochitl on install, restores original binary on remove. Tested on rM1, version 2.5.0.27.


raisjn:
awesome! this day has been a long time coming :-D

cc @ddvk

package/ddvk-hacks/package
```diff
@@ -0,0 +1,200 @@
...
+
+    echo
+    echo "Device: $device"
+    echo "Xochitl version: $xochitl_version"
+    echo "Patch version: $patch_version"
+    echo
+
+    if ! sha1sum -c <(echo "$original_hash  $_xochitl_path") > /dev/null 2>&1; then
+        echo "Invalid Xochitl checksum"
+        echo "Maybe ddvk-hacks are already installed?"
```
Eeems: Perhaps check the backup location that xochitl gets placed in when using @ddvk's script?

matteodelabre: Good point. If we detect its presence, should we just warn the user and ask them to do the cleanup? Or use the upstream script’s backup to restore Xochitl, then re-run our script as normal?

Eeems: Not quite sure. @ddvk what are you comfortable with?

LinusCDE:
Works great on my rM 2! Best tag me in discord once the details are fleshed out. I can re-review (and most likely approve) it then.

Eeems (dismissed):
Works fine for me on my rM1

Eeems:
Just waiting on input from @ddvk before approving.

package/ddvk-hacks/package
ddvk: probably it is better to let the user do the cleanup

ddvk:
> Just waiting on input from @ddvk before approving.

what happens after the device has been upgraded? 
the way i see it, the package has to be foce reinstalled

LinusCDE:
To loosely repeat some of the discussions in discord relating to this ([starting here])(https://discordapp.com/channels/385916768696139794/756853330382422037/804318381644972033)):

The package doesn't contain any files outside of `/opt` and `/home/root` so I won't get automatically reinstalled after an upgrade. This behaviour also may not be fully wished or has to be altered since right after an update, there will most likely not be a patch for the new xochitl.

There was also some discussion on how to treat and where to store the backup of the xochitl binary do we don't accidentally downgrade it.

Eeems:
So what is this waiting on?

matteodelabre:
I don’t see any remaining issue with the PR. So this is just waiting on someone to review it.

Eeems (approved):

PR #249, opened 2021-01-20
Fix link in docs/package.md

jonahweissman:
The link was to `scripts/install-lib`, but because `package.md` isn't in the root of the repo, it pointed to the wrong place. It should point to the correct file now.


matteodelabre (approved):
Good catch! Thanks for your first contribution @jonahweissman!

PR #250, opened 2021-01-20
bumping rmkit apps to jan/20/2021

raisjn:
rmkit (requires bumping all apps):
  * fix font artifacts when using large font, specifically showed up in minesweeper
  * warn when running on rM2 without rm2fb

remux:
  * use ?MB for apps we don't know mem usage of
  * idle fixes for rM2
  * properly draw 8bit grayscale suspend screens
  * fix touch flood to use ABS_DISTANCE=2 instead of 0

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -9,12 +9,12 @@ license=MIT
 
 image=python:v1.1
 source=(
-    https://github.com/rmkit-dev/rmkit/archive/0b4bda6a31a075dc3be093c338322203137549bf.zip
+    https://github.com/rmkit-dev/rmkit/archive/3d8ba83684f76b3e4146af9f295de4615a416cb8.zip
     remux.service
     genie.service
 )
 sha256sums=(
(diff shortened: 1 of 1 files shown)


LinusCDE (approved):
LGTM. Tested on both models. Not sure if it's placebo, but the power button seems a lot more responsive now on rM 2 with remux in xochitl. :+1: 

Also for some reason I already had the genie version. Not sure what happened there but this version obviously bumps it.

LinusCDE:
Note: Seems that I my device now fails to connect to WiFi. These issues continue however when I switch from remux to xochitl and restart as well. So I can't put my finger on it if remux is causing it (also have ddvk-hacks installed but it seems to work fine until now). I'm a bit confused here.



(I also had a point where the device thought it was connected but I couldn't even ping it).

Please retest this on the rM 2 as well (checking wheter wifi works and also continues to work after one suspend/wakeup).

LinusCDE:
My previous comment seems to be most likely unrelated to remux (or any launcher for that matter). Go ahead and merge it. :+1:

raisjn:
thanks for review and testing!

PR #232, opened 2021-01-16
Stable merge window for week 03 of 2020

matteodelabre:
New packages:

* decay 2.1\~beta-1

Updated packages:

* draft 0.2.0-16 → 0.2.0-17
    - Fix install script asking to disable shutdown.target
* erode 2.0.3\~beta-1 → 2.1\~beta-1
    - Add .oxide file
    - Automatically reload oxide on install
* fret 2.0.3\~beta-1 → 2.1\~beta-1
    - Add .oxide file
    - Automatically reload oxide on install
* oxide 2.0.3\~beta-1 → 2.1\~beta-1
    - Add .oxide file
    - Automatically reload oxide on install
* rot 1.0.0 → 2.1\~beta-1
    - Fix wrong version
* tarnish 2.0.3\~beta-1 → 2.1\~beta-1
    - Fix install script asking to disable shutdown.target
* koreader 2020.12-2 → 2020.12-3
    - Add .oxide file
    - Automatically reload oxide on install
    - Remove warning relative to Oxide
* plato 0.9.1-13 → 0.9.13-1
* genie 0.1.0-1 → 0.1.1-1
* simple 0.1.0-3 → 0.1.1-1
* remux 0.1.5-1 → 0.1.6-2
* toltec-bootstrap 0.0.1-2 → 0.0.1-3
    - Update wget to 1.21.1
* xochitl 0.0.0-2 → 0.0.0-3
    - Add .oxide file
    - Automatically reload oxide on install

Install:

* Update wget to 1.21.1
* Upgrade old opt.mount bind mount from remarkable_entware
* Keep branch from existing install if applicable

Documentation:

* Link to code of conduct and security policy.

Tooling:

* Add flag field to disable auto stripping of binaries
* Send Discord notifications on update

PR #252, opened 2021-01-21
Apply Oxide hotfix

Eeems:
Hotfix for issues found in #232

### DIFF
--- a/package/oxide/package
+++ b/package/oxide/package
@@ -3,13 +3,13 @@
 # SPDX-License-Identifier: MIT
 
 pkgnames=(erode fret oxide rot tarnish decay)
-pkgver=2.1~beta-3
+pkgver=2.1~beta-4
 timestamp=2021-01-07T03:28Z
 maintainer="Eeems <eeems@eeems.email>"
 license=MIT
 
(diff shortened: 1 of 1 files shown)


raisjn:
how do i see what patches went into the release zip? did you upload a new tarball manually to v2.1 release on github?

Eeems:
I replaced the zip files from the latest build on the `support/v2.1` branch. You can see the changes here: https://github.com/Eeems/oxide/compare/support/v2.1

matteodelabre:
Tested on my rM1 and this fixes the issues around the `events` field registration. Thanks!

Eeems:
Do we want to wait for a rM2 test or just merge?

raisjn:
i think if its an issue that applies to both, it should be fine to test+accept with rM1 unless there is a reason to think rM2 won't work or has a different code path

Eeems:
> 
> 
> i think if its an issue that applies to both, it should be fine to test+accept with rM1 unless there is a reason to think rM2 won't work or has a different code path

I expect them both to work the same. I'm just waiting on an approval to merge.

raisjn (approved):

PR #256, opened 2021-01-22
Fix missing prerm script in rm2fb

matteodelabre:
Due to a logic bug fixed in <https://github.com/toltec-dev/toltec/pull/222/files#diff-d7f7031a7af306332472a25649f7ebc8a7de7c7ce2eb421a30c8f76106f3b0fbR231>, prerm scripts were not generated for corresponding preremove functions.

rm2fb’s version was not increased since this fix was merged, therefore the current package (1.0.0-6) does not contain a prerm script.

This fix increments rm2fb’s version to trigger a rebuild of the package.


raisjn (approved):
for testing purposes in the future, i will likely create a dummy package that echo's the stage being executed, then install / reinstall / remove, etc and verify each stage is being echo'd properly

PR #257, opened 2021-01-22
Make bootstrap execution conditional on hash verification

danshick:
This should keep users from copy and pasting all three commands and running a bootstrap script even if the hash check fails.


matteodelabre (changes requested):
Thanks for the PR! May I also suggest to update these lines in `README.md`?

danshick:
> Thanks for the PR! May I also suggest to update these lines in `README.md`?

Done

matteodelabre (approved):
Tested on my rM1, working fine.

PR #260, opened 2021-01-25
Support auto-extracting more archive types

### LINKED ISSUE #259: Missing extraction formats in python tooling
The packaging documentation says that `.tar.xz` and `tar.bz` [are supported](https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#source), but the new python tooling [does not support these two formats](https://github.com/toltec-dev/toltec/blob/testing/scripts/toltec/util.py#L105-L129). In addition, it [does not check the return value](https://github.com/toltec-dev/toltec/blob/testing/scripts/toltec/builder.py#L194) of `auto_extract` to catch unsupported formats.

matteodelabre:
(Fixes #259.)

Add support for `.tar`, `.tar.xz` and `.tar.bz2` archives in auto-extract procedure. Print to debug log when a source file is not auto-extracted due to an unsupported archive type.


Eeems:
Should we also throw in support for zstd?

matteodelabre:
This would require introducing a new set of Python dependencies since zstd is not currently supported in the Python standard library (see https://bugs.python.org/issue37095). Unless we have some project which only distributes its sources in `tar.zst` archives, I think we should leave zstd support out for now.

LinusCDE:
Sounds great. Especially xz seems great since it can compress with multiple cores and also has one of the strongest compressions available.

Arch switched from `tar.xz` to `tar.zstd` because while zsdt a tiny bit less efficient in compression it has a 13x faster decompression rate ([blog post](https://archlinux.org/news/now-using-zstandard-instead-of-xz-for-package-compression/)). Since we don't actually ship the packages in zstd, I don't think there is a reason to peruse it.

Eeems:
> Arch switched from `tar.xz` to `tar.zstd` because while zsdt a tiny bit less efficient in compression it has a 13x faster decompression rate ([blog post](https://archlinux.org/news/now-using-zstandard-instead-of-xz-for-package-compression/)). Since we don't actually ship the packages in zstd, I don't think there is a reason to peruse it.

Right, but this isn't about shipping packages, it's about consuming sources. So if we want to use a zstd source to build the package we have to install extra software.

LinusCDE:
> Right, but this isn't about shipping packages, it's about consuming sources. So if we want to use a zstd source to build the package we have to install extra software.

Was aware but liked the post as it had some comparisons to xz. The switch might also increase popularity, but I don't think we will get any sources that require zstd support anytime soon.

matteodelabre:
Indeed, we can always add support for zstd when we need it. In the meantime, do you think this PR can be merged?

Eeems:
Code changes look good, but I haven't done any testing.

raisjn (approved):
shipit

PR #263, opened 2021-01-25
Print msgs on error

LinusCDE:
Here is the change to print the last 200 lines when a bash script fails, @matteodelabre .

Might want to refactor it a bit to fit better into the toolchain. I hope this is somewhat similar to the current coding style already.

(I didn't put it in utils.py since the function refers to `bash.ScriptError` and utils seems to be more general with not imports from the own application at all.)

### DIFF
--- a/scripts/toltec/bash.py
+++ b/scripts/toltec/bash.py
@@ -9,6 +9,7 @@ from docker.client import DockerClient
 
 AssociativeArray = Dict[str, str]
 IndexedArray = List[Optional[str]]
+LogGenerator = Generator[str, None, None]
 Any = Union[str, AssociativeArray, IndexedArray]
 Variables = Dict[str, Optional[Any]]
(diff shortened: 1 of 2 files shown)


matteodelabre:
Thanks for implementing this @LinusCDE! I’ll review it shortly. I think 200 lines might be a bit too much though, maybe something like 50 to 100 lines would be more reasonable?

LinusCDE:
Good point. I used 200 lines, since buildroot [did a similar thing](https://git.buildroot.org/buildroot/tree/.gitlab-ci.yml?h=2019.08.x#n52). Might be helpful for some compiler errors, but have to agree in most of our cases it's a bit much.

matteodelabre (approved):
I changed the context size to 50 lines. If you agree with this change, this looks good to me for merging. Thanks again!

PR #265, opened 2021-01-27
[restream][1.1] - Updated Package

danshick:
Official release notes state:
```
- low-latency streaming of the reMarkable framebuffer
- support for reMarkable generation 1 and 2
- POSIX shell on host machine supporting linux, mac and windows
- Rust executable on reMarkable
```

It should be noted that the only actual code changes that took place since the version we most recently packaged were:
```
- Disable password authentication for ssh
- Add warning message on ssh failure
```

Both of these changes took place on the computer side script, so the real advantage of this update is to start building against the tagged releases instead of individual commits, and to update the toolchain version.

### DIFF
--- a/package/restream/package
+++ b/package/restream/package
@@ -5,15 +5,15 @@
 pkgnames=(restream)
 pkgdesc="Binary framebuffer capture tool for the reStream script"
 url=https://github.com/rien/reStream
-pkgver=0.0.0-2
-timestamp=2021-01-01T13:55:28Z
+pkgver=1.0.0-1
+timestamp=2021-01-05T19:08:03Z
 section="screensharing"
 maintainer="Dan Shick <dan.shick@gmail.com>"
 license=MIT
 conflicts=(rm2fb)
-image=rust:v1.2.1
-source=(https://github.com/rien/reStream/archive/c41b87778953513cd52d9c58b3cc5ce52825700e.zip)
-sha256sums=(89da0932f17a546f194b816eb28a83f84f2484a1508d545bfce2115908248fc3)
+image=rust:v1.3
+source=(https://github.com/rien/reStream/archive/1.0.tar.gz)
+sha256sums=(07457895927c9dd3d9566d6d432dbd0b76f97dfff45dc2964adfcc3091a2ce0f)
 
 build() {
     # Fall back to system-wide config
--- a/package/restream/package
+++ b/package/restream/package
@@ -10,7 +10,6 @@ timestamp=2021-01-28T23:25:40Z
 section="screensharing"
 maintainer="Dan Shick <dan.shick@gmail.com>"
 license=MIT
-conflicts=(rm2fb)
 image=rust:v1.3
 source=("https://github.com/rien/reStream/archive/${pkgver%-*}.tar.gz")
 sha256sums=(89fa4c8adfcdfb5266e11d1f8ed4c5d8dac12a68a7ee5622cf21f833bca1704f)
--- a/package/restream/package
+++ b/package/restream/package
@@ -5,15 +5,15 @@
 pkgnames=(restream)
 pkgdesc="Binary framebuffer capture tool for the reStream script"
 url=https://github.com/rien/reStream
-pkgver=1.0.0-1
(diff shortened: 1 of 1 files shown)


danshick:
I think it is also worth mentioning here that https://github.com/rien/reStream/pull/41 has not been accepted yet, so we are still unable to remove the conflict with rm2fb.

Eeems:
> 
> 
> I think it is also worth mentioning here that [rien/reStream#41](https://github.com/rien/reStream/pull/41) has not been accepted yet, so we are still unable to remove the conflict with rm2fb.

:(

rien:
I've created a new release: https://github.com/rien/reStream/releases/tag/1.1

The rust binary `restream.rs` and the script `restream.sh` now follow the same versions to avoid confusion.

matteodelabre:
I updated the version and took the liberty to remove the duplication of the version number in the source array. I also removed the part of the configure message that asks the user to link `/home/root/restream` to `/opt/bin/restream` since restream supports the standard Toltec location if I’m not mistaken.

danshick:
> I've created a new release: https://github.com/rien/reStream/releases/tag/1.1
> 
> The rust binary `restream.rs` and the script `restream.sh` now follow the same versions to avoid confusion. 

Thanks for the new release and for including rm2fb support!

danshick:
> I updated the version and took the liberty to remove the duplication of the version number in the source array. I also removed the part of the configure message that asks the user to link `/home/root/restream` to `/opt/bin/restream` since restream supports the standard Toltec location if I’m not mistaken.

Both changes look good to me. /opt/bin was added in my PR, so that message can be safely removed. I've also removed the rm2fb conflict!

Edit:

We can also update the release notes I mentioned above to include everything in the official 1.1 release message. Should that be duplicated here for inclusion?

matteodelabre (approved):
Tested and confirmed to work on rM1 and rM2.

1ykos:
I did install `restream` from toltec-testing and linked `/root/restream` to `/opt/bin/restream` but I'm getting an error message when running `./reStream.sh`:
```bash
[...]
error: The following required arguments were not provided:B f=0/0   
    --height <height>
    --width <width>
    --bytes-per-pixel <bytes>
    --file <path>

USAGE:
    restream [OPTIONS] --height <height> --width <width> --bytes-per-pixel <bytes> --file <path>
[...]
```
I have no idea what the correct argument for `--file` would be, otherwise I could test more.

danshick:
> I have no idea what the correct argument for `--file` would be, otherwise I could test more.

You should have gotten the following message when installing.

```
This app is only the device-side half of reStream. The companion script for consuming the output of this app can be found at <https://github.com/rien/reStream>.
```

1ykos:
That's what I did. I got `./reStream.sh` from that link and ran it, which produced the error message that I showed. The version I need is probably not in master yet. To investigate further I tried running the executable on the device directly, but I don't know what to specify in the `--file` parameter, whereas the other parameters are mostly self-explanatory.

danshick:
> That's what I did. I got `./reStream.sh` from that link and ran it, which produced the error message that I showed.

Ah, misread your last comment.

Current master branch of reStream has this line in the script.

https://github.com/rien/reStream/blob/57dcd7ca271df898471662c1bca0b8230ecb4e65/reStream.sh#L200

$fb_file is going to depend on which version of the rM you have, and whether or not you are using rm2fb. It (and all of the other arguments to this binary) gets set earlier in the conditional block between lines 100 and 129.

> To investigate further I tried running the executable on the device directly, but I don't know what to specify in the `--file` parameter, whereas the other parameters are mostly self-explanatory.

I wouldn't recommend doing this. You'll just get a very overwhelming spam of ASCII/binary noise representing the contents of the framebuffer dumped into your console. It won't be human readable output even if it is working.

I think we need to see more of the output from `./reStream.sh` before and after the error message.

1ykos:
I think I was using an old version of `./reStream.sh` somehow. Another pull fixed the issue, sorry for the inconvenience and thank you all for your work! It now works with fb2 on my rm2 too.

PR #266, opened 2021-01-28
Rewrite repo-build-web in Python

matteodelabre:
This does not change the final HTML output, but only changes its generation script to use Python (as part of the Python rewrite effort from #211).

* Use the lightweight Jinja templating engine
* Add `Repo.make_listing()` method for generating the listing
* Call `make_listing` as part of the `repo_build.py` script


raisjn (approved):
lgtm

PR #267, opened 2021-01-28
Upgrade VNSee to 0.3.1

matteodelabre:
This VNSee release fixes compatibility with TigerVNC servers.


raisjn (approved):
shipit

i assume this is the fix for bit depth in the vnsee commit history

PR #268, opened 2021-01-30
Update retris to 0.6.3-1

LinusCDE:
Should now be fluid/enjoyable on the rM 2 but may have excessive ghosting. Might also be a bit fluid on the rM 1 as well.

Also now warning the user when launching the game without rm2fb.

Just check that the game launches on rM 2 and movements are fluid. Can test rM 1 as well but not needed imo.


raisjn (dismissed):
looks good, tested on rM1 and rM2. tested on rM2 without rm2fb, got a nice message. 

on the rM1, the blocks lose their shape as they fall (a Z block may look like a 2x3 rectangle instead of a Z, for example). this doesn't happen on the rM2, afaict

danshick:
Wrong place for this, but could we build an rm2fb shim package for the rm1 that just allows other packages to depend on it regardless of the device?

raisjn:
this was briefly discussed on discord 2 weeks ago, i'm in favor of it, but it is not the approach that's being taken currently

reasons i like it:

* i think it's equal or less work than tying every package to rm2fb on rm2 using a new rm2_depends field
* it wouldn't require as much tooling changes to support
* it would have no difference between rM1 and rM2 packages
* the rM2 repo would have 1 package that is a few mb instead of 200+ MB

reason it was rejected:

> "I think that this solution would be more complex to setup than the device-specific architecture approach, and it would not re-usable by other device-specific packages"

upon reading this response, i think maybe the proposal was not explained well because i think it is less complex to setup (imo).

LinusCDE:
> looks good, tested on rM1 and rM2. tested on rM2 without rm2fb, got a nice message.
> 
> on the rM1, the blocks lose their shape as they fall (a Z block may look like a 2x3 rectangle instead of a Z, for example). this doesn't happen on the rM2, afaict

Thanks for testing. Upon testing on the rM 1 again, I agree that the amount of ghosting is just ugly and unnecessary considering the [A2-ish mode](https://github.com/canselcik/libremarkable/blob/340bc9ab838562997628620db14c300bae7e422f/src/framebuffer/common.rs#L354) is almost as fast but has almost no ghosting/artefacts.

The DU/FastMono mode on the rM 2 has a lot less of this ghosting/artefacts while having more long term ghosting which I don't consider problematic. A2-ish makes the game lag a lot more there. Didn't really notice it before, but in that sense, rm2fb beats the hardware epdc of the rM 1 in that aspect (or the panel, dunno which is responsible more).

I updated retris to make playing on the rM 1 better again.

Also attached the ipk files I forgot in my releases.

raisjn (approved):

Issue #254, opened 2021-01-22
Standard install directory structure


matteodelabre:
There is currently no guidance as to what files must be installed in which directories on the device. For example, we have packages that install binaries to `/opt/lib`, there are files both in `/opt/usr/share` and `/opt/share`, some packages install files to the root that should go to `/opt` instead.

We should agree on a structure and enforce it through lint checks.

Eeems:
To simplify things we should probably start with a standard structure already defined by arch or debian and just transpose it inside /opt

matteodelabre:
Agreed. I think we should also seek compatibility with the structure of Entware packages (if they do have a consistent structure).

LinusCDE:
Also agreed. We had this problem earlier though I guess. One example are the document readers that would normally go into `/opt`. I install plato to `/opt/lib/plato` (was the best way of an opt-like directory imo) while koreader goes into `/opt/koreader` if I'm not mistaken.

Problem was that `/opt/opt` was less ideal because of chroot possibilities if not mistaken.

PR #269, opened 2021-01-30
[zshelf][0.2.0] - New Package

khanhas:
🤞


package/zshelf/package
```diff
@@ -0,0 +1,63 @@
...
+}
+
+package() {
+    install -d "$pkgdir"/opt/share/zshelf
+
+    cp -r "$srcdir"/backend "$pkgdir"/opt/share/zshelf/backend
+    cp "$srcdir"/config.json "$pkgdir"/opt/share/zshelf
+    install -D -m 755 -t "$pkgdir"/opt/share/zshelf "$srcdir"/zshelf
+
+    ln -s "$pkgdir"/opt/share/zshelf/zshelf "$pkgdir"/opt/bin
```
LinusCDE: You're including a buildtime specific path here. I suggest to remove the first `"$pkgdir"` as this will by taken over 1-to-1 to the device.

package/zshelf/package
```diff
@@ -0,0 +1,63 @@
...
+    install -D -m 644 -t "$pkgdir"/opt/etc/draft "$srcdir"/zshelf.draft
+    install -D -m 644 -t "$pkgdir"/opt/etc/draft/icons "$srcdir"/zshelf.png
+}
+
+configure() {
+    echo ""
+    echo "To download books, you have to configure domain and cookie first."
+    echo "Please follow instructions on:"
+    echo "https://github.com/khanhas/zshelf#important-configure-domain-and-cookie"
+    echo "Config file locates at:" "$pkgdir"/opt/share/zshelf/config.json
```
LinusCDE: The `"$pkgdir"` also seems not necessary here as this code will run on the device and the `$pkgdir` doesn't exist there.

LinusCDE (changes requested):
Hi, that software looks awesome!

Installing zshelf seems to throw two errors for me though:

```shell
reMarkable: ~/ opkg install zshelf
Installing zshelf (0.1.0) to root...
Downloading https://rmtestrepo.cosmos-ink.net//zshelf_0.1.0_armv7-3.2.ipk
Installing libnghttp2 (1.41.0-1) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/libnghttp2_1.41.0-1_armv7-3.2.ipk
Installing libuv (1.40.0-1) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/libuv_1.40.0-1_armv7-3.2.ipk
Installing libcares (1.17.1-1) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/libcares_1.17.1-1_armv7-3.2.ipk
Installing libatomic (8.4.0-11) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/libatomic_8.4.0-11_armv7-3.2.ipk
Installing node (v12.19.0-1a) to root...
Downloading https://bin.entware.net/armv7sf-k3.2/node_v12.19.0-1a_armv7-3.2.ipk
Configuring libatomic.
Configuring libnghttp2.
Configuring libuv.
Configuring libcares.
Configuring node.
Configuring zshelf.

To download books, you have to configure domain and cookie first.
Please follow instructions on:
https://github.com/khanhas/zshelf#important-configure-domain-and-cookie
//opt/lib/opkg/info/zshelf.postinst: line 68: pkgdir: unbound variable
Collected errors:
 * extract_archive: Cannot create symlink from ./opt/bin to '/tmp/tol1/toltec/build/package/zshelf/pkg/zshelf/opt/share/zshelf/zshelf': File exists.
 * pkg_run_script: package "zshelf" postinst script returned status 1.
 * opkg_configure: zshelf.postinst returned 1.
```

The `/tmp/tol1/toltec/...` path was where my clone of your repo was located when building the package on my pc.

package/zshelf/package
khanhas: Thank you for your review!
Which `"$pkgdir"` is that I need to remove and why only the first one? Did you mean remove all the prefixed `"$pkgdir"`?

package/zshelf/package
khanhas: I can do this 👍

package/zshelf/package
khanhas: Oh I got it now. At the `ln` command isn't it? Github showed 3 lines above it and I was confused haha.

LinusCDE: Yes. Sorry that I didn't just point that out, was in a hurry and didn't want to make a wrong/untested suggestion.

package/zshelf/package
```diff
@@ -0,0 +1,63 @@
...
+}
+
+package() {
+    install -d "$pkgdir"/opt/share/zshelf
+
+    cp -r "$srcdir"/backend "$pkgdir"/opt/share/zshelf/backend
+    cp "$srcdir"/config.json "$pkgdir"/opt/share/zshelf
+    install -D -m 755 -t "$pkgdir"/opt/share/zshelf "$srcdir"/zshelf
+
+    ln -s /opt/share/zshelf/zshelf "$pkgdir"/opt/bin
```
LinusCDE:
```suggestion
    install -d "$pkgdir"/opt/bin
    ln -s /opt/share/zshelf/zshelf "$pkgdir"/opt/bin/zshelf
```

Before `/opt/bin` was the symlink to the binary which at best is a broken package and at worst one that can break others (luckily opkg doesn't allow it to overwrite/delete the `/opt/bin` directory). The suggestion above fixes that and the package should run just fine then.

package/zshelf/package
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(zshelf)
+pkgdesc="Z-Library browser and downloader"
+url=https://github.com/khanhas/zshelf
+pkgver=0.1.1
```
LinusCDE:
```suggestion
pkgver=0.1.1-1
```

This adds the package version. It is separate from your software version in the github repo and indicates how often we packaged that particular software version. When for example some repo changes affect your package or the packaging changes (e.g. a hook/script change or what files get included) this last part gets increased (either by the toltec maintainers (!) your yourself) to indicate an update in packaging but not the actual software itself.

An example of us increasing this number due to changes of many packages metadata can be seen [here](https://github.com/toltec-dev/toltec/pull/226/files).

package/zshelf/package
```diff
@@ -0,0 +1,63 @@
...
+    cd backend
+    npm install
+}
+
+package() {
+    install -d "$pkgdir"/opt/share/zshelf
+
+    cp -r "$srcdir"/backend "$pkgdir"/opt/share/zshelf/backend
+    cp "$srcdir"/config.json "$pkgdir"/opt/share/zshelf
+    install -D -m 755 -t "$pkgdir"/opt/share/zshelf "$srcdir"/zshelf
```
LinusCDE: I don't think that putting the binaries into `/opt/share/zshelf/` is a good idea. `/opt/share` is opkg's analogy to `/usr/share` which is meant to only host applications resources and not the applications itself.

Alternatives would be `/opt/zshelf/` or `/opt/lib/zshelf`. I'm not sure what is best either.

@matteodelabre @Eeems @raisjn Any ideas/notes here?

LinusCDE (changes requested):
Took more time for a review this time. There was only one error from getting it to run (see suggestion).

Also not sure about `/opt/share/zshelf` for the binaries (mentioned that in a suggestion to get feedback from other toltec members as well).

Otherwise it's awesome! :confetti_ball: 



After the new suggestions are resolved, I'll consider this package ready for merge. :+1: 

---

On another point regarding the new python tooling:

@matteodelabre : I think the remove option in the python toolchain should note when the build directory is still owned by /root due to a faulty run. Used that option for re-building the package and the build failed even though I didn't see a complain (or I didn't notice it since it didn't abort the operation).
```shell
The build directory 'build/package/zshelf' for recipe 'zshelf' already exists.
Would you like to [c]ancel, [r]emove that directory, or [k]eep it (not recommended)? [C/r/k] 
```

Best would be to abort when `r` fails to delete anything due to insufficient permissions.

package/zshelf/package
Eeems: According to Wikipedia `/usr/share`, which is what `/opt/share` is an analog to, is for
> Architecture-independent (shared) data. 

Since this isn't shared data, and is instead libraries and binaries used to run zshelf, it would make more sense to place it in `/opt/lib` which is analogous to `/usr/lib`. From Wikipedia for `/usr/lib`:
> Libraries for the binaries in /usr/bin and /usr/sbin.

matteodelabre: This is related to #254. I agree with what @Eeems said. Additionally, configuration files like `config.json` should go in `/opt/etc`.

khanhas: That makes sense. I will move them to `/usr/lib`.
About config.json: can I still keep it in same directory as binary and just symblink it to `/opt/etc`?

LinusCDE: > That makes sense. I will move them to `/usr/lib`.
Should do (`/opt/lib`*).

>About config.json: can I still keep it in same directory as binary and just symblink it to `/opt/etc`?

Sounds good. In fact I did the same [for plato](https://github.com/toltec-dev/toltec/blob/dfa23f55aab9219a4a35d6fa2efcf488b7527e73/package/plato/package#L61):

Minor nitpick: If possible, it would be cool though to have the actual config in `/opt/etc` and symlink from the binaries director to there. Maybe move the config in the packaging for that. The symlink also works as intended if the config doesn't exist in the beginning.

khanhas: It's possible but I don't see it would make any difference though.

khanhas: Did anyway 😆

LinusCDE: Just for the case of uninstallation to have a config stick around in /etc rather than /opt/lib. But not sure how it is handled excactly and nitpicky.

LinusCDE (dismissed):
Tested on both models and works flawless!

:shipit:

khanhas:
Thank you all! 😘

LinusCDE (approved):
Works now. Tested on both models again.

The bug I mentioned to @matteodelabre before was indeed not related to tooling but the nodejs/npm from apt being outdated and randomly failing to build.

Downloading a recent version seemed to make the builds reliable for me again.

We should probably consider adding a node image to the tooling in the future.

danshick:
Maybe we want to wrap nvm or asdf or something similar?

asdf could actually be used to create a more generic toolchain image.

Eeems:
> Maybe we want to wrap nvm or asdf or something similar?

This likely is what the package should be doing instead of using the npm/node packages.

danshick:
Well, it's tricky because the node interpreter is required after build to run the app...unless it is packed with pkg or nexe. But then the binary is huge comparatively which isn't ideal on a disk constrained device.

Eeems:
> 
> 
> Well, it's tricky because the node interpreter is required after build to run the app...unless it is packed with pkg or nexe. But then the binary is huge comparatively which isn't ideal on a disk constrained device.

Right, so require the node package, but for build time use nvm to pull down a known good one to build node_modules.

danshick:
> Right, so require the node package, but for build time use nvm to pull down a known good one to build node_modules.

Ah, gotcha. Agreed.

LinusCDE:
Is there anything preventing this from getting merged? Sure the way npm is pulled is not optimal, but same would go if we were to apt install it and it seems to build and run fine for now.

Eeems:
I don't believe so.

Eeems (approved):

PR #272, opened 2021-02-03
Add back RECIPE-push in Makefile

jonahweissman:
Somewhere along the line, `RECIPE-push` got deleted. I modified the recipe for `push` to have similar behavior to what `RECIPE-push` used to do.

Would it make sense to add a `ssh root@"$(HOST)" opkg install rsync` line at the start of both `push` and `RECIPE-push`? The current error messages when `rsync` isn't present on the reMarkable are a little confusing. SSHing twice might be annoying for anyone who doesn't have a password-less key though. Maybe just a note in the `make help` documentation instead?

There's also no mention of `make push` in the `make help` documentation.


raisjn:
thank you! ooc, would `push` have the same behavior as `RECIPE-push` for most things? is there some situation where `push` is pushing more than you want?

> The current error messages when rsync isn't present on the reMarkable are a little confusing. SSHing twice might be annoying for anyone who doesn't have a password-less key though. Maybe just a note in the make help documentation instead?

can you fix the error message printed using an `||` or other mechanism?

raisjn (dismissed):

jonahweissman:
> would `push` have the same behavior as `RECIPE-push` for most things? is there some situation where `push` is pushing more than you want?

`RECIPE-push` only pushes `.ipk` files that start with the given RECIPE, rather than all of them. I have a whole bunch of `.ipk` files sitting in my `build/repo` folder, and usually I only want to push one of them. I don't feel especially strongly about this, so if other people want to get rid of `RECIPE-push` and just have `push`, that's fine with me as long as the documentation gets updated.

> > The current error messages when rsync isn't present on the reMarkable are a little confusing. SSHing twice might be annoying for anyone who doesn't have a password-less key though. Maybe just a note in the make help documentation instead?
> 
> can you fix the error message printed using an `||` or other mechanism?

Good thinking! It seems like [`rysnc` returns an exit value of 127](https://rsync.samba.org/doxygen/head/errcode_8h-source.html) when there is no `rsync` program found on the remote host. I added code to check the error message and recommend installing rsync if we see 127. Unfortunately, about half the time, the stream closes too early and I get a different error code and this message:
```
rsync --rsync-path /opt/bin/rsync \                 
      --archive --verbose --compress --delete \                                                          
      build/repo/ \                                                                                      
      root@"remarkable":~/.cache/toltec/                                                                 
sh: /opt/bin/rsync: No such file or directory 
rsync: connection unexpectedly closed (0 bytes received so far) [sender]                                 
rsync error: error in rsync protocol data stream (code 12) at io.c(235) [sender=3.1.3]                   
make: *** [Makefile:65: push] Error 12  
```

raisjn:
> RECIPE-push only pushes .ipk files that start with the given RECIPE, rather than all of them. I have a whole bunch of .ipk files sitting in my build/repo folder, and usually I only want to push one of them. I don't feel especially strongly about this, so if other people want to get rid of RECIPE-push and just have push, that's fine with me as long as the documentation gets updated.

i thought that rsync will only cause the changed .ipk files to transfer, but i guess there are situations where you still don't want that. i'm fine with adding RECIPE-push back

> Good thinking! It seems like rysnc returns an exit value of 127 when there is no rsync program found on the remote host. I added code to check the error message and recommend installing rsync if we see 127. Unfortunately, about half the time, the stream closes too early and I get a different error code and this message:

if you can't fix the error message, what will you do? revert back to previous and add note to docs?

jonahweissman:
> if you can't fix the error message, what will you do? revert back to previous and add note to docs?

I just pushed a version that reminds people to install rsync on their reMarkable if rsync fails for any reason. I think this is pretty harmless because it's not hiding any error messages, so, worst case, it's slightly annoying for people who are having a different problem with rsync. I also updated the `make help` docs to include `make push` and to note that rsync is required.

> i thought that rsync will only cause the changed .ipk files to transfer, but i guess there are situations where you still don't want that. i'm fine with adding RECIPE-push back

You're absolutely right, it's a pretty minor thing. The primary reason I'm doing this PR is just to keep the behavior aligned with the docs.

jonahweissman:
I just came up with a reason for an explicit RECIPE-push. When I'm working on a package recipe, I will iterate with a `.ipk` file without updating the version string, and I only ever want the most recent one to be on my reMarkable. However, since the last modified date on the packages is always the `timestamp`, `rsync` will not send the most recent `.ipk` over, since a file already exists with the same name and modified date.

I updated `make RECIPE-push` to ignore the timestamps, so it sends all the matching files over, every time. I'm not quite sure how to succinctly describe this difference in behavior in the `make help` text.

raisjn (dismissed):
lgtm

> However, since the last modified date on the packages is always the timestamp, rsync will not send the most recent .ipk over, since a file already exists with the same name and modified date.

rsync also uses file size, but the behavior is configurable. (--size-only, --ignore-times, etc). i'm surprised that if you are changing a file (and its size changes) that rsync isn't uploading it.

matteodelabre:
While I agree that `RECIPE-push` is useful, I’d like to point out that the solution implemented here does not work for recipes containing packages whose names do not start with the recipe name. For example, `rmkit` builds `mines`, `harmony`, `rmkit`, …, of which only `rmkit` starts with the recipe name. A more robust solution would need to parse the recipe’s declaration (ideally using the Python tooling) to find out the list of packages that it builds, and then to invoke rsync accordingly.

jonahweissman:
> I’d like to point out that the solution implemented here does not work for recipes containing packages whose names do not start with the recipe name. For example, `rmkit` builds `mines`, `harmony`, `rmkit`, …, of which only `rmkit` starts with the recipe name. A more robust solution would need to parse the recipe’s declaration (ideally using the Python tooling) to find out the list of packages that it builds, and then to invoke rsync accordingly.

@matteodelabre I modified the rsync invocation so that it iterates through the `pkgnames` attribute of the given recipe, which should solve this problem. I definitely agree that in the long run the Python tooling is a better place for this sort of functionality to live. I was intending this PR as just a quick fix, but I could re-write this in Python if you prefer.

jonahweissman:
> rsync also uses file size, but the behavior is configurable. (--size-only, --ignore-times, etc). i'm surprised that if you are changing a file (and its size changes) that rsync isn't uploading it.

I think you're right. I'm not sure what I was doing earlier, but without `--ignore-times` it sends the file if the file size changes. I think having the `--ignore-times` flag still makes sense because I could imagine a case where someone fixes a typo in an install script that doesn't impact file size.

Makefile
```diff
@@ -62,10 +68,25 @@ $(RECIPES): %:
...
-	      root@"$(HOST)":~/.cache/toltec/
+	      root@"$(HOST)":~/.cache/toltec/; then \
+		echo "Failed to push. Make sure rsync is installed on your reMarkable."; \
+	fi
+
+$(RECIPES_PUSH): SHELL:=/bin/bash
+$(RECIPES_PUSH): %:
+	source package/$(@:%-push=%)/package; \
+	if ! rsync --rsync-path /opt/bin/rsync \
+	      --archive --verbose --compress --delete --ignore-times \
```
raisjn: not sure `--delete` should be passed here, it makes more sense with repo (deletes everything not in repo). what do you think? (it was a surprise when i ran this and a bunch of files got deleted)

jonahweissman: Good point! I agree. (I just copied the rsync command from `push` without thinking about it)

raisjn (approved):

PR #276, opened 2021-02-06
[rmkit] add bufshot app, add lamp, add iago, add changelog

raisjn:


LinusCDE:
Is bufshot supposted to not recognize the rM 1?

Gray values are also seemingly inverted.

| Photo | Screenshot with modified [live](https://github.com/canselcik/libremarkable/blob/master/examples/live.rs) that outputs a png instead of jpg | Screenshot with bufshot |
| --- | --- | --- |
| 20210207_024808_HDR_ | livepng | fb

Note that "live" is an example of the libremarkable library and such uses the shim to read the framebuffer data. Since the image was also generated by another example "demo" which also uses the shim, I guess that most rgb data is exclusive to software that uses the shim. Not sure though as I'm quite confused how many pixel formats rm2fb now actually handles/remembers.

The reduced bit depth seen in bufshot is in line with what the device shows, but note that the bottom right gradient that gets inverted.

Zooming in on the text also shows that the aliasing is inverted.

raisjn:
it should work on rm1 or rm2, ~~i just checked and my rM1 now reports "remarkable 1.0" (but used to report "remarkable 1"), so~~ i'll update and re-test before taking out of draft. 

for the colors: oof. that is not good, i normally just take screenshots of grayscale content so don't see the gradient effect. this might take a while to debug.

raisjn:
running bufshot on rM1 

---

that was a tough few hours (https://github.com/rmkit-dev/rmkit/commit/994088028de955ca213f5d09eacbaeb75a76655c), but now it works on rM1 and rM2. will update this PR later.

LinusCDE:
Will look at it again tomorrow.

raisjn:
i think i prefer to merge this after #286 to make things a little easier on me. then i just need to add bufshot() section and source/hashes don't need to change

danshick:
I'll test this out tomorrow if there are no major blockers. Any known issues keeping it in "draft" status? I've been looking for a good excuse to install a packaged version of iago.

raisjn:
i previously had it in draft until the rmkit PR went in. after adding iago, i was keeping it in draft until i am happy with launching iago (and not fielding requests about multiple swipes not working, etc). i've now added a fix to genie which should let multiple launches of iago go smoothly

danshick (approved):
Everything installs and runs fine on rM2.

PR #277, opened 2021-02-07
Add bash-completion and toltec-completion packages

### LINKED ISSUE #264: Provide Bash completion of the opkg command

LinusCDE:
Since I already had a opkg completion draft in #264, I now made it into the package `toltec-complete` and added a package for the missing `bash-complete` as well. This will likely resolve #264.

There are still some questions:
 - Backporting `_opkg` [here](https://github.com/scop/bash-completion): I didn't add a license header to the file because of that
 - The functionality of `_opkg` is pretty primitive rn and should be expanded in the future (also the reason why I don't want to backport it rn)
 - :exclamation: I currently show nearly the same message in configuration() for both packages. Where should we show it, if at all?
 - Other approach to this? I personally like this way since we can easily add more completion files (starting with `_`) to package/toltec-completion to e.g. complete packages in the bootstrap command or similar user made packages in this repo.
 - I place the sourcable file for bash-completion currently into `/etc/profiles.d` and not `/opt/etc/profiles.d` to make our bootstrap take care of reinstalling it later. Maybe we should have a unified file in `/etc/profiles.d` that includes all files in the opt-counterpart automatically and manage that through our bootstrap script as well?

I tested this package on both models and it seems to work well so far. Any feedback?


LinusCDE:
(I'm a bit torn whether i should fix all nooks and crannies of `_opkg` since I mostly kept to what other `_` files in bash-completion used and I'm not well versed with arrays in bash.)

Eeems:
>
>   * I place the sourcable file for bash-completion currently into `/etc/profiles.d` and not `/opt/etc/profiles.d` to make our bootstrap take care of reinstalling it later. Maybe we should have a unified file in `/etc/profiles.d` that includes all files in the opt-counterpart automatically and manage that through our bootstrap script as well?

Probably just a file that includes `/opt/etc/profiles.d/*` would be the best option. Although I do agree that should be part of bootstrap, that way we don't have to reinstall as many packages on reenable.

LinusCDE:
> - :exclamation:  I currently show nearly the same message in configuration() for both packages. Where should we show it, if at all?

I removed the hint from "bash-completion". The hint should that way only appear for the first package that actually adds completions at all.

> > I place the sourcable file for bash-completion currently into /etc/profiles.d and not /opt/etc/profiles.d to make our bootstrap take care of reinstalling it later. Maybe we should have a unified file in /etc/profiles.d that includes all files in the opt-counterpart automatically and manage that through our bootstrap script as well?
> 
> Probably just a file that includes /opt/etc/profiles.d/* would be the best option. Although I do agree that should be part of bootstrap, that way we don't have to reinstall as many packages on reenable.

Good point, @Eeems. :+1: I'll leave it as is for now to keep the scope smaller here. We can changes this later when we add the change. `bash-completion` should be an easy package and I don't want it have hung up on such architectural consideration rn.

---

I redid the required formatting changes. Applying some suggestions but ignoring others with some reasoning. I intended to guide the completion after examples from [bash-completion](https://github.com/scop/bash-completion). Shellcheck really dislikes some of their choices though.

I think the packages should be ready now.

May current plan it to add the `bash-completion` package as is (no modifications or "patched" in completions).

`toltec-completion` currently contains completions for opkg and the wip [toltecctl](https://github.com/toltec-dev/toltec/pull/356/files#diff-c8550e36fa5f6cb5660870fd2224905c817d8b1c855efa2b5943ebce6e48e991R168). The opkg completions a still pretty basic rn, but I think we can just improve them in the future. When they are matured, we could add them to (bash-completion)[https://github.com/scop/bash-completion] and remove them from here once we update the bash-completion package. Then `toltec-completions` will only complete toltecctl and we may even put that in toltec-bootstrap and the this package all together then.

Eeems (approved):
Worked well for me on my rM1

PR #280, opened 2021-02-10
change web background color to #fcfaf8

raisjn:
#fcfaf8 is similar proportions as before but hue shifted into orange


raisjn:
Screenshot_2021-02-09_18-50-05

danshick:
Looks good to me. While we're thinking about this, maybe we want to support [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)?

Edit:
That may be best left to it's own issue/PR, as it involves many more style changes now that I'm thinking about it.

matteodelabre:
This background color does look better indeed, thanks! What about the package listing pages? We should probably think about a solution for making the home page and the listing pages share the same stylesheet.

LinusCDE:
Sounds like a good idea. That way the design could be consistent with same fonts, sizes, colours and such.

matteodelabre (approved):
Looks good to me.

PR #281, opened 2021-02-10
[quickjs][2020.11.08] - New Package

khanhas:


package/quickjs/package
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(quickjs)
+pkgdesc="A small and embeddable Javascript engine"
+url=https://bellard.org/quickjs/
+pkgver=2020.11.08-1
+section=utils
+timestamp=2020-11-08T00:00Z
```
Eeems:
```suggestion
timestamp=2020-11-08T13:44Z
```

package/quickjs/package
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(quickjs)
+pkgdesc="A small and embeddable Javascript engine"
+url=https://bellard.org/quickjs/
+pkgver=2020.11.08-1
+section=utils
```
Eeems:
```suggestion
section=devel
```

package/quickjs/package
```diff
@@ -0,0 +1,32 @@
...
+timestamp=2020-11-08T00:00Z
+maintainer="khanhas <xuankhanh963@gmail.com>"
+license=MIT
+
+image=base:v1.1
```
Eeems: Any reason you aren't using the latest?
```suggestion
image=base:v1.3.2
```

Eeems:
I wonder if the patch is even needed due to the cross compile setup in the toolchain?

khanhas:
>I wonder if the patch is even needed due to the cross compile setup in the toolchain?

It can cross compile but it only support `WIN32`, `DARWIN` or `x86_64 Linux`. `$CROSS_PREFIX` is left blank and use `gcc` by default if platform doesn't fall in any of them. So I have to patch to force it to use `arm gcc`

khanhas:
Nah, you're right. I could just export those env before running `make`

raisjn (dismissed):
🚀 🦑

package/quickjs/package
```diff
@@ -0,0 +1,32 @@
...
+)
+
+build() {
+    patch -u Makefile -i quickjs.patch
+    make qjs.arm
+}
+
+package() {
+    install -d "$pkgdir"/opt/bin
+    install -D -m 755 -t "$pkgdir"/opt/bin/qjs "$srcdir"/qjs.arm
```
raisjn: i think this is creating a directory /opt/bin/qjs, should remove the -t as i assume you mean to create `/opt/bin/qjs` (renaming qjs.arm in the process)

khanhas: Oh heck, you're right.

raisjn (approved):

PR #284, opened 2021-02-12
Add first draft for pr templates

LinusCDE:
Here are some initial templates.

GitHub seems to support them in this location. I used [Eeems' issue templates](https://github.com/Eeems/oxide/tree/master/.github/ISSUE_TEMPLATE) for the initial guidance.

For the content, I hope that the text seems inviting. Not sure on the usage of the emojis, but I think they make the text more inviting and get the attention of the user. My main concern is that users may get discouraged by a big wall of text with requirements and not create an issue at all.

If anyone has suggestions, rewording for the templates feel free to commit them to it. A feature pr would probably be nice as well, but I'm not sure what I could write there. Probably something along the lines of the bug template.


LinusCDE:
Just noticed that the bug template is probably useless as a PR. We could maybe re purpose that for issue templates at some point. Not sure if the issue should also add those.

raisjn (dismissed):
thanks for doing this!

Eeems (dismissed):

LinusCDE:
Should I still wait for the other members to approve (@matteodelabre, @dixonary)?

matteodelabre:
In package-related templates, should we also direct users to the following resources?

* Style guide: https://github.com/toltec-dev/toltec/blob/stable/docs/contributing.md#style-guide
* Expected structure of a recipe: https://github.com/toltec-dev/toltec/blob/stable/docs/package.md

The test plan skeleton that’s included at the bottom of the *Update package* template could also be added to the *New package* template IMO. I think we should also state what we expect to see in the test plan (this could either go in the template itself or in a documentation file). @raisjn sent me this resource which has some useful info: https://secure.phabricator.com/book/phabricator/article/differential_test_plans/.

LinusCDE:
> In package-related templates, should we also direct users to the following resources?
> 
>     * Style guide: https://github.com/toltec-dev/toltec/blob/stable/docs/contributing.md#style-guide
> 
>     * Expected structure of a recipe: https://github.com/toltec-dev/toltec/blob/stable/docs/package.md
> 
> 
> The test plan skeleton that’s included at the bottom of the _Update package_ template could also be added to the _New package_ template IMO. I think we should also state what we expect to see in the test plan (this could either go in the template itself or in a documentation file). @raisjn sent me this resource which has some useful info: https://secure.phabricator.com/book/phabricator/article/differential_test_plans/.

Sounds great. I'm all for ironing this out before adding it. Feel free to add the points.

Eeems:
What is this waiting on?

matteodelabre:
We haven’t applied the suggested changes yet (see my comment above). I’ll look into it as soon as possible.

Eeems:
@matteodelabre were you going to get around to this soon? Or should someone else pick it up?

matteodelabre:
Sorry, I have to admit I forgot about this PR! I’ve uploaded the templates from this PR to the following test repository: <https://github.com/matteodelabre/test-pr-template>.

If you try to create a PR, you’ll see that GitHub doesn’t ask you to select a template and leaves the description box empty (this is different from the behavior with issue templates). You have to add a query parameter to the PR creation URL to make it autofill the template, so for example <https://github.com/matteodelabre/test-pr-template/compare/matteodelabre-patch-1?expand=1&template=add_package.md> will use the `add_package` template.

matteodelabre:
On the other hand, if one adds a default PR template, it always gets inserted in subsequent PRs without needing to fiddle with the URL query parameter. I created a second repo to test this: <https://github.com/matteodelabre/test-pr-template-default/compare/matteodelabre-patch-1?expand=1>.

matteodelabre:
I’ve merged the three PR templates into a single one which handles the various cases. (See the reasons in my comments above.)

matteodelabre (approved):

PR #298, opened 2021-02-26
[rm2fb] update rm2fb with wait ioctl and no-op on rM1

raisjn:
this updates rm2fb with matteo's changes for rM1 compat, as well as adds the wait ioctl support. imo, it will need a bit of testing (locally and from outside people) before we merge it into stable.

### DIFF
--- a/package/rm2fb/package
+++ b/package/rm2fb/package
@@ -8,7 +8,7 @@ maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
 pkgdesc="Interface to the reMarkable 2 framebuffer"
 url="https://github.com/ddvk/remarkable2-framebuffer"
-pkgver=1.0.0-7
+pkgver=1.0.0-8
 section="devel"
 
 image=qt:v1.1
--- a/package/rm2fb/package
+++ b/package/rm2fb/package
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 pkgnames=(rm2fb)
-timestamp=2020-10-10T01:41+00:00
+timestamp=2021-02-21T01:41+00:00
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
 pkgdesc="Interface to the reMarkable 2 framebuffer"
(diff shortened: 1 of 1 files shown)


LinusCDE (dismissed):
Seems to work fine. I didn't specifically test the ioctl (I can do that later today).

I think this will be fine though. Just needs a version bump.

Eeems:
> 
> 
> Seems to work fine. I didn't specifically test the ioctl (I can do that later today).
> 
> I think this will be fine though. Just needs a version bump.

Could you test with Oxide to see if it resolves that issue?

danshick:
> Could you test with Oxide to see if it resolves that issue?

I can try it out too. Should I use the 2.1.2 PR? Or the current version in testing?

Eeems:
> 
> 
> > Could you test with Oxide to see if it resolves that issue?
> 
> I can try it out too. Should I use the 2.1.2 PR? Or the current version in testing?

Please test with the 2.1.2 PR.

danshick (approved):
I've been running this for a while now with no ill effect on rM2. It may contribute to fewer weird sleep screen contentions with oxide/decay, but it is hard to say as they were so intermittent before. It certainly doesn't break anything or cause any form of regression, so I'm good for testing.

Given the nature of this change, someone with an rM1 should also review to ensure the no-op behavior is working.

Eeems:
> It may contribute to fewer weird sleep screen contentions with oxide/decay, but it is hard to say as they were so intermittent before.

Some users were getting them quite often from what I understand.

danshick:
> Some users were getting them quite often from what I understand.

I definitely had times when I had to cycle it to sleep and awake a few times to get back to where I was. I can't recall it happening since this fix so I really do think it is solved (the no pin issue being different, as we've discovered). But I've also disabled sleeping in xochitl, so that could be contributing to it not being problematic. I should go back to defaults for a week to see if it remains solved.

matteodelabre (approved):
I can confirm that the package is now safe to install on rM1. Some scenarios I tested after installing the updated rm2fb package:

* Enable the `xochitl` service.
* Manually start and wrap `xochitl` with `rm2fb-client`.
* Start Oxide and an launch an app through it.
* Reboot the tablet, make sure it does not go into a bootloop and system rollback.

PR #301, opened 2021-03-02
[rm2fb] update rm2fb to work with xochitl 2.6 (v1.0.1)

raisjn:


Eeems:
I would recommend changing the source to pull from the tag as well.

raisjn:
> I would recommend changing the source to pull from the tag as well.

feel free to, otherwise will to that on the next PR

raisjn:
i'll deal with the build stuff later, i thought it would be a simple change (but need to update some other stuff inside rm2fb)

raisjn:
(looks like we need to have the python tooling print out the package files)

matteodelabre:
> (looks like we need to have the python tooling print out the package files)

Currently it does so only in verbose mode. Do you think we should print it even in normal mode? Or should we switch the CI to verbose mode? (This will include all the build logs, even if the build succeeds.)

matteodelabre:
The `rm2fb-client` and `rm2fb-server` scripts, `rm2fb` systemd service, and `rm2fb.conf` systemd configuration need to be updated for the new library version. I added two symlinks `librm2fb_{client,server}.so.1` -> `librm2fb_{client,server}.so.1.0.1` to ease future updates and changed the configurations accordingly.

raisjn:
i think it's fine to always print out when in ci. can be all logs or just files

thanks for fixing the symlinks

matteodelabre (approved):
I did not find any issue when testing this on rM2, both with versions 2.5 and 2.6. Here’s what I tested:

* That the rm2fb service starts and stays alive.
* That running `rm2fb-client /usr/bin/xochitl` starts xochitl and shows the correct address for each function replaced by the shim.
* That running another app with `rm2fb-client` works, provided the app is compiled for the right Qt version if applicable.

raisjn:
I just finished testing installation/upgrade on 2.5. I did have a small hiccup after I installed new rm2fb where screen had artifacts, but I restarted rm2fb and remux and it went away. Not sure what caused it yet (likely two server or xochitl + server?)

* installed package
* restart rm2fb
* restart remux
* click around inside xochitl
* see artifact
* restart both rm2fb/remux, artifacts go away

PR #304, opened 2021-03-06
[rmkit] patch genie to fix crash in testing

raisjn:
from discussion with matteo, there is a crash in genie on testing (see test plan) when using the example genie config. i tested this patch and verified it fixes the crash. 

test plan:

* use 3 finger tap on rM2 with example genie config (/opt/etc/genie.example.conf). watch genie output and verify that it recognizes the gesture. gesture several times, trying to trigger crash. it's useful to verify that genie in testing crashes for you and that this version does not

### DIFF
--- a/package/rmkit/2021_03_05_genie.diff
+++ b/package/rmkit/2021_03_05_genie.diff
@@ -1,0 +1,29 @@
+diff --git a/src/genie/gesture_parser.cpy b/src/genie/gesture_parser.cpy
+index fa82b67..c94fd59 100644
+--- a/src/genie/gesture_parser.cpy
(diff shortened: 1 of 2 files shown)


danshick:
I'll update tomorrow and check. Should any repeated gesture for any command cause a crash or only the default config?

raisjn:
any gesture for any command (it's a timing thing), but its easy to use the 3 finger tap rapidly

matteodelabre (approved):
Tested on rM1 and rM2, works well. Looks good to me!

danshick:
I can't create a crash with this patch.

PR #310, opened 2021-03-11
Implement architecture separation

### LINKED ISSUE #129: feature: prevent installation of rM2 incompatible apps on rM2

### LINKED ISSUE #248: rm2fb: disable LD_PRELOAD for busybox services
from https://github.com/ddvk/remarkable2-framebuffer/issues/35#issuecomment-762977355, we need to disable the LD_PRELOAD in /lib/systemd/system/busybox-udhcpd@.service and /lib/systemd/system/busybox-ifplugd@.service (and probably more).

one way could be to add the `Environment=LD_PRELOAD=` line to the systemd services we think should have rm2fb client shim skipped.

matteodelabre:
This PR improves Toltec’s support for reMarkable 2 in a few ways:

* Prevents users from installing packages that are known to be incompatible with their device (fixes #129).
* Automatically installs and sets up the rm2fb server when an app that uses the framebuffer is installed.
* Stops relying on LD_PRELOAD to load the rm2fb shim into apps that need it. Instead, change their dynamic dependencies to include the shim directly. Thanks to this change, we no longer need to set LD_PRELOAD globally (fixes #248) and users can launch apps that need the shim without prefixing them with `rm2fb-client`.

It replaces the previous PR #136.

#### Tooling changes

More details in the updated [docs/package.md](https://github.com/toltec-dev/toltec/blob/tooling/split-architectures/docs/package.md).

* Added the `archs` field for declaring which device(s) a recipe is compatible with. Supported values are `rmall` (for device-independent packages), `rm1` (for reMarkable 1) and `rm2` (for reMarkable 2). When multiple arch values are set for a recipe, the recipe is generated once per entry, which allows specializing builds for specific devices.
* Replaced the `depends` field with an `installdepends` field used for declaring install-time dependencies. Previously, dependencies declared in the `depends` field were fulfilled both at build time and at install time, with no way of declaring install-time-only dependencies.
* Added a `replaces` field that sets the `Replaces` value in the generated package. Used for migrating users from the `rm2fb` package to the `display` and `rm2fb-client` packages (see below).
* Added a `patch_rm2fb` flag for recipes to instruct the build system to patch all the generated binaries that use the rM1 framebuffer with the rm2fb client shim.
* Changed [the web package listing](https://toltec-dev.org/dev/split-architectures) to include download links for each available architecture of each package.
* Packages are now stored in subdirectories named after their architecture. This is because opkg does not like seeing packages for architectures other than the ones it is configured to accept in opkg.conf. With this layout change, we can make opkg happy by only adding subdirectories for enabled architectures.

#### Install changes

* The bootstrap script now adds the `rmall` architecture (and folder) on all installs, and the `rm1` or `rm2` architecture (and folder) depending on the type of the device it’s executed on. It will fail if the device is not of a recognized type (based on the value of `/sys/devices/soc0/machine`).
* Running the new bootstrap script on an existing install will upgrade the configuration to use the new architecture subdirectories layout.

#### rm2fb changes

* The package is split into two packages, `display` (which provides the rm2fb server on rM2 and is an empty package on rM1) and `rm2fb-client` (which provides the client shim on rM2 and a mock shim on rM1).
* The `rm2fb-server` command is removed (the most common use-case being that the display server is started from the systemd unit).

#### Apps changes

Apps marked as reMarkable-1-only:

* vnsee (https://github.com/matteodelabre/vnsee/issues/15)
* whiteboard-hypercard (https://github.com/fenollp/reMarkable-tools/issues/12)

Apps marked as reMarkable-2-only:

* rm2-suspend-fix

Apps updated to include the rM2 shim:

* appmarkable
* calculator
* chessmarkable
* draft
* fingerterm
* keywriter
* koreader (through setting LD_PRELOAD in the existing wrapper script)
* oxide
* plato
* recrossable
* remarkable-splash
* retris
* rmkit
* wikipedia
* xochitl (through a new wrapper script at /opt/bin/xochitl)
* zshelf

#### Test plan

On rM1 and rM2 (system version: 2.5).

* Starting from a clean install of Toltec stable or testing, download the updated bootstrap script and run:

```console
$ toltec_branch=dev/split-architectures bash bootstrap
```

* This should update the `/opt/etc/opkg.conf` configuration file to enable the `rmall` architecture and, depending on the device you’re running this on, the `rm1` or the `rm2` architecture.
* It should also replace the `src/gz toltec` line with two source lines, `src/gz toltec-rmall` and `src/gz toltec-rm1` or `src/gz toltec-rm2`, pointing to subfolders of `https://toltec-dev.org/dev/split-architectures`.
* Run `opkg upgrade`. Most installed packages should get upgraded, and if you had installed any packages that use the screen, the `display` and `rm2fb-client` packages should get installed. If you previously installed the `rm2fb` package, it should get removed automatically.
* Try installing `rm2-suspend-fix` on a reMarkable 1, or `vnsee` on a reMarkable 2: both operations should fail.
* Test all updated packages to make sure they can properly draw to the screen (both on rM1 and rM2) and that you can interact with them through the touchscreen and the pen, when applicable. It is no longer needed to prefix each app invocation with `rm2fb-client` on the command line on reMarkable 2.
* In particular, make sure that apps work properly when launched from draft, oxide or remux (things to check: launching, pausing, killing apps, switching back and forth from xochitl).

#### Rollout plan

When this PR hits testing and stable, existing users will no longer receive updates until they run the new bootstrap script on their existing installs. They will also probably see an error when they do `opkg update`. We should decide whether we accept this or whether we want to make the transition easier, for example by leveraging the `toltec-bootstrap` package.


matteodelabre:
I made a few design choices on this PR that I realize may not suit everyone. Comments on the design or the implementation are very welcome. Some issues remain with the Oxide and remux launchers:

* ~~On rM2, the display glitches when launching apps from Oxide, even though it’s using the rm2fb shim. I haven’t found the source of this issue yet.~~ Fixed — see below.
* ~~remux does not use the xochitl wrapper script since it’s using the path from `/lib/systemd/system/xochitl.service`. A solution to this is being discussed.~~ Fixed in the latest rmkit release.

Eeems:
>     * On rM2, the display glitches when launching apps from Oxide, even though it’s using the rm2fb shim. I haven’t found the source of this issue yet.

Have you changed anything with the Oxide package?

matteodelabre:
Yes. I changed the package to build from source and to patch binaries that I believe use the framebuffer to add a dynamic dependency on `librm2fb_client.so`.

Eeems:
> Yes. I changed the package to build from source and to patch binaries that I believe use the framebuffer to add a dynamic dependency on `librm2fb_client.so`.

Right, we stopped building from source a while back due to issues with the current toltec toolchain. I expect that might be the issue still.

As for which binaries use the framebuffer, all but fret and rot at this point I believe.

matteodelabre:
> Right, we stopped building from source a while back due to issues with the current toltec toolchain. I expect that might be the issue still.

Indeed. I’ll check whether using the pre-built binaries fixes the issues I’m seeing.

> As for which binaries use the framebuffer, all but fret and rot at this point I believe.

So `tarnish` would need to have the shim too, which it doesn’t in the current state of the PR. That might also be the cause of the issue.

Eeems:
> So `tarnish` would need to have the shim too, which it doesn’t in the current state of the PR. That might also be the cause of the issue.

This like is why the screen messes up. When you launch an application, the current screen is saved. Resuming an application will redraw that saved screen.

matteodelabre:
> This like is why the screen messes up. When you launch an application, the current screen is saved. Resuming an application will redraw that saved screen.

That was it! Thanks for the insights.

danshick:
Quick question, for packages like reStream that automatically support rM1, rM2 stock, and rM2 with rm2fb, would there be a way to write an rM2 recipe that only optionally depends on rm2fb? Or would we just leave the `display` dependency off and if it happens to be installed, make use of it anyway?

Eeems:
I think that depends on how they've implemented their screen support.

danshick:
One of three ways. rM1 reads directly from `/dev/fb0`. rM2 without rm2fb finds the xochitl process and directly hooks its memory. rM2 with rm2fb uses `/dev/shm/swtfb.01`.

raisjn:
this looks great! i've read through all the recipes and am making my way through the build scripts. in the package recipes, it strikes me as repetitive that everyone needs to know to use patchelf in their package, is there a way to simplify it / reduce the burden for package writers?

matteodelabre:
> One of three ways. rM1 reads directly from `/dev/fb0`. rM2 without rm2fb finds the xochitl process and directly hooks its memory. rM2 with rm2fb uses `/dev/shm/swtfb.01`.

Unless there is a specific advantage in using reStream with rm2fb, I’d say that we should leave out the dependency.

scripts/toltec/builder.py
```diff
@@ -334,6 +385,8 @@ def _package(self, package: Package, src_dir: str, pkg_dir: str) -> None:
         logs = bash.run_script(
             script=package.functions["package"],
             variables={
+                **package.parent.parent.variables,
```
raisjn: nit: this is a "code smell" in that there is some accessing of a parent's parent data. it now couples the layout of the hierarchy with the usage. term for this is [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter)

matteodelabre:
> in the package recipes, it strikes me as repetitive that everyone needs to know to use patchelf in their package, is there a way to simplify it / reduce the burden for package writers?

Good point. Unfortunately there isn’t a mechanism in place to load a library of pre-defined Bash functions into recipes. I’m a bit reluctant to add such a mechanism in this PR since it’s already quite complex. Do you think it’s worth implementing this right now?

If we had a function that automated this, it would need to do three things:

1. Add the `build:patchelf` dependency to the `makedepends` array
2. Set the `nostrip` option
3. Run `patchelf` and `strip` with appropriate arguments

Steps 1. and 2. need to modify variables at the recipe scope and would need to be called outside of the build function. Step 3. needs to access the build artifacts and would need to be called inside the build function. Therefore we would probably need to define two helper functions, one to do 1. and 2., and one to do 3.

raisjn (commented):
overall the source looks good. the complexity is definitely on my mind

raisjn:
> Good point. Unfortunately there isn’t a mechanism in place to load a library of pre-defined Bash functions into recipes. I’m a bit reluctant to add such a mechanism in this PR since it’s already quite complex. Do you think it’s worth implementing this right now?

my thought here is that we know people copy code. there will be people copying this code and not copying it and we will both have to go fix it later and constantly explain this patchelf stuff to people contributing a package (who forgot to copy the code or apply it correctly). past experiences have shown me this is the case and so i prefer not to create a messy situation if we know we will have to clean it up anyways. 

my hope is that we figure out if a binary uses the framebuffer (check for fb0? check for some symbols?) and automatically do all the strip+patchelf stuff. every package will implicitly rely on patchelf.

matteodelabre:
> my hope is that we figure out if a binary uses the framebuffer (check for fb0? check for some symbols?) and automatically do all the strip+patchelf stuff. every package will implicitly rely on patchelf.

I’m all for automating this as much as possible. Deciding whether a binary needs the shim is tricky though: it’s not only a matter of checking whether it contains a call to `open("/dev/fb0")`, because it may open the framebuffer when run on rM1 and talk to rm2fb on rM2 (such apps will not need the shim, see for example reStream and VNSee).

scripts/toltec/builder.py
matteodelabre: Thanks for the review! I improved this by making the children’s `variables` dictionaries extend the parent dictionaries. I also removed the top level `GenericRecipe.variables` field, which was a leftover from a previous code iteration.

raisjn:
> Deciding whether a binary needs the shim is tricky though: it’s not only a matter of checking whether it contains a call to open("/dev/fb0"), because it may open the framebuffer when run on rM1 and talk to rm2fb on rM2 (such apps will not need the shim, see for example reStream and VNSee).

let's automate the majority case and we'll deal with the exceptions by disabling automation? better this way, i think, than make everyone deal with the burden

matteodelabre:
> let's automate the majority case and we'll deal with the exceptions by disabling automation? better this way, i think, than make everyone deal with the burden

matteodelabre:
Sorry—wrong button

matteodelabre:
> let's automate the majority case and we'll deal with the exceptions by disabling automation? better this way, i think, than make everyone deal with the burden

I improved on this by adding a new flag for recipes called `patch_rm2fb`. For recipes that have this flag enabled, all the dynamically-linked generated binaries that contain a reference to `/dev/fb0` (in the read-only data segment) will be patched to add a dynamic dependency on `librm2fb_client.so.1` automatically. An install-time dependency on the `rm2fb-shim` package will also be added to all the generated packages.

matteodelabre:
remux support is now fixed thanks to the latest rmkit update! However I’m seeing a duplicate Xochitl entry in the launcher (both entries are functional, and starting one shows the other as started), any idea why this might happen @raisjn? I’ll continue investigating.

duplicate-xochitl

raisjn:
> I improved on this by adding a new flag for recipes called patch_rm2fb

awesome! 

> The rm2fb-client command is renamed to rm2fb-shim, with the expectation that users will not need it anymore, unless for development purposes.

can we do this in a phased way (maybe first phase, we give a deprecation notice and then in a month or two we remove)? we don't know if we will break someone by deleting it out from them

matteodelabre:
> can we do this in a phased way (maybe first phase, we give a deprecation notice and then in a month or two we remove)? we don't know if we will break someone by deleting it out from them

Good point. The renaming is not that important anyway, so I just reverted it.

matteodelabre:
> remux support is now fixed thanks to the latest rmkit update! However I’m seeing a duplicate Xochitl entry in the launcher (both entries are functional, and starting one shows the other as started), any idea why this might happen @raisjn? I’ll continue investigating.

Fixed this by adding a patch that makes remux ignore the Xochitl draft file. Thanks for the help @raisjn!

sorenisanerd:
If rm1 and rm2 need different toolchains, what will rmall be built with?

matteodelabre:
rmall is meant for packages that can run unmodified on both the reMarkable 1 and the reMarkable 2. Currently, the major known difference between the two devices is in how the display is used—but the same toolchain can still be used to target both devices. Therefore we can put almost all of the packages in rmall. If at some point in the future two different toolchains are needed to build binaries, we will move packages that contain binaries to rm1 and rm2, and rmall will only contain packages with device-independent artifacts (resources, scripts, …).

raisjn (dismissed):
shipit

README.md
```diff
@@ -3,7 +3,7 @@
 ![Status of the stable repository](https://github.com/toltec-dev/toltec/workflows/stable/badge.svg)
 ![Status of the testing repository](https://github.com/toltec-dev/toltec/workflows/testing/badge.svg)
 [![rm1](https://img.shields.io/badge/rM1-supported-green)](https://remarkable.com/store/remarkable)
-[![rm2](https://img.shields.io/badge/rM2-experimental-yellow)](https://remarkable.com/store/remarkable-2)
+[![rm2](https://img.shields.io/badge/rM2-supported-green)](https://remarkable.com/store/remarkable-2)
```
raisjn: not sure this is green yet

matteodelabre: What would it take to be able to say that we fully support rM2, in your opinion?

raisjn: all rm2 apps working, bootstrap working, upgrades working (retaining rm2fb shim when toltec re enables) and some community testing period (week or two of it being in stable and people using it?) 

if this is all set, i think it can be green

raisjn: also, would prefer that we take care of oxide and draft service files to launch xochitl if their binaries are unavailable as a failsafe (so one more piece is that soft bricking should be difficult)

raisjn: and uninstaller (if that's not already there), but i put that at p1

matteodelabre: rM2 apps, bootstrap and upgrades should be working already (if we disregard 2.6 issues, but that is a separate problem).

Having a testing period seems fair.

Launcher failsafe & uninstaller are not available for the rM1 either: should we revert its status back to experimental?

raisjn: unsure, I'm in favor of fixing those, so probably just fix instead of changing badge

LinusCDE: Just happened to see this convo. I agree with @raisjn's concerns but agree with @matteodelabre's latest comment:

> Having a testing period seems fair.
> 
> Launcher failsafe & uninstaller are not available for the rM1 either: should we revert its status back to experimental?

Requiring features like an uninstaller seems like holding the rM2 badge to a higher standard than the rM1. These features affect toltec in general and not rM2 compatibility.

I can understand the concern about softbricking through failed rm2fb upgrades/re-enables though. This fix would have a higher likelyhood to trigger a soft-brick on rM2 and seems valid for me in that case.

I also think that any bigger update is firstly a stability risk until tested for at least a week by testers. So even with this change being merged into testing, I would still like a more cautious description for compatibility at first. But since this architecture change could also affect the rM 1 compatibility, it shouldn't affect the rM 2 badge status.

I would suggest:
- Ensuring the fallbacks to xochitl in service files
- Ensuring re-enables work either nice with rm2fb or at least trigger the xochitl-failsafe in service files.
- Changing the rM 2 bade to "beta" or similar (I would even accept dark green in this case).
- After a week in either testing or better stable we could update the state in either a merge release or a separate PR.

---

P.S.: I haven't looked much into the changes yet, but overlooking the diff seems promising. Maybe I can give it a try this weekend. Props for that, matteo! :+1:

matteodelabre:
Thanks @raisjn & @LinusCDE for your valuable input on the support status. I agree these are all things we should double check before announcing rM2 support. Since this is already a large PR with a fair amount of discussion, I simply reverted the badge change, and we shall discuss this further in another thread.

raisjn (dismissed):

matteodelabre:
I re-did some tests and there seems to be a bug with remux and KOReader. When I tap on KOReader in the list of apps, two instances are launched (can be seen in [h]top). KOReader in itself works fine but does not appear as launched in the remux menu. Tapping again on KOReader in remux creates even more instances of the app.

This bug does not exist in testing, so this is a regression from this PR.

package/koreader/koreader
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-# Copyright (c) 2020 The Toltec Contributors
+# Copyright (c) 2021 The Toltec Contributors
 # SPDX-License-Identifier: MIT
-
-KO_DONT_GRAB_INPUT=1 bash /opt/koreader/koreader.sh
+LD_PRELOAD=/opt/lib/librm2fb_client.so.1 KO_DONT_GRAB_INPUT=1 exec bash /opt/koreader/koreader.sh
```
raisjn: this `exec` is confusing remux in regards to koreader. I think it replaces the current process and now remux can't find /opt/bin/koreader - normally remux uses the launching binary to find if the process exists.

matteodelabre: Thanks for investigating that! Yes, exec replaces the parent process. It makes sense to use it here because we don’t need the parent Bash process to linger after it has launched KOReader. I think we can use the (Bash-specific) `-a` flag to change the name of the replacing process and thereby avoid confusing remux. I’ll try it out.

matteodelabre:
This change seems to fix the issue. Thanks for your help @raisjn.

raisjn (approved):

PR #308, opened 2021-03-10
Update appmarkable to 0.0.0-9 and rmservewacominput to 0.3.0-1

LinusCDE:
This changes the pattern to quite appmarkable and improves rmservewacominput to pick the correct event file up.

A test case could be:
 - Install both packages (rmservewacominput depends on and used appmarkable)
 - Run [this script](https://github.com/LinusCDE/rmWacomToMouse/blob/master/pcWacomToMouse.py) on your PC (probably set ONLY_DEBUG=True to only get output)

PR #312, opened 2021-03-12
Update recrossable

matteodelabre:
This PR updates the recrossable package to include the following fixes:

* <https://github.com/sandsmark/recrossable/pull/10>
* <https://github.com/sandsmark/recrossable/pull/11>

Test plan:

* On either rM1 or rM2, check that the numbering now starts at 1 in the upper left corner.
* On rM2, the numbers in each box of the grid should have more contrast than before.


matteodelabre:
Added a change in the `repo_build.py` script to generate the `Packages{,.gz}` index file after the existing packages have been fetched from the remote repository. Otherwise the repo build fails when the first built package depends on other Toltec packages (as is the case here), since the index is only created after each package is built. This is probably the issue that @LinusCDE encountered yesterday in #308.

danshick (approved):
Installs and runs fine on rM2.

PR #316, opened 2021-03-14
Pin the Ubuntu version used in workflows to 20.04

matteodelabre:
GitHub recently transitioned the `ubuntu-latest` label from Ubuntu 18.04 to Ubuntu 20.04 (see <https://github.com/actions/virtual-environments/issues/1816>). This caused our workflows to fail because the bsdtar binary was moved to a different package in 20.04.

To prevent this kind of failure in the future, we should specify a fixed Ubuntu version in our workflows. This PR implements that change.


Eeems (approved):
lgtm

PR #318, opened 2021-03-22
[gocryptfs] Add gocryptfs package (build without openssl)

plan5:


package/gocryptfs/package
```diff
@@ -0,0 +1,43 @@
...
+source=(
+    https://github.com/rfjakob/gocryptfs/archive/$_srcver.zip
+)
+
+sha256sums=(
+    3d66368cfc79a300de5a22fe01788d0c702f9107731db63fcd8850157d105cc0
+)
+
+prepare() {
+    # Official build script runs compiled binary to show version.
```
Eeems:
```suggestion
    # Official build script runs compiled binary to show version.
```

package/gocryptfs/package
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(gocryptfs)
+pkgdesc="An encrypted overlay filesystem written in Go."
+url="https://nuetzlich.net/gocryptfs/"
+_srcver=v2.0-beta2
+pkgver=$_srcver-1
```
Eeems:
```suggestion
_srcver=2.0-beta2
pkgver=$_srcver-1
```

package/gocryptfs/package
```diff
@@ -0,0 +1,43 @@
...
+license=MIT
+
+depends=(fuse-utils)
+
+makedepends=(build:golang build:git)
+
+image=base:v1.4
+
+source=(
+    https://github.com/rfjakob/gocryptfs/archive/$_srcver.zip
```
Eeems:
```suggestion
    https://github.com/rfjakob/gocryptfs/archive/v$_srcver.zip
```

package/gocryptfs/package
plan5: Can't merge because I already removed the trailing space on my branch before seeing this.
So it should be taken care of :)

package/gocryptfs/package
```diff
@@ -0,0 +1,43 @@
...
+
+sha256sums=(
+    3d66368cfc79a300de5a22fe01788d0c702f9107731db63fcd8850157d105cc0
+)
+
+prepare() {
+    # Official build script runs compiled binary to show version.
+    # This will cause the build to fail if built on another architecture
+    # So the line will be removed here.
+    sed -i "s/^.\/gocryptfs.*//" $srcdir/build.bash
```
Eeems:
```suggestion
    sed -i "s/^.\/gocryptfs.*//" "$srcdir"/build.bash
```

Eeems (changes requested):

matteodelabre:
I would suggest using the recently added [`golang:v1.5` toolchain image](https://github.com/toltec-dev/toolchain#images) and removing the `build:golang` dependency to speed up the build process.

plan5:
> I would suggest using the recently added [`golang:v1.5` toolchain image](https://github.com/toltec-dev/toolchain#images) and removing the `build:golang` dependency to speed up the build process.

Ok, good to know that's there.  I've update the PR (hope I'm doing this right...).

Btw. should I rather try to merge a branch or should I merge directly into testing as I did now?

plan5:
I've now also put all $_srcver in quotes to comply with the style guide.

package/gocryptfs/package
```diff
@@ -0,0 +1,43 @@
...
+_srcver=2.0-beta2
+pkgver="$_srcver"-1
+timestamp=2021-03-22
+section=utils
+maintainer="plan5 <30434574+plan5@users.noreply.github.com>"
+license=MIT
+
+depends=(fuse-utils)
+
+makedepends=(build:git)
```
Eeems, line 17:
```suggestion
```

Since you are using the golang image now.

Eeems: I'm guessing the build uses git?

plan5: I'm checking if it's still necessary with the golang image

plan5: It's not. I've removed it.

Eeems:
@plan5 what would be the best way to test this?

plan5:
> @plan5 what would be the best way to test this?

Are you asking for a manual or an automatized solution? 
Can you send me an example test for another package if for automized testing?

For manual testing:
To test the binary, you can just run it and see if it displays the help message.

To test functionality you can set up any directory to be encrypted. e.g.
```
mkdir /tmp/cipher /tmp/plain
echo "testpass"|gocryptfs -init /tmp/cipher
echo "testpass"|gocryptfs /tmp/cipher /tmp/plain
echo "content" > /tmp/plain/secretfile
fusermount -u /tmp/plain
```

/tmp/plain should now be empty and /tmp/cipher should contain an encrypted file.

```
echo "testpass"|gocryptfs /tmp/cipher /tmp/plain
cat /tmp/plain/secretfile
```
This should output "content".

Eeems:
> Are you asking for a manual or an automatized solution?

Manual testing. 

Thanks, I'll try to take a look later today

Eeems:
Looks like we are missing something? (Testing on a rM1)
image

plan5:
> Looks like we are missing something? (Testing on a rM1)
> image

The gocryptfs -init command only takes one argument (the to be encrypted dir):
`gocryptfs -init /tmp/cipher`

plan5:
But also fuse isn't working, hmm. I'll have to check on rM2 later.

Eeems:
> 
> 
> But also fuse isn't working, hmm. I'll have to check on rM2 later.

Looks like the module isn't included on the rM1. We may need to package that as well. See the wireguard package for information on howw to do that.

plan5:
Ok I'll have a look at it

plan5:
I've added a PR #331 to add a fuse module for rM1 and I've added the module as a dependency for rM1.

plan5:
@Eeems this still shows a requested change from 14 days ago that I've addressed in an independent commit.

danshick:
You can resolve the requested change if it has been addressed. Let's merge the fuse module before this lands.

plan5:
So I had accidentally deleted the gocryptfs branch on github and I've reuploaded it from local. It seems that nothing has broken but I thought I'd better mention this. 
I'm recreating the plan5-fusemodule PR because it was based on plan5-gocryptfs branch instead of testing branch. I will mention this PR there.

Edit: I've resolved it with the existing PR instead.

package/gocryptfs/package
```diff
@@ -0,0 +1,42 @@
...
+url="https://nuetzlich.net/gocryptfs/"
+_srcver=2.0-beta2
+pkgver="$_srcver"-1
+timestamp=2021-03-22
+section=utils
+maintainer="plan5 <30434574+plan5@users.noreply.github.com>"
+license=MIT
+
+makedepends=()
+installdepends=(fuse-utils)
```
Eeems: This will need to be updated to require the fuse module on rM1. @matteodelabre do we have a mechanism to have different install dependencies between reMarkable versions?

Eeems (changes requested):

package/gocryptfs/package
danshick: We could depend on a metapackage like we do for `display`.

Eeems: @danshick could you add information to the fuse PR with how to accomplish that for @plan5?

matteodelabre: To make the dependency rM1-only:

```bash
archs=(rm1 rm2)
# …
installdepends_rm1=(fuse-module)
```

Note that this will generate a separate package for each of the two architectures.

Eeems: @matteodelabre Which is better, two copies of the package or a metapackage?

matteodelabre: To be precise, the `display` solution is not technically using a metapackage: the `display` package exists, is empty on rM1, and contains rm2fb on rM2. You could do that also for the fuse module (i.e., make it contain the module on rM1 and be empty on rM2). I think that would preferable since it would save space on our server compared to duplicating the gocryptfs package (and that’s about the only meaningful difference between the two approaches).

danshick: You're right. Was just looking at it now. I think that's the right approach.

danshick: Added a comment over in the fuse package pointing to the relevant parts of `display`.

plan5: I've removed the fuse archs section entirely and added a logic to the package section to only install files on rm1. 
The install dependency in gocryptfs is now only pulled for rm1 via installdepends_rm1.
Does this resolve it all?

danshick: You can actually now make the fuse package a dependency for all architectures. Because the package is empty for rm2, it'll "install" it, but the package is basically a noop.

This means that there's only one gocryptfs package instead of two, which is desirable because the fuse package is actually the one that needs to vary by device.

Edit:

...and sorry for the complicated advice. We are just now introducing these split architecture packages into testing and still deciding how packages should be built in situations like this one.

plan5: Ok, I've moved it to the general installdepends.

btw I will add the gocryptfs simple-script GUI as a separate package soon.
I'll have to first find an elegant way to launch it before launchers/xochitl via systemd.
I've already asked about this in the Discord/Matrix-Room but if you have a suggestion I'm all ears :)

Eeems: I think systemd allows units to specify that they need to launch before something else does.

Eeems:
I'm waiting for the fuse PR to merge before approving this. I'd also like for someone with a rM2 to test the install/uninstall flow with the new fuse dependency.

plan5:
I've just locally merged fuse and then gocryptfs into testing and noticed that #318 conflicts with #331 because I've deleted an accidental copy of the gocryptfs recipe that ended up in the fuse branch. ("CONFLICT (modify/delete): package/gocryptfs/package deleted in HEAD and modified in gocryptfs."). 

Should I just leave it at that and you will fix it when the merge happens or can I do something to smooth it out?

Eeems:
We'll have to resolve the merge conflict after one of them gets merged into testing.

Eeems (dismissed):
Works for me with the latest build of fuse on my rM1. It's probably best to merge this after that PR though.

Eeems:
@plan5 you seem to have removed all changes to this PR?

plan5:
> @plan5 you seem to have removed all changes to this PR?

Oof, I just clicked the button to bring the branch up to date with testing. I thought this would omit the changes to the branch. I'll check out what happened tomorrow and revert it m)

plan5:
> @plan5 you seem to have removed all changes to this PR?

Ok I think it's fixed. It was a mixup because I kept the gocryptfs package in my local testing instead of its own branch. Once it's all merged I'll re-fork and keep things tidy ( I hope ;) )

Eeems:
Looks like you accidentally undid changes on the testing branch.

plan5:
> Looks like you accidentally undid changes on the testing branch.

Ah yes that makes sense, with the revert.

I've merged the official testing branch again. I hope it's fixed now.

Eeems:
It is not, using a revert commit will not allow merges to fix the issue. You'll have to go in and manually revert the changes or do some history rewriting to fix it up.

plan5:
> It is not, using a revert commit will not allow merges to fix the issue. You'll have to go in and manually revert the changes or do some history rewriting to fix it up.

Ok, that's why I tried it with merging testing from toltec-dev into testing from plan5. 
But I can see how this reverts some files to previous states, looking at the PR changes now.

Before doing further radical steps: What do you think about this solution?
1. Clone into toltec-dev/toltec testing
2. Clone into plan5/toltec testing (where the gocryptfs package resides)
3. Delete all files (apart from .git/* and packages/gocryptfs/*) from the local plan5 testing branch
4. Copy all files (apart from .git/* and packages/gocryptfs/*) from the local toltec-dev/toltec testing branch into the local plan5 testing branch
5. git add -A&&git commit -m "Align with current upstream"
6. Push to plan5/toltec testing.

Or would this lead to potential further breakage? I can imagine that this would have to be redone before an actual merge in case there were further changes to the testing branch until then. But as I understand it, they would be applied on top of this commit so it wouldn't be a problem.

The more manual method seems to do the same to me (look at the commit changes of the PR and identify all files that are changed but shouldn't be and undo the changes locally, then commit and push).

Eeems:
At this point I'd just suggest recreating this branch with the files in the state you want them to be in.

plan5:
So that would imply creating a new PR or is there a way to keep this one? I assume recreating the branch would break the PR.

I've tried my suggestion and re-added the gocryptfs package. Right now it would merge without changing anyhing apart from the gocryptfs package, btw.

Eeems:
> So that would imply creating a new PR or is there a way to keep this one? I assume recreating the branch would break the PR.

No need to create a new PR, just force push the new branch.

> I've tried my suggestion and re-added the gocryptfs package. Right now it would merge without changing anyhing apart from the gocryptfs package, btw.

By the looks of it this worked, so we can review it as is instead of replacing the branch.

Eeems (approved):

Issue #322, opened 2021-03-26
Support for system update 2.6


matteodelabre:
A new official system update, 2.6, is being rolled out to users. [Most system libraries have been updated](https://github.com/toltec-dev/toolchain#version-matrix) in this new version, which makes apps compiled for previous versions incompatible. In particular:

* Qt was updated from 5.11.3 to 5.15.1 and the libqsgepaper.a library does not work as-is. This is because the library interfaces with private Qt classes whose API have changed between the two versions (see Eeems/oxide#187, sandsmark/recrossable#15, khanhas/zshelf#3, dixonary/fingerterm-reMarkable#8, dps/remarkable-keywriter#18).
* libcrypto.so was updated from 1.0.2 to 1.1, which causes the apps that use this library to not work unless the old lib is copied (see pl-semiotics/rM-vnc-server#8).
* Method signatures in xochitl and remarkable-shutdown, used by rm2fb have changed (see ddvk/remarkable2-framebuffer#18, ddvk/remarkable2-framebuffer#55, rmkit-dev/rmkit#117).

[A new release of the Toltec toolchain](https://github.com/toltec-dev/toolchain/releases/tag/v2.0) was made to support building apps that target the new 2.6 update. However, most packages compiled for 2.6 will not work for 2.5, and vice-versa. Our current plan is to take advantage of the new features in #310 to create two separate repos containing packages for each of the two versions.

raisjn:
rm2fb needs to update server to point at xochitl instead of remarkable shutdown, as there are new waveform modes (issue 56 in rm2fb) 

as part of this, rm2fb server should check that the preload exists before running, otherwise it will invoke xochitls main and run two at once if someone blows the home partition away

matteodelabre:
Just released toolchain v2.0.1 to fix some crashes that were seen on Qt apps such as fingerterm and recrossable.

matteodelabre:
Note to users: If you upgraded to 2.6, you can use the `switch.sh` script (followed by a reboot of your device) in https://github.com/ddvk/remarkable-update to temporarily roll back to 2.5.

JanLuca:
@matteodelabre May I ask why the branch `package/support-2.6` with the updated packages is not merged yet? I compiled some of them manually and they just seem to work well on my rM2 with release 2.7. Are there still some issues and/or is some beta testing wished? :)

Eeems:
[...]

PR #325, opened 2021-04-02
Use xochitl instead of remarkable-shutdown in rm2fb-server

matteodelabre:
In the last software relase (2.6), new waveform modes were added to xochitl to make the pinch-to-zoom feature work. Unfortunately, the remarkable-shutdown binary was not updated to support those new waveforms. As a consequence, using pinch-to-zoom with rm2fb results in a less smooth experience than without rm2fb, since the screen gets refreshed less frequently during the gesture.

See:

* <https://github.com/toltec-dev/toltec/issues/322#issuecomment-808235577>
* <https://github.com/ddvk/remarkable2-framebuffer/issues/56>

To address this issue, this PR changes the rm2fb-server to preload the `/usr/bin/xochitl` binary instead of `/usr/bin/remarkable-shutdown` in both the `rm2fb-server` command and `rm2fb.service`.

A start condition is added to `rm2fb.service` to only start if the librm2fb_server library actually exists on the filesystem. This is to
avoid having two concurrent xochitl processes running, as the loader ignores non-existent files provided in LD_PRELOAD.

Test plan:

* Manually run `rm2fb-server` and `rm2fb-client xochitl`, make use of the pinch-to-zoom gesture, and make sure no message of the form `Unable to complete update: invalid waveform ( 6 )` appears in the server process.
* Check that the display continuously updates when using the pinch-to-zoom gesture, even if the `rm2fb.service` service is running and xochitl is started with `rm2fb-client xochitl`.
* Remove the `/opt/lib/librm2fb_server.so.1` symlink and restart `rm2fb.service`. The service should not start. Make sure that there’s no duplicate `xochitl` process in this case.

### DIFF
--- a/package/rm2fb/overwrite-server-argv0.patch
+++ b/package/rm2fb/overwrite-server-argv0.patch
@@ -1,0 +1,15 @@
+diff --git a/src/server/main.cpp b/src/server/main.cpp
+index 755be17..4186060 100755
+--- a/src/server/main.cpp
++++ b/src/server/main.cpp
+@@ -173,6 +173,10 @@ int __libc_start_main(int (*_main)(int, char **, char **), int argc,
+   swtfb::SDK_BIN = argv[0];
+   fprintf(stderr, "BIN FILE: %s\n", argv[0]);
+ 
++  size_t argv0_len = strlen(argv[0]);
++  strncpy(argv[0], "rm2fb-server", argv0_len);
++  argv[0][argv0_len] = 0;
++
+   return func_main(server_main, argc, argv, init, fini, rtld_fini, stack_end);
+ };
+ };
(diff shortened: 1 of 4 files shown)


raisjn:
yes, it works fine but problematically remux is sending it SIGSTOP signals when remux runs because rm2fb now has the bin of 'xochitl' (since remux doesn't assume it is "owning" a process, it scans for all processes that use the bin specified in the draft file).

matteodelabre:
Fixed this with a patch to the server library to overwrite `argv[0]` after initialization. That way, the value visible in `/proc/<pid>/cmdline` is `rm2fb-server` (or a prefix of that string if `argv[0]` was not large enough to begin with). I did not change the value in `/proc/<pid>/comm` since remux does not use it, so it will still contain `xochitl`.

raisjn (approved):
verified it now works with remux

PR #326, opened 2021-04-02
[puzzles][0.2.2] add simon tatham's portable puzzle package

raisjn:


raisjn:
main screen
<details>

main

</details>

game screen (blackbox):
<details>

blackbox

</details>

game screen with side panel:
<details>

blackbox_side

</details>

raisjn:
gentle ping - thoughts on naming? (for the package and for the entry in draft)

danshick:
The Android port simply calls it "Puzzles" in the app menu. That's also the name it gets when installed as a monolithic app on Windows with the official installer. I think that'd be fine here too because it is the first puzzle collection we have.

If you want to differentiate it, I'd just go for the whole "Simon Tatham's Puzzles".

raisjn:
I've been trying to avoid taking generic names, but i think this one is ok since it is very common

Eeems:
Tested and this works well on the rM1. I encountered some odd display issues on some of the puzzles (primarily untangled) but nothing to keep the package from being merged.

LinusCDE (approved):
Seems to work fine. I also got the render issue with untangled, but find it really minor. Tested on both models.

Eeems (approved):

PR #331, opened 2021-04-06
[gocryptfs] Add fuse kernel module

plan5:
This will be a dependency to the gocryptfs package for the rM1 which doesn't have fuse built into the kernel.

I could not test it because I don't have a rM1 available so it should be considered WIP.

- [ ] Update to version 2.0-beta3 and test


plan5:
I've just noticed that I've accidentally based this PR on a branch forked from plan5-gocryptfs instead of testing. I'm closing it and creating a clean PR.

package/fusemodule/package
```diff
@@ -0,0 +1,73 @@
...
+
+_kernelrepo=https://github.com/remarkable/linux
+_kernelrev=1774e2a6a091fdc081324e966d3db0aa9df75c0b
+_defconfig=arch/arm/configs/zero-gravitas_defconfig
+
+image=base:v1.3.2
+source=(
+)
+sha256sums=(
+)
```
Eeems:
```suggestion
source=()
sha256sums=()
```

matteodelabre: Or omit the two fields altogether since [they’re optional](https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#source-field).

package/fusemodule/package
```diff
@@ -0,0 +1,73 @@
+e!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(fusemodule)
```
Eeems: I'd recommend just calling this fuse instead of fusemodule.

Eeems (changes requested):
See other comments for changes. Otherwise testing it seems to work fine.

package/fusemodule/package
```diff
@@ -0,0 +1,73 @@
...
+pkgdesc="FUSE (Filesystem in Userspace) Kernel Module"
+archs=(rm1)
```
danshick: Consider adding rm2 as a second arch but using logic [similar to the display package](https://github.com/toltec-dev/toltec/blob/3f12d75d214a3e85654e523c285d172e514ddb19/package/display/package#L54-L62) to avoid installing anything for rm2.

package/fusemodule/package
plan5: I've removed them

package/fusemodule/package
plan5: Done!

package/fusemodule/package
plan5: It's done :) Why is this better? (Nevermind, I've found the discussion in #318 )

package/fuse/package
```diff
@@ -0,0 +1,70 @@
...
+
+package() {
+    if [[ $arch = rm1 ]]; then
+        mkdir -p "$pkgdir/lib/modules"
+        cp -r "$srcdir/pkg"/* "$pkgdir/lib/modules"
+    fi
+}
+
+configure() {
+    depmod -a
```
Eeems: Should this and postremove also be doing $arch checking of some sort?
@matteodelabre is $arch available in those methods?

matteodelabre: Yes, it should check for $arch before trying to add/remove the module. See [the `display` recipe](https://github.com/toltec-dev/toltec/blob/a7a3a78cc23cef3b81344119839a793601062ef7/package/display/package#L65) and [the docs](https://github.com/toltec-dev/toltec/blob/testing/docs/package.md#install-section) for reference.

Eeems (changes requested):
Add `$arch` check to `configure`, `postremove` and `postupgrade`

package/fuse/package
```diff
@@ -0,0 +1,76 @@
+e!/usr/bin/env bash
```
Eeems:
```suggestion
#!/usr/bin/env bash
```

package/fuse/package
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(fuse)
```
Eeems, line 6:
```suggestion
archs=(rm1 rm2)
pkgnames=(fuse)
```

Eeems (changes requested):
Right now this generates a rmall package that doesn't contain the module, and will never run any of the scripts. Instead a version of the package is required to be generated for rm1 and rm2. The rm1 package will contain everything, the rm2 package will be empty.

Eeems:
Has anybody tested the rM2 version of the package to make sure it has no side effects?

LinusCDE:
> Has anybody tested the rM2 version of the package to make sure it has no side effects?

Works fine for me.

Eeems (approved):

Issue #290, opened 2021-02-21
Make a github release for each merge to stable


lambada:
Whenever you merge to stable could you cut a new tag + github release? That way I can subscribe in github for release notifications so I get prompted to go and update the remarkable :slight_smile:
The only other closest option would be to get notified of all PR activity on it which... is a bit much as just an end user :slight_smile:
GitHub.

It looks like github actions exist for making a release so it should be fairly straight forward to add into the github actions stable workflow.

danshick:
I think an [ISO week date](https://en.m.wikipedia.org/wiki/ISO_week_date) would make a reasonable tag.

LinusCDE:
Seems like a nice idea. I first though it be too much hassle, but GH Actions should do the trick.

Otherwise a quick solution would be to use almost any RSS reader to subscribe to commits to the stable branch on this url: https://github.com/toltec-dev/toltec/commits/stable.atom

matteodelabre:
I didn’t know GitHub published RSS feeds, thanks @LinusCDE! (I couldn’t find any documentation for them though.) Does this fit your use case @lambada?

lambada:
Not particularly, if only because I don't actually use an RSS reader these days, though I could probably make it work by finding an RSS -> email service

danshick:
Even with the RSS option, I think tagging releases is doable.

PR #335, opened 2021-04-08
Create GitHub release on push to stable.

lambada:
Resolves https://github.com/toltec-dev/toltec/issues/290

This adds a step to the stable workflow that will create a release.

The release will be named in accordance with the[ISO Week Date](https://en.wikipedia.org/wiki/ISO_week_date) format (e.g. 2021-W03-2 for a release made on the Tuesday in the third ISO Week of 2021).

The name of the release will be taken from the message of the commit that has been pushed to master - e.g. "Stable merge window for week XX of YYYY".

A proof of concept for this working can be seen in my fork: https://github.com/lambada/toltec/releases/tag/2021-W14-4

The `hub` cli tool this calls out to is https://github.com/github/hub and is preinstalled.

**Limitation: Only one release per day will be able to be created** but I feel that should be sufficient given the use case in the original issue - a notification mechanism to allow people to know that there are updates to install. As this step is marked with `continue_on_error` this is harmless in terms of continuing on to the next steps.

**This requires a new secret env var to be added to the repo configuration** This should be a Personal Access Token / API Key that has `repo` scope, and should be added with the name `RELEASE_TOKEN`. The default GITHUB_TOKEN that workflow has access to does not have permission to create a release.

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -37,7 +37,7 @@ jobs:
                 remote-path: ${{ secrets.REMOTE_SSH }}:/srv/toltec/web
             - name: Create Github Release
               continue-on-error: true
-              run: hub release create -t ${{ github.sha }} -m "${{ github.event.commits[0].message }}" $(date +%Y-W%V)
+              run: hub release create -t ${{ github.sha }} -m "${{ github.event.commits[0].message }}" $(date +%Y-W%V-%u)
               env:
                 GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
             - name: Send notification to Discord
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -49,4 +49,4 @@ jobs:
                 title: New Toltec stable update available
                 link: https://toltec-dev.org/stable
                 color: 0x2ea043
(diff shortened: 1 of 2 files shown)


matteodelabre:
Thanks for taking the time to set this up @lambada! Your changes look good to me.

Just a minor note: you used `github.event.commits[0]` to refer to the pushed commit, and just below we use `github.event.head_commit` when sending the Discord notification (and also in the `testing.yml` workflow). I think we should harmonize the two so as to avoid running into strange behavior differences between the release creation and the Discord notification. I wasn’t able to find documentation on the `head_commit` attribute, so `commits[0]` seems preferable.

lambada:
Good catch, didn’t even notice that we were already doing something similar for commit message!

I only found one reference to head_commit which was someone claiming it had been removed (obviously not entirely true given it still seems to work for now) but it’s definitely no longer documented.

i’ve updated that discord notification (and also the staging one), though I don’t have an easy to use test environment for discord actions right now to prove it’s the same format

danshick:
That's okay. At least for me, minor format changes in testing aren't a big deal. We can adjust before it is promoted to stable if required.

matteodelabre (dismissed):

matteodelabre (approved):

Issue #337, opened 2021-04-10
[rm2fb] re-consider way of specifying LD_PRELOAD for launchers


raisjn:
compiling binaries for usage in toltec testing channel requires using patchelf on them. 

this is a strange situation since rM1 does not require rm2fb or necessarily have it installed, so a binary that has been patchelf'd is incompatible with rM1 that does not have toltec installed with rm2fb. i think this means that i need to supply people with two sets of binaries and compilation instructions: one for toltec enabled rM1/rM2 and one for rM1 without toltec

raisjn:
actually, never mind me

danshick:
I'm confused. If a package is in Toltec, it has a build recipe designed to produce binaries that are distributed as part of Toltec. If you build an app with a Makefile or similar build infrastructure directly from it's repo, it likely wont produce patchelf'd binaries. The patchelf'ing only happens for Toltec in a Toltec package recipe.

raisjn:
previously: all launchers had LD_PRELOAD enabled for them through a systemd override. any app launched through launcher will have rm2fb placed onto it, regardless if app comes from toltec or not. 

now: launcher is patchelf and there is no global LD_PRELOAD. if the launcher wants to launch an app, it needs to setup LD_PRELOAD for that app or know that is has been patchelfed. 

now: if i want to give someone a binary, the binary needs to be patchelf (if running on rM2) or it needs to **not** be patchelf (if running on non-toltec rM1). it is ok to give patchelf binary to rM1 with toltec and non-patchelf binary to rM2 without toltec. 

as an example: remux either needs to be compiled with patchelf or not. if someone want to use my repo to compile remux and use it with toltec, they will need to use the patchelf version. if they have an rM1 and no toltec, they will need to use the non-patchelf. which do i choose by default? either way, some percentage of people will be confused when they didn't need to be

danshick:
Please take everything below with a grain of salt. Maybe I'm still wrong in my understanding of your concern.

I think you should distribute a non-patchelf version of your launcher from your repo and let patchelf happen only in Toltec. 

I think it is reasonable to expect that folks who aren't using Toltec will:

A.) Have an rM1 and not need a preload or rm2fb and will not want any patchelf or preload of any kind.

-or-

B.) Have an rM2 and have installed rm2fb manually and understand the need for both it and the shim/preload. Instructions for that (including the global systemd override setup) should be kept with rm2fb in its repo with manual install instructions.

[...]

PR #338, opened 2021-04-10
[display] add remux.service.conf override to display package on rm2

raisjn:
this turns on the display shim for remux on rm2 apps. the reasoning for this is so that people can launch homebrew apps that they compiled from other sources without having to use patchelf. 

TEST PLAN:

* install display package on rm2, verify that /etc/systemd/system/remux.service.d/remux.service.conf exists and has an Environment= line in it
* restart remux, verify that applications can be launched
* custom compile an application without patchself (or remove the DT needed using patchelf) and verify remux can launch it


raisjn:
over time, if everyone starts using patchelf, this may be less and less needed. before then, i will work on being able to disable the preload on an app by app basis so people can disable it if they like. 

this PR is related to https://github.com/toltec-dev/toltec/issues/337

matteodelabre:
Thanks for raising this point @raisjn. I agree with you. The current state of things is that all apps that work on rM2 assume that they’ll be used with LD_PRELOAD, so making a hard transition to requiring each of those apps to be patchelf-ed will confuse users. For consistency, this should be applied to all the launchers, not only remux (any thoughts on this @Eeems @dixonary?).

In the long run (if everyone agrees that this is a good thing to do), I think we should update rm2fb’s docs to recommend that users use patchelf when they self-compile their apps and do not rely on LD_PRELOAD being there. We could also open issues with rM2-compatible projects to ask if they can update their build instructions and pre-built binaries too.

Eeems:
I'm okay with applying this to Oxide as well for now. I do think that patchelf is the right longer term solution though.

matteodelabre:
I moved those overrides to the `rm2fb-client` package as they depend on the librm2fb_client library, which is provided by that package instead of `display`. I also renamed the files to a name that more explicitly states their intent.

matteodelabre (approved):
I tested it on my rM2 by removing the librm2fb_client dependency on the calculator binary and starting the app from each of the three launchers. This is ready for merging if you agree with my minor changes above @raisjn

PR #340, opened 2021-04-13
Move website generation to separate repository

matteodelabre:
This PR removes all the logic and resources related to generating the Toltec website. These have been moved to the [toltec-dev/web](https://github.com/toltec-dev/web) repository.

This change enables us to iterate on the website without being constrained by the stable release cycle from the main repository.

To make sure that the website always contains the latest stable bootstrap script and checksum, the stable build workflow of this repository sends `repository_dispatch` events to the toltec-dev/web repository. See <https://github.community/t/triggering-by-other-repository/16163/2> for details.


Eeems:
Does this handle validating that the web repository has the up-to-date bootstrap sha256sum?

matteodelabre:
Yes, it does! The web build process [fetches the latest bootstrap](https://github.com/toltec-dev/web/blob/dca22ff1d4678188205f45d26a19090a9aaf5952/Makefile#L19) from the main repo’s stable branch, computes its hash, and generates the index page based on that. Note that the bootstrap script is no longer uploaded to the live server by the main repository’s workflow, but by the web repository. This guarantees that the checksum on the website always matches the file at <https://toltec-dev.org/bootstrap>.

Eeems (approved):
Changes in this PR seem good to me. I'm not sure what other testing we'd really like to do before merging. It might make more sense to merge, and then fix any issues we run into as part of using it in a new PR.

raisjn (approved):
this looks good, i'm excited for a better communication channel with the community - thanks! to be clear, this is only moving the main site (not the package listing) to a separate repo, right?

matteodelabre:
> to be clear, this is only moving the main site (not the package listing) to a separate repo, right?

Yup! The package listing generation logic belongs more with the build system than with the homepage generation in my opinion.

PR #344, opened 2021-04-18
[plato] Update to 0.9.17-1

LinusCDE:
This update doesn't add any custom reMarkable functionality.

It does however handle a breaking change that happened with [release 0.9.14](https://github.com/baskerville/plato/releases/tag/0.9.14) of the upstream plato. I ported the migration script from python to baseh (with jq as new dependency) but only support the default library location. In any case the user should get a warning anyway.

The migration messages may need some refinement. I'll probably do those tomorrow.

Tested it on the rM 1 and 2 (both with firmware 2.5.X). Is a 2.X toolchain required for the 2.6.X firmware? The other doesn't seem to work at least (using the package with 2.1 toolchain on a 2.5.X device) gives the following error:

```shell
reMarkable: ~/ rm2fb-client plato
./plato: /lib/libc.so.6: version `GLIBC_2.28' not found (required by ./plato)
```

If the 1.6 toolchain works on the 2.6.X firmware in that regard, I don't see a reason to update it yet since plato doesn't use QT.

Steps for testing:
 - Have 0.9.13 or earlier installed
 - Have some document in ~/plato-media that e.g. has some other last opened page
 - Update to this version
 - The update should automatically migrate the metadata and thus e.g. keep the last opened page
 - (It should not attempt to migrate an already up-to-date metadata file and warn if an "-old" variation was already created.)

(Some more info about this release can be found on [my release page](https://github.com/LinusCDE/plato/releases/tag/0.9.16-rm-release-8).)

---

P.S.: I'll hopefully be able to do more again in the coming days.

### DIFF
--- a/package/plato/package
+++ b/package/plato/package
@@ -102,7 +102,7 @@ configure() {
             jq "$change_cmds | del($del_list)" < $metadata_file_old > $metadata_file
             echo 'Metadata file was updated! The old one was renamed to ".metadata-old.json".'
 
-        elif [ $max_key_len -gt 16 ] &&  [ -e $metadata_file_old ]; then
+        elif [ $max_key_len -gt 16 ] && [ -e $metadata_file_old ]; then
             echo 'MANUAL INTERVENTION REQUIRED!!!'
             echo 'Plato 0.9.14 changed the format of metadata files.'
             echo 'Your metadata file not found but may be outdated!'
--- a/package/plato/package
+++ b/package/plato/package
@@ -76,8 +76,9 @@ configure() {
         local metadata_file=$mediadir/.metadata.json
         local metadata_file_old=$mediadir/.metadata-old.json
 
-        local max_key_len=$(cat $metadata_file | jq --raw-output 'keys[]' | wc -L)
-        if [ $max_key_len -gt 16 ] && [ ! -e $metadata_file_old ]; then
+        local max_key_len
+        max_key_len=$(jq --raw-output 'keys[]' < $metadata_file | wc -L)
+        if [ "$max_key_len" -gt 16 ] && [ ! -e $metadata_file_old ]; then
 
             echo 'Plato 0.9.14 changed the format of metadata files.'
             echo 'Your metadata file was found and determined to be outdated!'
@@ -89,7 +90,7 @@ configure() {
             local change_cmds=""
             local del_list=""
             for key in $(jq --raw-output 'keys[]' < $metadata_file_old); do
-                hex_key=$(printf %016X $key)
+                hex_key=$(printf %016X "$key")
                 #echo "Key: $key --> $hex_key"
 
                 [ -z "$del_list" ] || del_list="${del_list}, "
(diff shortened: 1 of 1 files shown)


LinusCDE:
The messages should be alright now.

package/plato/package
```diff
@@ -62,12 +62,56 @@ package() {
...
+            echo "Update complete. The old file was renamed to '.metadata-old.json'."
+            echo ""
+
+        elif [ "$max_key_len" -gt 16 ] && [ -e $metadata_file_old ] || [ ! -e $metadata_file ]; then
+            echo ""
+            echo "In Plato 0.9.14 the metadata format for libraries changed."
+            echo "MANUAL INTERVENTION REQUIRED\!\!\!"
+            echo "Please follow the following guide to update your library metadata manually:"
+            echo "https://github.com/baskerville/plato/releases/tag/0.9.14"
+            echo "Please LAUNCH plato ONLY AFTER this was done\!"
```
raisjn: i didn't go through this, i just blew it away and restarted. i think this should be done automatically (since it makes a backup, i think?) and the user notified

raisjn (dismissed):

package/plato/package
Eeems: I agree, this should be done automagically by the package. It's one of the use-cases for pre/post upgrade scripts.

LinusCDE: I used configure as I feared pre/post upgrade might miss cases where a user fully reinstalled a package. The script checks for old versions of the metadata and I attempted to make it check more often/reliably this way.

I can change this tomorrow though. Is still a good point.

raisjn: i think i'm a bit confused. when i upgraded plato, i got the message (`MANUAL INTERVENTION REQUIRED`), but it looks like you automatically do the migration for people. Did I get that message because I didn't have any metadata?

LinusCDE: I think this happened because you seem like you didn't do the migration but also have a `.old` (backup) file. I just didn't support that usecase and added that hint instead of trying to deal with it.

LinusCDE: Also just looked at the [docs](https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#install-section):

> When upgrading a package from version A to B, the following happens:
> 
>  - `preupgrade B`, if it exists, is called from version A
>  -  Old package files are removed (except configuration files)
>  -  `postupgrade B`, if it exists, is called from version A
>  -  New package files are unpacked and installed
>  -  `configure`, if it exists, is called from version B

If I read this correctly, calling postupgrade wouldn't make sense since it would only call if this package ever gets upgraded to a future one.

matteodelabre (approved):
Thanks for the PR @LinusCDE! I made some minor changes to the migration logic: to make the script only output a message if there is actually a migration to be done (i.e., if there is an existing `.metadata.json` file with the old format), and to make the backup copy have a randomly-generated extension, so that the migration works even if there’s already a file called `.metadata.json.old`. The rest of the PR looks good to me, I tested it on my rM2 with no issues.

PR #352, opened 2021-04-30
[TilEm][0.0.6] New Package

timower:
Adds tilem, my port of the ti calculator emulator.
Currently only support ti84+. It should popup to download a ROM file if none is available.

### DIFF
--- a/package/tilem/package
+++ b/package/tilem/package
@@ -12,8 +12,8 @@ license=GPL-3.0
 section="utils"
 image=base:v2.1
 
-source=(https://github.com/timower/rM2-stuff/archive/refs/tags/v0.0.6.tar.gz)
-sha256sums=(ff3e5a903cb5e375a66eb0445ef107ce93823cad723b37b85ec39dff87f6c015)
+source=(https://github.com/timower/rM2-stuff/archive/8dc94a360358bf74da244beda74739663dd44741.tar.gz)
+sha256sums=(23192117d3e8bb3ddba5d909e80dea7354aa4c5e2936047ca2cb3d35e906ecc1)
 
 build() {
     mkdir build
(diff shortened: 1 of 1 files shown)


LinusCDE (changes requested):
Tested on both rM 2 and 1. On both, I get the dialog to download the ROM, but afterwards it doesn't accept any input anymore. Logs don't seem to indicate any error. Did I miss something to get it running properly?

UI seems to look fine though:

(PNG-Grafik, 1404 × 1872 Pixel) - Skaliert (46%)

timower:
I think you need to press ON

LinusCDE:
Ok, my bad. 😅 Haven't used it in a while. I actually used the TI84+ in school. And often the WabbitEmu app for Android as well.

Works on rM 2. On the rM 1 the lcd is displayed wrong and not really readable:

 &nbsp;

Oxide also doesn't officially support arguments in the `call=` of draft anymore. So a wrapper script is probably necessary for proper launcher support. It is also probably more ideal to clear the background when launching (I launched from terminal here. Not sure if some launchers automatically clear the fb).

timower:
Okay, I should've fixed the issues. I've used a commit in the package for now. If everything works I can push a new tag and update the package to use that.

I didn't use a wrapper script but just use the default path if no cmd line arg is specified. If you want to change the rom location you can create a script or provide the arg yourself.

LinusCDE:
Works great now!

Btw I don't think you don't need the [toltec license header](https://github.com/timower/rM2-stuff/blob/8dc94a360358bf74da244beda74739663dd44741/apps/tilem/draft/tilem.draft#L1) in your draft file if it is not part of the toltec-repository.

I'll guess since you want to add the tag first, I'll wait until then and approve it afterwards. :+1:

timower:
Done, yeah I should remove the license header. Technically I copy pasted it from the repo so it's still licensed by that license 😃

LinusCDE (approved):

PR #353, opened 2021-04-30
[Yaft] Set build type to release

timower:
The default build type is debug which can produce quite slow binaries.


raisjn (approved):
tested on rM2, works fine

Issue #341, opened 2021-04-14
Improve documentation for how to switch to the testing branch


gtsiam:
I tried to switch to the testing branch for the newer koreader version. The [documentation](https://github.com/toltec-dev/toltec/blob/stable/docs/branches.md) isn't all that clear in what I need to do for an existing toltec installation:

> To use the testing branch, update the /opt/etc/opkg.conf file on an existing install

Yes, update _how_? Well, after some searching, I'm pretty sure what I'm supposed to do is change the package feed, from:
```
src/gz toltec https://toltec-dev.org/stable
```
to
```
src/gz toltec https://toltec-dev.org/testing
```

However this still gives me the following error when I try `opkg update`:
```
Downloading https://toltec-dev.org/testing/Packages.gz
wget: note: TLS certificate validation not implemented
wget: server returned error: HTTP/1.1 404 Not Found
*** Failed to download the package list from https://toltec-dev.org/testing/Packages.gz

Collected errors:
 * opkg_download: Failed to download https://toltec-dev.org/testing/Packages.gz, wget returned 1.
```
At least at the time of writing this issue, both https://toltec-dev.org/testing/Packages.gz and https://toltec-dev.org/testing/Packages return 404.

matteodelabre:
**Note:** This comment is outdated, there is now a simpler way to switch to the testing branch. Refer to the comment at the end of this thread.

~Hi @gtsiam, thanks for your report. You’re right that the documentation isn’t very precise as to how one can switch to the testing branch. This is definitely something that we plan on improving in the future; we’ve been discussing the addition of a script to switch branches automatically for some time now. There are several other issues with a higher priority that we’re working to address first, though.~

~We’ve recently changed the repository layout of the testing branch to better separate packages that work on rM1 only from those that work on rM2 only (see #310 for details). This layout change hasn’t landed in stable yet, hence the difference you’re seeing between <https://toltec-dev.org/stable> and <https://toltec-dev.org/testing>. With this new layout, the packages are now sorted into subfolders depending on which devices they support.~

~To switch to the testing branch, you’ll therefore need to update your Opkg configuration (located in `/opt/etc/opkg.conf`) for the new layout, as follows.~

```diff
-src/gz toltec https://toltec-dev.org/stable
+arch rmall 200
+arch rm1 250
+src/gz toltec-rmall https://toltec-dev.org/testing/rmall
+src/gz toltec-rm1 https://toltec-dev.org/testing/rm1
```

[...]

PR #342, opened 2021-04-14
Update the bootstrap script to install wget-ssl

matteodelabre:
Entware recently renamed its `wget` package to `wget-ssl`, and made `wget` a virtual package provided both by `wget-ssl` and `wget-nossl` (see <https://github.com/openwrt/packages/pull/13613>). The latter seems to be favored by the package manager, so doing `opkg install wget` will actually install the version of wget that does not support TLS.

This PR changes the bootstrap script to ensure that the TLS-enabled version of wget gets installed. To test it, run the new bootstrap script on a clean system and make sure that no message saying “TLS certificate validation not implemented” is ever printed out.

PR #356, opened 2021-05-08
Add uninstall script and make the bootstrap script install-only

### LINKED ISSUE #295: Provide automatic uninstall of Toltec and all installed packages
We should have a single command approach to uninstalling all packages and reverting the bootstrap script.

### LINKED ISSUE #332: Bootstrap clean-up should revert back the system to its original state
When an error occurs in the bootstrap script, it is currently set up to [remove everything Toltec-related from the device](https://github.com/toltec-dev/toltec/blob/5c0d508b6e68fe85de2cbffa6101b288f4146351/scripts/bootstrap/bootstrap#L61). This works well when the script is run on a default system, but if the script is run on top of an existing install it could potentially wipe out data located in `/opt`. We should change the script so that it initially detects the current system state and, in case an error occurs, rolls back to that state.

matteodelabre:
This PR adds a new `toltecctl` script, installed by the `toltec-bootstrap` package, that supersedes the former `entware-reenable` script and provides a way to uninstall Toltec (fixes #295). This script could be extended to add new install management features in the future. The uninstall script does the following things:

* Uninstalls all Toltec and Entware packages, running removal scripts for each package.
* Removes the bind mount of Toltec files on /opt (the mount is removed lazily so that it does not fail even if some processes that have files opened in that folder are still running)
* Removes all Toltec files
* Clears Toltec definitions from .bashrc
* Re-enables xochitl
* Reboots the device

Another major change of this PR is to make the bootstrap script fail if files from previous installs remain in the folders used by Toltec (fixes #332). With this change, the bootstrap script only has a single role, that is to create a new Toltec install in a clean system. Roles previously assumed explicitly or implicitly by the bootstrap script are deprecated:

* To re-enable an existing install after a system update, users should run `toltecctl enable` instead of re-running the bootstrap script.
* To create a new clean install, users should first uninstall the existing one and then run the bootstrap script.
* To switch to another branch (see #341), users should manually edit the `/opt/etc/opkg.conf` file instead of re-running the bootstrap script (a future PR could extend the `toltecctl` script to handle switching branches).

Test plan
=========

(Only tested on rM2 so far.)

Uninstall script
----------------

* Upgrade the toltec-bootstrap package to version 0.0.5-1 (this will remove the `bootstrap` and `entware-reenable` scripts from the home folder and add the `toltecctl` script to the `~/.local/bin` folder)
* Run `~/.local/bin/toltecctl enable` and re-source .bashrc (to add .local/bin to the PATH)
* Run `toltecctl uninstall`
* After the tablet reboots, make sure no traces of the Toltec install remain (in ~/.entware, /opt, /lib/systemd/system/opt.mount)
* If you had disabled xochitl, make sure that it starts up again after uninstalling

Re-enable script
----------------

* Upgrade the toltec-bootstrap package to version 0.0.5-1 (this will remove the `bootstrap` and `entware-reenable` scripts from the home folder and add the `toltecctl` script to the `~/.local/bin` folder)
* Simulate a system update, for example by switching to the other root partition or using [remarkable-update](https://github.com/ddvk/remarkable-update)
* Run `~/.local/bin/toltecctl enable`
* Make sure that Toltec has been re-enabled (files visible in /opt, opkg command working, apps running)

Bootstrap script
----------------

* Copy the new bootstrap script to the tablet.
* Try running the bootstrap script with an existing Toltec install (the script should fail and tell you that Toltec is already installed)
* Try running the bootstrap script with only some parts of a Toltec install remaining (for example, only the .entware folder - the script should fail and tell you which files need to be cleaned up)
* Run the bootstrap script after uninstalling Toltec or from a clean system, and make sure the install works.


package/toltec-bootstrap/package
```diff
@@ -3,24 +3,17 @@
 # SPDX-License-Identifier: MIT
 
 pkgnames=(toltec-bootstrap)
-pkgdesc="Toltec installation script"
+pkgdesc="Manage your Toltec install"
 url=https://toltec-dev.org/
-pkgver=0.0.4-1
-timestamp=2021-04-04T21:55Z
+pkgver=0.0.5-1
```
Eeems, line 8:
```suggestion
pkgver=0.0.6-1
```
There is another PR with the version being bumped to 0.0.5-1

LinusCDE (changes requested):
Tested most of your steps on both devices and they seem to be fine for the most part.

For bootstrap I used `toltec_branch=testing` since the architecture urls don't exist on stable yet.

A cool feature would be to have toltec-bootstrap automatically add `/home/root/.local/bin` to `PATH` upon upgrade to the newer version.

Disabling toltec also doesn't affect the services. It removes `/opt`, but systemd may still be configured to attempt launching remux. In my case it also didn't fall back to xochitl, so resetting the service files similar to on uninstall may be needed. Or at least warn the user about what to do.

TL;DR:
 - Also add new path to .bashrc upon package update
 - Warn user or manually force xochitl to be enabled on `toltecctl disable`

This is a really cool utility. It'll definitely make toltec more user friendly! :+1:

matteodelabre:
Thank you very much for your review @LinusCDE!

> A cool feature would be to have toltec-bootstrap automatically add `/home/root/.local/bin` to `PATH` upon upgrade to the newer version.

I added a new `toltecctl set-path` action that automatically fixes or upgrades the PATH in bashrc. I added this command to the package’s `configure()` block so that it is invoked when the package is updated.

> Disabling toltec also doesn't affect the services. It removes `/opt`, but systemd may still be configured to attempt launching remux. In my case it also didn't fall back to xochitl, so resetting the service files similar to on uninstall may be needed. Or at least warn the user about what to do.

This is a good point. Actually, I’m not sure whether the action of disabling Toltec without uninstalling it is useful to have in `toltecctl`. I assume most users will want to completely uninstall it. So I’ve removed this option to make things less confusing.

Also, thanks for working on the Bash completions! I updated the `_toltecctl` completion script to match the new set of actions.

LinusCDE:
Cool. I tested the `set-path` command on the rm1 and it works fine!

Would you want me to retest everything again or can I just accept it? I looked at the new commits and I don't see any issues or things that need a full re-review imo.

matteodelabre:
> Would you want me to retest everything again or can I just accept it? I looked at the new commits and I don't see any issues or things that need a full re-review imo.

On my side, I re-did the tests for the uninstall and bootstrap scripts (on rM2), and also tested that upgrading the package sets the PATH as expected. I don’t think it’s necessary to re-do all the other tests since there weren’t any major changes.

matteodelabre:
Maybe we should revert @Eeems’ suggestion from above of bumping `toltec-reenable`’s version from 0.0.4 to 0.0.6 if this PR is to be merged before #342.

matteodelabre:
Reverted the version to 0.0.5. I will set PR #342’s version to 0.0.6.

LinusCDE (approved):

isaacwisdom:
Hi, I'm trying to run the uninstall command added, but get this error:
"/home/root/.local/bin/toltecctl: line 181: uninstall_force: command not found"
Any ideas what's going on?

isaacwisdom:
> Hi, I'm trying to run the uninstall command added, but get this error:
> "/home/root/.local/bin/toltecctl: line 181: uninstall_force: command not found"
> Any ideas what's going on?

Just kidding, figured it out. uninstall-force() is defined but uninstall() references uninstall_force(). Changing the underline  to a hyphen fixes the issue.

PR #358, opened 2021-05-19
Update innernet to v1.3.0

Eeems:
@plan5 Looks like I beat you to this.

### DIFF
--- a/package/innernet/package
+++ b/package/innernet/package
@@ -5,7 +5,7 @@
 pkgnames=(innernet-client)
 pkgdesc="A private network system that uses WireGuard under the hood."
 url="https://github.com/tonarino/innernet"
-pkgver=1.3.0-beta-5-1
+pkgver=1.3.0-beta-7-1
 timestamp=2021-05-19T08:17:04Z
 section="utils"
 maintainer="plan5 <30434574+plan5@users.noreply.github.com>"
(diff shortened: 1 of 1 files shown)


plan5:
@Eeems My plan is foiled!

Eeems:
@plan5 since you are the maintainer of the innernet package, could you test on the rM2 to make sure the package works and then approve the PR if it does?

plan5:
@Eeems Yes, I will. I've PMed you on Discord for some help with the server.

plan5:
Tested on 2.6.x and 2.7.x and innernet launches fine on rM2 with previously created config.
Connection to server is made and IP is obtained. Server can be pinged via VPN IP.

Eeems:
@plan5 innernet 1.3.0 just came out, so we'll have to retest.

plan5 (commented):
opkg considers this a downgrade and refuses to downgrade from 1.3.0-beta-x-1 to 1.3.0-1.

I assume this is rather an issue for opkg than for this PR and since beta was never merged to stable I think this can be neglected (but I'm not approving right away because of that).

I've forced the installation and the tests mentioned above all still work.

Eeems:
Since this never made it to stable I'm inclined to agree with ignoring this issue. I think in the future we should skip the beta builds.

@matteodelabre thoughts?

matteodelabre:
I think it’s OK. Users who have chosen to use the testing branch should not expect smooth upgrades anyway.

@plan5 In the future, you can use `~` to tag beta versions, since this character is sorted lower than any other one (for example `1.3.0~beta-x-1` would have been sorted lower than `1.3.0-1` by opkg). This can be verified with the following command:

```console
$ opkg compare-versions "1.3.0-beta-x-1" "<<" "1.3.0-1" && echo "Lower" || echo "Not lower"
Not lower
$ opkg compare-versions "1.3.0~beta-x-1" "<<" "1.3.0-1" && echo "Lower" || echo "Not lower"
Lower
```

plan5 (approved):
Works for me on 2.7 on rM2

Ok, and I'll keep an eye on the sorting next time. If I get around to building a tinc package, I'll probably do the "pre" version because it features invites and it has been "pre" forever.

danshick (approved):

PR #359, opened 2021-05-21
Create symlink for rm2fb-server to fix exe name in /proc/<id>/stat

JanLuca:
The current patch just fix the name of rm2fb-server in `/proc/<id>/cmdline` but not the executable name in `/proc/<id>/stat`. Therefore in some process list still the server is shown as `xochitl`. This patch creates a symlink `/opt/bin/rm2fb-server` and use this as exec in the systemd unit.


matteodelabre:
While this works, a cleaner solution to me would be to use `prctl(PR_SET_NAME, "rm2fb-server");` to set the process name [in rm2fb](https://github.com/ddvk/remarkable2-framebuffer/blob/e594fc4430446f9321b716da8eb0193dc2b2b9f2/src/server/main.cpp#L176). I tested it and it does change the value that appears in `/proc/*/stat`.

JanLuca:
That's right, sounds like a good approach as well. I suggested the symlink idea since your approach needs the patch to change the argv[0] value, too. But I have no strong opinion so let's go with the prctl. I just have seen the problem and wanted to provide a first idea how to fix it :)

matteodelabre:
Do you have a preference? @raisjn

raisjn:
prctl sounds better, thank you (and thanks for finding this, jan)

matteodelabre:
I added the `prctl` call to upstream. I’ll close this PR and open a new one to update the package to the latest upstream release. Thanks for your help @JanLuca!

PR #361, opened 2021-06-01
[innernet] Update to v1.3.1

plan5:
Still needs testing


plan5:
@Eeems made it! :)

Eeems:
@plan5 does this include the new bash completions?

plan5:
> @plan5 does this include the new bash completions?

Ah yes, it does. Should they go to /opt/usr/share/bash-completion/completions/ ?

Edit: There are also completions for elvish, fish, zsh (and powershell), so they need a place, too (except powershell).

Eeems:
> 
> Ah yes, it does. Should they go to /opt/usr/share/bash-completion/completions/ ?
> 
> Edit: There are also completions for elvish, fish, zsh (and powershell), so they need a place, too (except powershell).

I'm not really sure. @LinusCDE I think you've poke around with bash completions, where are our standard locations?

LinusCDE:
I think the directory mentioned by @plan5 is right (`/opt/usr/share/bash-completion/completions/`). I think I've added the completions for `_opkg` there as well, but am not sure whether this is the best way of "extending" bash-completions which provides a lot of helper functions and infra structure found in /opt`/usr/share/bash-completions`.

I have the draft for `opkg` completions as a seperate `toltec-completions` package in #277 . Didn't get to finalize it yet. Any input there is welcome as well and the package should already work for the most common use cases.

LinusCDE:
Just made #277 ready for review. Once that is in, we will also have the bash-completions package to enable innernet completions to work.

You could probably make the package `makedepend` on that package then. But I don't think that this is a requirement.

Eeems (changes requested):
Add bash completions

plan5:
> Add bash completions

I'll try to look into it next week. If anyone wants to do it before then, you're welcome :)

Eeems:
@plan5 will you be able to add the bash completions anytime soon? I'd be willing to merge without them if you intend to add them as a separate package.

plan5:
> @plan5 will you be able to add the bash completions anytime soon? I'd be willing to merge without them if you intend to add them as a separate package.

Unfortunately I'm not able to work on this right now, hopefully things will slow down again a bit next month. So feel free to merge. I also didn't find time to wrap my head around the right way to provide completions yet 
m)

Eeems (approved):
Tested and it works. I did see that I had to manually run `systemctl daemon-reload && systemctl restart innernet@*`, so that should be added at some point.

PR #362, opened 2021-06-03
Update chessmarkable to 0.7.0-1

LinusCDE:
This change integrates a requested feature to add savestates (more on the [release page](https://github.com/LinusCDE/chessmarkable/releases/tag/0.7.0-1)).

Since the changes don't really affect anything to do with models, just testing on one model is enough.

Testplan:
 - Start a game (mode doesn't matter). Remember the chosen slot.
 - Make at least one move
 - Hit "Save & Quit"
 - Check that the file `/home/root/chessmarkable/savestates.yml` was created and your chosen slot contains a string
 - Start a game from your chosen slot
 - The game should have remembered the piece locations

### DIFF
--- a/package/chessmarkable/package
+++ b/package/chessmarkable/package
@@ -5,17 +5,17 @@
 pkgnames=(chessmarkable)
 pkgdesc="Chess game"
 url=https://github.com/LinusCDE/chessmarkable
-pkgver=0.6.0-4
-timestamp=2021-01-03T05:26Z
+pkgver=0.7.0-1
+timestamp=2021-06-03T10:47Z
 section="games"
 maintainer="Linus K. <linus@cosmos-ink.net>"
 license=MIT
 installdepends=(display)
 flags=(patch_rm2fb)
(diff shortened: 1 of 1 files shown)


Eeems:
I'm not really sure that `/home/root/chessmarkable` is the correct location for savestates. Following the XDG Base Directory specification (https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), I would expect it to live in `$XDG_DATA_HOME`, or `$HOME/.local/share`.

That said, I've run through the test steps and while the game does save and recall properly, including across launches, there is no folder/files at `/home/root/chessmarkable`. After a quick search I was able to find it in `/home/root/.config/chessmarkable/savestates.yml`

LinusCDE:
Thanks. I think I failed copy&paste there. It is indeed intended to be in `/home/root/.config/chessmarkable/`.

Eeems (approved):
Works for me on my rM1

PR #363, opened 2021-06-04
Update rm2fb to v0.0.4

matteodelabre:
This update includes a change in the server code to set the server process name to 'rm2fb-server' (instead of the default 'xochitl'). See also #359 and rmkit-dev/rmkit#127. I also removed Toltec’s patch to set `argv[0]` since it’s included in upstream now.

Test plan: Started the 'rm2fb' service, retrieved the server PID, and made sure that both `/proc/PID/cmdline` and `/proc/PID/stat` contained 'rm2fb-server' instead of 'xochitl'.


Eeems:
Installed fine for me on rM1 with no bad side effects.

JanLuca:
Installed and works fine for me on rM2. Thank you for updating the package!

raisjn (approved):
(based on jan's testing)

matteodelabre:
Thanks all for providing additional testing.

PR #366, opened 2021-06-07
Update packages for release 2.6 and onwards

### LINKED ISSUE #322: Support for system update 2.6

matteodelabre:
With release 2.6 having reached most users, I think it makes sense to rebuild our packages to use the 2.x toolchain. This will make Toltec work with updates 2.6 and onwards (fixes #322), at the price of **breaking compatibility with system updates 2.5 and earlier.**

Testing plan: Having update 2.7.1 installed, I went through basic checks for each package on rM2 to make sure that each app at least starts up and seems to react to user input.

### DIFF
--- a/package/appmarkable/package
+++ b/package/appmarkable/package
@@ -5,7 +5,7 @@
 pkgnames=(appmarkable)
 pkgdesc="Front-end for apps that do not have a graphical user interface"
 url="https://github.com/LinusCDE/appmarkable"
-pkgver=0.0.0-10
+pkgver=0.0.0-11
 timestamp=2021-03-10T18:36Z
 section="devel"
 maintainer="Linus K. <linus@cosmos-ink.net>"
@@ -13,7 +13,7 @@ license=MIT
 installdepends=(display)
 flags=(patch_rm2fb)
 
-image=rust:v1.4
+image=rust:v2.1
 source=(https://github.com/LinusCDE/appmarkable/archive/c44ee87ea2b1f1e41c9592476c076150c9a1acf4.zip)
 sha256sums=(76e151aeae0f18b206dd3c6258bf74bcb5256ee2f803e1ed2073278831158f60)
 
--- a/package/bash-completion/package
+++ b/package/bash-completion/package
@@ -5,7 +5,7 @@
 pkgnames=(bash-completion)
 pkgdesc="Programmable completion functions for bash"
 url=https://github.com/scop/bash-completion
-pkgver=2.11-1
+pkgver=2.11-2
 timestamp=2020-07-25T00:00Z
 section="utils"
 maintainer="Linus K. <linus@cosmos-ink.net>"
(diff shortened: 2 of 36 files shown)


matteodelabre:
The only issue I found for now is that @Eeems’s Oxide has some missing icons (top-left menu button and top-right power button). Other than that, the launcher works well.

Another issue that is minor in the context of this PR is that @fenollp’s whiteboard-hypercard does not build with the new image, because it features an updated Rust nightly toolchain that is no longer compatible with one of its crates. Since it does not use Qt libs, it should work without needing an update to the newer toolchain for now.

Eeems:
> The only issue I found for now is that @Eeems’s Oxide has some missing icons (top-left menu button and top-right power button). Other than that, the launcher works well.

I'm including a font to drive those. Are the files not being pulled in?

matteodelabre:
The missing icons bug is already tracked in issue https://github.com/Eeems/oxide/issues/194 and PR https://github.com/Eeems/oxide/pull/195.

matteodelabre:
Added a patch for Oxide (taken from Eeems/oxide#195) that fixes the missing icons bug.

ralsei:
I can confirm the following applications work on 2.8 on rM1:
- All parts of Oxide (except `decay`, not because it doesn't work, but because I don't use it)
- ZShelf
- Wikipedia
- Yaft
- nao
- Plato
so at first glance it doesn't look like 2.8 bumped the Qt version in a way that matters.

Eeems:
There have been several users who have used this branch on 2.6+, what is currently blocking this from merging?

matteodelabre:
To me this PR is clear for merging. @raisjn and you did mention the need for adding a check that the user is actually running version ⩾ 2.6 before upgrading. This can also be done in a separate PR, since this is going to testing and not directly to stable.

Eeems (approved):

Issue #368, opened 2021-06-18
Installation not working / Invalid checksum for the local wget binary


clemmrich:
When trying to install, I get the following error messages:

root@reMarkable:~# echo "46f556b06f5624b48e974ae040b6213828eff6aa2cc78618a4d8961a27cdc8b3  bootstrap" | sha256sum -c && bash bootstrap
bootstrap: OK
INFO:  Fetching secure wget
ERROR: Invalid checksum for the local wget binary
ERROR: Unexpected error on line bootstrap:105 in function wget-bootstrap
ERROR: This script failed to install. If you can't solve the above
       issue yourself, please report it at:

          https://github.com/toltec-dev/toltec/issues

       (Please also include these error logs to help solving the
       problem faster. Thank you!)

matteodelabre:
Hi @clemmrich, thanks for the report. Did you try running the bootstrap again, just in case this was a one-time network error?

clemmrich:
Yes, I tried it 3 times...

matteodelabre:
Can you please try to download the wget binary manually to see if this is an issue with the bootstrap script or with the network?

```console
$ wget https://toltec-dev.org/thirdparty/bin/wget-v1.21.1
$ sha256sum wget-v1.21.1
```

If this works, what checksum does the last command report? It should be `8798fcdabbe560722a02f95b30385926e4452e2c98c15c2c217583eaa0db30fc`. If it doesn’t match, can you inspect the type of the downloaded file by running the `file wget-v1.21.1` command? The file type should start with `ELF 32-bit LSB executable`.

clemmrich:
Manual download is ok, sha256sum does match. 
The bootstrap script does not seem to dwonload anything, the error message is displayed almost instantly without a delay for a download…

 

 

Von: Mattéo Delabre ***@***.***> 
Gesendet: Freitag, 18. Juni 2021 12:07
An: toltec-dev/toltec ***@***.***>
Cc: clemmrich ***@***.***>; Mention ***@***.***>
Betreff: Re: [toltec-dev/toltec] Installation not working / Invalid checksum for the local wget binary (#368)

 

Can you please try to download the wget binary manually to see if this is an issue with the bootstrap script or with the network?

$ wget https://toltec-dev.org/thirdparty/bin/wget-v1.21.1
 
$ sha256sum wget-v1.21.1

[...]

PR #369, opened 2021-06-18
Hotfix: Fix local wget download issue in stable

matteodelabre:
The bootstrap script that’s currently in stable creates an empty `/home/root/.local/bin/wget` file before downloading its contents using `wget`. If a user happens to have `/home/root/.local/bin` in their PATH before the `/usr/bin` folder, then the script will try to use the empty `/home/root/.local/bin/wget` binary to download itself! (See #368.)

This issue is already fixed in testing, so this PR is just a backport of this fix to the stable branch.

Test plan: From a clean system, with no file in `/home/root/.local/bin/wget`, set your `PATH` to include the `/home/root/.local/bin` folder before the `/usr/bin` folder. With the old bootstrap script, this should result in a checksum mismatch error. With this PR’s bootstrap script, the install should complete successfully.

I tested this on my rM2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To automatically install Opkg, Entware and Toltec, run the bootstrap script in a
 
 ```sh
 $ wget http://toltec-dev.org/bootstrap
-$ echo "46f556b06f5624b48e974ae040b6213828eff6aa2cc78618a4d8961a27cdc8b3  bootstrap" | sha256sum -c && bash bootstrap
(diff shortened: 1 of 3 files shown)


raisjn (approved):

matteodelabre:
This will be the first “live” test of the logic that triggers a rebuild of the [website](https://github.com/toltec-dev/web) whenever we merge to stable.

matteodelabre:
> This will be the first “live” test of the logic that triggers a rebuild of the [website](https://github.com/toltec-dev/web) whenever we merge to stable.

Correction: Actually not, since this hotfix does not merge the updated workflows to stable, this merge will just use the old way of building the website from the main repo.

PR #372, opened 2021-06-20
Fix toltecctl set-path command if .bashrc does not exist

matteodelabre:
The `toltecctl set-path` command is used to make sure that the user’s shell PATH is properly configured to see Toltec binaries. It first scans the existing .bashrc to remove old Toltec-related PATH definitions, and then appends its own definitions.

In the current implementation, if there’s no existing /home/root/.bashrc file, the script will exit with an error. This is a problem because `set-path` is called as part of the bootstrap script, so if someone runs bootstrap after doing a factory reset (which removes the .bashrc file), the bootstrap will fail.

This PR changes the implementation to skip cleaning the .bashrc if it doesn’t exist. It also fixes two wrong messages in the bootstrap script that refer to the `toltecctl enable` command, that is actually called `toltecctl reenable`.

Test plan: Remove (or rename) your .bashrc and upgrade toltec-bootstrap to the new version. Since `set-path` is called automatically when the package is upgraded, this should recreate a new .bashrc containing Toltec-related PATH definitions.

I tested this on my rM2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To automatically install Opkg, Entware and Toltec, run the bootstrap script in a
 
 ```sh
 $ wget http://toltec-dev.org/bootstrap
-$ echo "47f208632d93f368b267e66413150aa7c7da43bac0431f893b1fd2ea146ceced  bootstrap" | sha256sum -c && bash bootstrap
+$ echo "c930e94a4145c31e63ac8c66b1b0f5a192a91f31946d758bfc1470338cb64bae  bootstrap" | sha256sum -c && bash bootstrap
 ```
 
 > **Warning:**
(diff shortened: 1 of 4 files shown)


LinusCDE (approved):
Tested on rm 1 and works great!

```shell
reMarkable: ~pkgs/ mv ~/.bashrc{,.bak}
reMarkable: ~pkgs/ opkg install toltec-bootstrap_0.0.7-1_rmall.ipk 
Upgrading toltec-bootstrap on root from 0.0.6-1 to 0.0.7-1...
Configuring toltec-bootstrap.
Please restart your SSH session or run 'exec bash -l' to use Toltec
reMarkable: ~pkgs/ cat ~/.bashrc
# Added by Toltec bootstrap (do not modify!)
PATH="/opt/bin:/opt/sbin:/home/root/.local/bin:$PATH"
# End of Toltec bootstrap additions
reMarkable: ~pkgs/ 
```

matteodelabre:
Thanks for testing!

PR #373, opened 2021-06-21
Update chessmarkable to 0.7.1-1

LinusCDE:
Hi. Here is a small user friendliness update (more in [the release notes](https://github.com/LinusCDE/chessmarkable/releases/tag/0.7.1-1)).

Tested it on my rM 2. And I don't see a reason why it shouldn't work on the rM 1.

Testplan:

Just check that it runs. A notable change is that when a piece is selected, clicking a non-highlighted one will select that one instead of showing an error because of an invalid move.


matteodelabre (approved):
Looks good to me. Tested on rM2 following your test plan.

PR #375, opened 2021-06-22
Fix typo in bash-completion

LinusCDE:
Just spotted that I made a typo/grammar mistake in the `configure()` part of bash-completion.


matteodelabre (approved):
Good catch! Looks good to me.

PR #383, opened 2021-07-11
Create remarkable-stylus package

Eeems:


package/remarkable-stylus/package
```diff
@@ -0,0 +1,25 @@
...
+
+pkgnames=(remarkable-stylus)
+pkgver=0.0.3-1
+pkgdesc="hack to get the button of a lamy pen working"
+timestamp=2020-11-19T20:07:29Z
+maintainer="Eeems <eeems@eeems.email>"
+license=LGPL-3.0
+url=https://github.com/ddvk/remarkable-stylus
+section="admin"
+installdepends=(display "tarnish=$pkgver")
```
raisjn: what's this about?

Eeems: Leftover from the package I copied in. That's part of why this is still a draft :P

Eeems:
Looks like we just need someone with a lamy pen to test this on both a rM1 and rM2.

Eeems:
From initial testing on a rM1, it's been reported that this works through a launcher (Oxide), but not when running the systemd unit directly. We likely need to add another environment file for the systemd unit specifically.

Eeems:
> 
> 
> From initial testing on a rM1, it's been reported that this works through a launcher (Oxide), but not when running the systemd unit directly. We likely need to add another environment file for the systemd unit specifically.

This should be resolved by adding `systemctl daemon-reload` to the `xochitl`'s package configure script.

Eeems:
Oh, and this will resolve https://github.com/Eeems/oxide/issues/136 when merged.

matteodelabre (approved):

PR #384, opened 2021-07-12
Update innernet to 1.4.0

Eeems:


Eeems:
@LinusCDE or @matteodelabre could you review this?

matteodelabre:
@Eeems Do you have experience with innernet? I tried to follow the official docs to set up a local network between my computer and the tablet, but I can’t seem to add a new peer in the network after creating it. I’m not sure whether I should set up wireguard first or if innernet is supposed to take care of everything?

Eeems:
Internet is suppose to take care of everything, but you can run into trouble with NAT holepunching not being robust enough. I've had no issues with it when I've used it on my rM1, but I also modify the file to make sure it's using an internal IP to connect to my server instead of an external one.

matteodelabre:
Finally managed to make it work. Thanks for the help!

matteodelabre (approved):

PR #385, opened 2021-07-13
Update ddvk-hacks to latest

Eeems:
Also drop support for versions older than 2.6


Eeems:
I've only tested on a rM1 with 2.8.0.98. This will require testing on a rM2 as I've split the packages out to save space on each device.

matteodelabre (approved):
Tested on rM2 2.8.0.98, works perfectly! Very good idea to split the package in two.

PR #386, opened 2021-07-14
Fix xochtil chroot launching

Eeems:


matteodelabre (approved):
This fixes an issue where Xochitl sometimes doesn’t display anything on screen on rM2.

PR #392, opened 2021-07-22
Update chessmarkable to 0.8.0-1

LinusCDE:
This adds a new [PGN](https://en.wikipedia.org/wiki/Portable_Game_Notation) Viewer (can replay PGN files [like these](https://www.pgnmentor.com/files.html), when placed in `~/.config/chessmarkable/pgn`).

The release shouldn't change anything in terms of stability. @rmadhwal added this feature in a PR and initially tested it on his rM 2. I later checked it on my rM 1 and rM 2 as well.

Test plan:
 - Either just check that it runs
 - Or just test some pgn file as explained in the first paragraph if you want to

### DIFF
--- a/package/chessmarkable/package
+++ b/package/chessmarkable/package
@@ -5,8 +5,8 @@
 pkgnames=(chessmarkable)
 pkgdesc="Chess game"
 url=https://github.com/LinusCDE/chessmarkable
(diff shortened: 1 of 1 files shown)


raisjn (approved):
it works!

PR #395, opened 2021-07-24
netsurf: Improve rM2 compatibility

matteodelabre:
This PR adds a missing dependency of `netsurf` on the `display` package. This dependency ensures that rM2 users have the rm2fb server installed before they start netsurf. This diff also adds the `patch_rm2fb` flag to the recipe, so that netsurf can be started from the command line on rM2 without prefixing it with `rm2fb-client`.

### DIFF
--- a/package/netsurf/package
+++ b/package/netsurf/package
@@ -5,7 +5,7 @@
 pkgnames=(netsurf)
 pkgdesc="Lightweight web browser"
 url=https://github.com/alex0809/netsurf-reMarkable
-pkgver=0.4.0-1
+pkgver=0.4.0-2
 timestamp=2021-05-31T11:15+00:00
 maintainer="Alex Friesenhahn <rm-dev@alexfriesenhahn.de>"
 archs=(rmall)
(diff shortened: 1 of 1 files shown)


package/netsurf/package
```diff
@@ -43,6 +45,7 @@ sha256sums=(
 )
 
 build() {
+    ln -s /usr/bin/which /bin
```
Eeems, line 48: I'm assuming one of the scripts is trying to run `which` from a specific location that is incorrect?

matteodelabre: Yes, and `which` was moved from `/bin` to `/usr/bin` in Debian Bullseye (compare https://packages.debian.org/buster/amd64/debianutils/filelist with https://packages.debian.org/bullseye/amd64/debianutils/filelist, see https://salsa.debian.org/debian/debianutils/-/commit/1ceeff62723ecff6f843a68b56080b991a7d1c48). This seemed like the fastest way to fix the build to work with our toolchain v2 without delving too much into the specific build system of netsurf.

raisjn (approved):

PR #399, opened 2021-07-26
Build dotnet instead of using binary packages

Eeems:


matteodelabre:
The issue is that I can’t find in their docs how to make the build system do a cross-compiling build (we’d need to specify the sysroot and the compiler to use somewhere). Right now it’s looking for the x86 clang compiler, and failing since it’s not installed, but even if we install it the build is going to target x86 and not ARM.

Eeems:
Time to open an issue on their repo asking for help?

matteodelabre:
Yes, why not! I’ll continue investigating tomorrow.

Eeems:
@matteodelabre did you ever open an issue? I'd like to have it linked here so I can follow it if you did.

Eeems:
Upon looking into this again, I don't think it's worth implementing, due to how much we have to still depend on MS to host unmodified versions of the source files that their scripts will download to build. You have to download the installer repo, and then it will download sources, and then use them to download many more sources and then build those into the final result. The amount of space and time required does seem to be more than is worth it for our scenario.

PR #402, opened 2021-07-29
Improve Opkg completion

matteodelabre:
This PR builds upon the work of @LinusCDE on Bash completions for Opkg and adds some missing features.

* Add completion of flag names and values.
* Only suggest installed packages when completing `opkg remove`.
* And completion of other subcommands involving package names (depends, download, info, what.*, configure, files, status, upgrade, flag)
* Add completion of search and compare-versions operators.

This makes the completion function fairly complete as far as I can see.

A future area of work will be to cache the list of packages, currently it is re-generated for each call to the completion function, which takes some noticeable time.

### DIFF
--- a/package/toltec-completion/_opkg
+++ b/package/toltec-completion/_opkg
@@ -1,58 +1,134 @@
-#!/bin/bash
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+# shellcheck disable=SC2016,SC2207
 #
-# shellcheck disable=SC2016,SC2199,SC2207
-#
-# Some spellchecks were ignored to favor the style
+# Some shellchecks were ignored to favor the style
 # files in the bash-completion project are layed out.
 #
 # SC2016: Using '' after compgen commands is intentional
 # to make compgen evaluate the expression later.
-#
 
-# Currently pretty basic. Caveats / TODOs for anyone who has time:
-# - Does not support flag
-# - Only completes packages for install and remove commands
-# - The install command only supports very basic file (.ipkg) completion
-# - Package completion should probably get cached for speedup
-#
-# If this file gets properly supports opkg, we should probably
+# Once this file gets proper support for opkg, we should probably
 # consider backporting this to https://github.com/scop/bash-completion
+
+set +o posix
 
 _opkg() {
     local cur prev words cword split
     _init_completion -s || return
 
-    #echo "Cur: $cur" # Contains the partial word
-    #echo "Prev: $prev" # The previous word (the command if no previous one)
-    #echo "Words: $words" # Array of all words? (like $@ ?)
-    #echo "CWord: $cword" # Number for current word (first word = 1, ...)
-    #echo "Split: $split" # ???
-
-    # Since ALL words in $COMPREPLY get suggested (regardless whether
-    # the user partiall typed one) compreply handles only returning
-    # the relevant ones
+    #echo "cur: $cur" # Contains the partial word
+    #echo "prev: $prev" # The previous word (the command if no previous one)
+    #echo "words: $words" # Array of all words
+    #echo "cword: $cword" # Index of current word (first word = 1, ...)
 
     if [[ $cword -eq 1 ]]; then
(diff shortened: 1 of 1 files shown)


LinusCDE (approved):
Tested on a rM 2 and works really well for me! :+1:

---

I pondered some options regarding caching the package names.

I wanted some kind of binary table (btree, avl tree) to sped up the search instead our naive search.

Some options i pondered:
 - A fitting tool for that purpose: Maybe gnomes "tracker" may do that, but we don't have the package and I don't know how big the dependency footprint would be
 - A custom tool that uses some btree/avltree lib that can save the result to a cache file
   - Creating such a tool in e.g. rust would probably make maintainability hard

I finally settled on using the sqlite cli tool for that. Small footprint, indexing support to speed up search and usable from a bashscript.

Below is a basic script it did that uses sqlite for caching.

It needs about 300ms to create the cache and then about 33ms to open and query the cache db for packages. It is somewhat incomplete, still prints some debug/timing info and is subject to sql-injection. Feel free to do with it as you please.

```bash
#!/bin/bash

if ! which sqlite3 >/dev/null 2>&1; then
  echo "Error: Please install sqlite3-cli !" >&2
  exit 1
fi

DB_FILE="$HOME/.cache/fast_pkg_search.sqlite"

# The does the follwing steps:
#
# - Dump potentially existing sqlite db
# - Create new sqlite db
# - Create `package_names` table
# - Insert names from "opkg list" into that table
#
# Takes about 300ms on a rM2
regenerate_cache() {
  [ -e "$DB_FILE" ] && rm "$DB_FILE"
  local packages_db_values
  package_db_values=$(opkg list | awk '{ if (NR!=1) printf ","; printf "('"'"'%s\'"'"')", $1 ;}')

  sqlite3 "$DB_FILE" '
  CREATE TABLE `package_names` (`name` VARCHAR(64) PRIMARY KEY NOT NULL);
  INSERT INTO `package_names` (`name`) VALUES '"$package_db_values"';
  '
}

complete_partial_package() {
  if [ ! -e "$DB_FILE" ]; then
    echo "Generating cache..."
    time regenerate_cache
  fi
  # TODO: Other conditions that can expire the cache
  # e.g. checking for changed file sizes/dates in /opt/var/opkg-lists

  # Query existing database for packages starting with $1 (= WHERE `name` LIKE '$1%')
  sqlite3 "$DB_FILE" 'SELECT `name` FROM `package_names` WHERE `name` LIKE '"'$1%'"';'
}

#regenerate_cache
complete_partial_package $@
```

matteodelabre:
Very interesting approach @LinusCDE, thanks for sharing it! And thanks also for your review. It would be useful to benchmark your code snippet against more naive cache approaches to make sure it’s worth the effort.

LinusCDE:
image

Did some testing. Doing this in bash is pretty limited since i didn't find a good way to benchmark without spawning a new bash to be tested. I think that about 105ms in both cases is just bash starting up and everything else but the completion. (The echo output seems to take up at most only 1ms extra.)

Even though these tests aren't pretty accurate, it seems that the SQLite aproach is ever so slightly slower. I would guess that the OS can easily cache the opkg-lists files into ram and those are already pretty compact. awk checking the output of all packages seems to be a bit faster than loading the sqlite database. Even if the query should be faster in theory (the dataset is probably too small to make a big enough difference).

So I don't think that we could ever notice the speed increase of the autocomplete and thanks to caching, a relatively fast cpu and a not that big dataset there is no sense in speeding up this operation with a sqlite database.

---

On another note: When testing the caching, I somehow forgot to mention that the version of the package needs to be bumped @matteodelabre . Do you just want to push this directly to the testing branch? I find an increase of the minor version fitting for this (toltec-completion `0.2.0-1` -> `0.3.0-1`).

matteodelabre:
Thanks for doing the benchmarking. It’s unfortunate that this doesn’t yield better results, we’ll keep looking for better alternatives!

> On another note: When testing the caching, I somehow forgot to mention that the version of the package needs to be bumped @matteodelabre . Do you just want to push this directly to the testing branch? I find an increase of the minor version fitting for this (toltec-completion `0.2.0-1` -> `0.3.0-1`).

Sorry I’ve missed that too. I’ve opened PR #403 to that effect (direct pushes to testing being disabled).

PR #404, opened 2021-07-30
Add JQ package

ashermancinelli:
Version 1.6 is released, but it's not pre-autoconf'ed like it should be so I stuck with 1.6.

### DIFF
--- a/package/jq/package
+++ b/package/jq/package
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(jq)
+pkgdesc="JSON Manipulation Utility"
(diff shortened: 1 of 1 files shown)


Eeems:
Hey @ashermancinelli, thanks for submitting a PR to toltec to add a package! Unfortunatly this package is of an older version than the `jq` package in entware, so it wont install unless you specify the specific version. Did you need 1.5 instead of the 1.6 version supplied in the entware repository?

ashermancinelli:
Ah! My apologies. Where can I find this package? I couldn't find it in toltec or with `opkg search jq`.

Eeems:
When toltec is installed, it should also install the entware repository. You should be able to install it by just running `opkg install jq`. You can find the full list of packages here: https://bin.entware.net/armv7sf-k3.2/

PR #405, opened 2021-07-31
fuse: Fix on-demand loading after install on rM1

matteodelabre:
In the current state of the fuse package, the module will not get autoloaded if you try to use it right after installing. You can see this
on rM1 by installing gocryptfs and trying to decrypt a directory (without rebooting after installing the package).

This PR fixes that by forcing the creation of `/dev/fuse` that is used for on-demand loading of the fuse kernel module. If you install this new version of the package without having fuse installed before, the module should get autoloaded without needing a restart of the device.

### DIFF
--- a/package/fuse/package
+++ b/package/fuse/package
@@ -6,7 +6,7 @@ archs=(rm1 rm2)
 pkgnames=(fuse)
 pkgdesc="FUSE (Filesystem in Userspace) Kernel Module"
 url=https://github.com/libfuse/libfuse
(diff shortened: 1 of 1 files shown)


Eeems (approved):
Looks good to me. Unable to actually test at the moment. I'm assuming that you've tested on both rM1 and rM2?

matteodelabre:
Yes, I’ve tested it on rM1. Did not test on rM2 since fuse is a no-op package there anyway.

PR #408, opened 2021-08-07
Update micro to 2.0.10

Eeems:


LinusCDE (approved):
Works fine on the rM 2

PR #410, opened 2021-08-13
display: Update to release v0.0.6

matteodelabre:
This release adds support for the latest OS version, 2.9.0.210.

I also changed the version of this package to follow upstream’s versioning scheme, which makes it easier to see that our package is up to date. Our version was previously at 1.0.1 while upstream was at 0.0.6, with 1.0.1 being the lib’s version number. Because this change makes the new version number lower than the previous one, I had to increase the package’s [epoch](https://www.debian.org/doc/debian-policy/ch-controlfields.html#epochs-should-be-used-sparingly).

Tested on rM2 version 2.9.0.210.

### DIFF
--- a/package/display/package
+++ b/package/display/package
@@ -4,23 +4,26 @@
 
 archs=(rm1 rm2)
 pkgnames=(display rm2fb-client)
-timestamp=2021-02-21T01:41+00:00
+timestamp=2021-08-13T13:31:03Z
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
 url="https://github.com/ddvk/remarkable2-framebuffer"
-pkgver=1.0.1-7
+pkgver=1:0.0.6-1
+_release="${pkgver%-*}"
+_release="v${_release#*:}"
+_libver=1.0.1
 section="devel"
 
 image=qt:v2.1
 source=(
-    https://github.com/ddvk/remarkable2-framebuffer/archive/v0.0.5.zip
+    "https://github.com/ddvk/remarkable2-framebuffer/archive/$_release.zip"
     rm2fb.service
     rm2fb-client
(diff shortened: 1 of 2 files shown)


matteodelabre:
GitHub disallows colons in filenames, but we use colons to separate epoch numbers the rest of version identifiers, and those identifiers get included as-is in the generated package filenames. Since the package filename isn’t really relevant (it just needs to be unique), I added a small change to use underscores instead of colons when representing epochs in filenames.

Eeems:
> 
> 
> GitHub disallows colons in filenames, but we use colons to separate epoch numbers the rest of version identifiers, and those identifiers get included as-is in the generated package filenames. Since the package filename isn’t really relevant (it just needs to be unique), I added a small change to use underscores instead of colons when representing epochs in filenames.

We could also just tar the results to avoid having to work around it.

matteodelabre:
That would be a valid solution too, but I just went for the simplest option here. As an added bonus, Bash autocomplete on the device behaves a bit weird with files containing colons, so changing the character to an underscore makes it a bit easier to install packages by file name.

matteodelabre:
Marking this as draft until ddvk/remarkable2-framebuffer#68 is sorted out.

matteodelabre:
Update: The issue was not directly caused by rm2fb but by [remarkable-autoinstall](https://github.com/ddvk/remarkable-autoinstall) being out of date. @Eeems is working on a fix: ddvk/remarkable-autoinstall#18. I think we can go forward with this PR.

Eeems (approved):
lgtm, but I can't test.

matteodelabre:
Thanks for the review! I’m going to merge this since I’ve confirmed it to work on 2.9, and it’s unlikely to break other versions since I just added new addresses.

PR #412, opened 2021-08-14
scripts/toltec/bash.py: Disable security_ops

alistair23:
Specify security_ops "label=disable" to avoid permission issues with SE
Linux.

See the Docker documentation
https://docs.docker.com/engine/reference/run/#security-configuration for
more details.

Signed-off-by: Alistair Francis <alistair@alistair23.me>

### DIFF
--- a/scripts/toltec/bash.py
+++ b/scripts/toltec/bash.py
@@ -402,6 +402,7 @@ def run_script_in_container(
                 )
             ),
         ],
+        security_opt=['label=disable'],
         detach=True,
     )
 
--- a/scripts/toltec/bash.py
+++ b/scripts/toltec/bash.py
@@ -402,7 +402,7 @@ def run_script_in_container(
                 )
             ),
         ],
-        security_opt=['label=disable'],
+        security_opt=["label=disable"],
         detach=True,
     )
 


matteodelabre:
Hi! Did you get those permission issues when building a specific package? What was the exact issue? My impression is that the build process should not require special permissions, so I’d prefer fixing the underlying issue rather than disabling the Docker security measures, if possible.

alistair23:
I get an error on every package, the error changes per package though. This is what I see for bash-completion for example:

```
[    INFO] toltec.builder: bash-completion [rmall]: Fetching source files
[    INFO] toltec.builder: bash-completion [rmall]: Building artifacts
[    INFO] toltec.builder: bash-completion [rmall]: Only showing up to 50 lines of context. Use --verbose for the full output.
[   ERROR] toltec.builder: bash-completion [rmall]: environment: line 17: ./configure: Permission denied
[   ERROR] toltec.builder: bash-completion [rmall]: build() failed
Traceback (most recent call last):
  File "toltec/./scripts/repo_build.py", line 62, in <module>
    builder.make(generic_recipe, missing[generic_recipe.name])
  File "toltec/scripts/toltec/builder.py", line 143, in make
    if not self._make_arch(
  File "toltec/scripts/toltec/builder.py", line 168, in _make_arch
    self._build(recipe, src_dir)
  File "toltec/scripts/toltec/builder.py", line 357, in _build
    self._print_logs(logs, "build()")
  File "toltec/scripts/toltec/builder.py", line 689, in _print_logs
    raise err
  File "toltec/scripts/toltec/builder.py", line 670, in _print_logs
    for line in logs:
  File "toltec/scripts/toltec/bash.py", line 416, in run_script_in_container
    raise ScriptError(f"Script exited with code {result['StatusCode']}")
toltec.bash.ScriptError: Script exited with code 126
make: *** [Makefile:51: repo-local] Error 1
```

I found details on the change here: https://discussion.fedoraproject.org/t/docker-volume-permission-denied/1948

matteodelabre:
Thanks for the details! I have to admit I don’t know a lot about SELinux. From what I understand, containers are not allowed to interact with files from the host by default, and this includes bind mounts. The other option would be to use the `:Z` mount flag that relabels the mounted files to be accessible from the container. But for now it seems like your solution is the simplest, if we’re ready to accept that this will disable SELinux security measures. We might want to revisit the security options in the future.

matteodelabre (approved):

PR #420, opened 2021-08-29
Bump innernet to v1.4.1

plan5:
There are minor bugfixes in this package. Build works as expected and installed binary runs.

Autocompletions are not included as I don't really know how to approach this.


matteodelabre (approved):
Tested on rM2, seems to be working fine. Thanks!

PR #421, opened 2021-08-29
vnsee: Update to v0.4.1

matteodelabre:
This release fixes support for reMarkable 1 version 2.9.


Eeems:
Seems to install fine, but I can't test due to getting the following when trying to go through the [generic workaround instructions:](https://github.com/matteodelabre/vnsee/blob/master/docs/second-screen/linux-x11.md#generic-workaround)
image

I'm guessing it's because there are no nvidia instructions?

matteodelabre:
@Eeems Note that it’s not necessary to create a virtual output to test VNSee, you can also just mirror one of your existing outputs (i.e., replace OUTPUTNAME with the name of an active output [in this command](https://github.com/matteodelabre/vnsee/blob/master/docs/second-screen/linux-x11.md#start-the-vnc-server)). If the output you use is larger than the rM screen, it will be cropped to fit on screen.

LinusCDE (approved):
Tested on both devices.

| rM 1 | rM 2 |
| --- | --- |
| rm1 | rm2 |

Btw I love that it crops the image! Makes testing so much easier, since I can just test against a bog standard x11vnc server without adjusting any displays.

matteodelabre:
Thanks for your review Linus!

> Btw I love that it crops the image! Makes testing so much easier, since I can just test against a bog standard x11vnc server without adjusting any displays.

Glad you like it, I hope to add more options for controlling the image in the future (like rotating, panning, and scaling).

PR #423, opened 2021-08-29
bootstrap: Fix TLS, check if .entware exists, and improve error messages

### LINKED ISSUE #409: bootstrap can overwrite existing installs when run after a system upgrade
The `bootstrap` script is supposed to halt if it detects an existing Toltec install, to avoid any data loss. [The current implementation](https://github.com/toltec-dev/toltec/blob/ad682b5c60fd6467e26e7c50157626386ab8d050/scripts/bootstrap/bootstrap#L72) only checks if there’s a non-empty `/opt` folder. However, after upgrading the system, the `/opt` folder is no longer mounted even though the install still exists in `/home/root/.entware`. So if one runs the `bootstrap` script right after upgrading, it may overwrite their existing install.

### LINKED ISSUE #422: bootstrap fails on line 96 while fetching standalone wget due to a TLS error
Trying to install Totec

INFO:  Fetching secure wget
ERROR: Unexpected error on line bootstrap:96 in function wget-bootstrap
ERROR: This script failed to install. If you can't solve the above
       issue yourself, please report it at:
       
           <https://github.com/toltec-dev/toltec/issues>
       
       (Please also include these error logs to help solving the
       problem faster. Thank you!)
bootstrap: line 60: clean-path: command not found

Linux newbie -- any idea how to clear this one?

matteodelabre:
This PR contains several fixes for the bootstrap script.

* If the standalone `wget` binary cannot be fetched, a more helpful error message is printed instead of the catch-all error message (see #422).
* The standalone `wget` binary is fetched over HTTP instead of HTTPS, since the implementation of TLS in Busybox’s wget causes issues with the latest version of nginx. This does not reduce security since the checksum is still verified after fetching the binary (fixes #422, I believe).
* Abort installation if `/home/root/.entware` already exists. Previously, only the existence of `/opt` was checked, leading to #409.

Test plan (on rM1 and rM2):

* Try to run the script normally. Issue #422 should be fixed and the install should proceed.
* Try to run the script while not being connected to Wi-Fi. Expected output:

```sh
INFO:  Fetching secure wget
wget: bad address 'toltec-dev.org'
ERROR: Could not fetch wget, make sure you have a stable Wi-Fi connection
```

* Try to run it with `127.0.0.1 toltec-dev.org` in `/etc/hosts`. Expected output:

```sh
INFO:  Fetching secure wget
wget: can't connect to remote host (127.0.0.1): Connection refused
ERROR: Could not fetch wget, make sure you have a stable Wi-Fi connection
```

* Starting from a complete Toltec install, unmount /opt (`systemctl stop opt.mount`) and remove the /opt folder. Try to run the script. Expected output:

```sh
INFO:  Toltec is already installed or partially installed
INFO:  To re-enable Toltec after a system upgrade, run 'toltecctl reenable'
INFO:  To reinstall Toltec, run 'toltecctl uninstall' first
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To automatically install Opkg, Entware and Toltec, run the bootstrap script in a
 
 ```sh
 $ wget http://toltec-dev.org/bootstrap
-$ echo "9195122984700c76ccdc58e25d09d0fca486324e8fc55ba781f6e1b812cc186c  bootstrap" | sha256sum -c && bash bootstrap
+$ echo "2d1233271e0cc8232e86827bcb37ab2a44be2c5675cd15f32952614916ae246a  bootstrap" | sha256sum -c && bash bootstrap
 ```
 
 > **Warning:**
--- a/scripts/bootstrap/bootstrap
+++ b/scripts/bootstrap/bootstrap
@@ -33,13 +33,7 @@ toltec_remote="${toltec_remote:-https://toltec-dev.org/$toltec_branch/rmall}"
 
 # Remove all temporary files
 cleanup() {
-    if [[ -e $wget_path ]]; then
-        rm "$wget_path"
-    fi
-
-    if [[ -e $opkg_path ]]; then
-        rm "$opkg_path"
-    fi
+    rm -f "$wget_path" "$opkg_path"
 }
 
 # Remove unfinished installation after unexpected error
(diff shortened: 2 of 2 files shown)


raisjn (approved):
accept to unblock

matteodelabre:
Thanks!

PR #425, opened 2021-08-30
Update fbink to v1.24.0

Eeems:


matteodelabre (approved):
Tested on rM1 and rM2.

PR #431, opened 2021-09-14
[display] add udev rules for adding device targets

raisjn:
this can be used by other services as a dependency to make sure they have their necessary input devices available before starting.  this is just one way of approaching the problem (that input devices aren't available when remux is starting). alternatively, can make remux/genie die if the input device is unrecognized

STATUS: untested.

### DIFF
--- a/package/display/package
+++ b/package/display/package
@@ -60,8 +60,9 @@ display() {
             ln -s "$libname" "$pkgdir"/opt/lib/"${libname%.*.*}"
             ln -s "$libname" "$pkgdir"/opt/lib/"${libname%.*}"
             install -D -m 644 -t "$pkgdir"/lib/systemd/system "$srcdir"/rm2fb.service
-            install -D -m 644 -t "$pkgdir"/lib/udev/rules.d/ "$srcdir"/15-systemd-input.rules
         fi
+
+        install -D -m 644 -t "$pkgdir"/lib/udev/rules.d/ "$srcdir"/15-systemd-input.rules
     }
 
     configure() {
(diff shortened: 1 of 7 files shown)


package/display/package
```diff
@@ -58,10 +60,12 @@ display() {
             ln -s "$libname" "$pkgdir"/opt/lib/"${libname%.*.*}"
             ln -s "$libname" "$pkgdir"/opt/lib/"${libname%.*}"
             install -D -m 644 -t "$pkgdir"/lib/systemd/system "$srcdir"/rm2fb.service
+            install -D -m 644 -t "$pkgdir"/lib/udev/rules.d/ "$srcdir"/15-systemd-input.rules
```
matteodelabre: This will only include the udev rule in the rM2 display package. Shouldn’t we include it in the rM1 package too?

package/display/package
```diff
@@ -114,6 +118,7 @@ rm2fb-client() {
             install -D -m 644 -t "$pkgdir"/etc/systemd/system/draft.service.d "$srcdir"/rm2fb-preload.conf
             install -D -m 644 -t "$pkgdir"/etc/systemd/system/remux.service.d "$srcdir"/rm2fb-preload.conf
             install -D -m 644 -t "$pkgdir"/etc/systemd/system/tarnish.service.d "$srcdir"/rm2fb-preload.conf
+            install -D -m 644 -t "$pkgdir"/lib/udev/rules.d/ "$srcdir"/15-systemd-input.rules
```
matteodelabre: Why add the udev rule to both the display and the rm2fb-client packages?

matteodelabre (changes requested):
Thanks for the PR! I think it would also make sense to add the changes that make the launcher services depend on the input targets in this PR—I can add them to this diff myself if you want.

package/display/package
raisjn: I was a bit confused about this. I thought one is for rm1 and other is for rm2. my intention is to include this rule in rm1/rm2 in display package

raisjn:
yes, please add them!

raisjn:
PS: this is addressing https://github.com/rmkit-dev/rmkit/issues/141 (adding crosslink so that post gets updated too)

matteodelabre:
Moved the udev rule to the `toltec-bootstrap` package as discussed on Discord.

matteodelabre (approved):
Tested on rM2.

okayzed:
i think remux needs to depend on `toltec-bootstrap` in order to get the udev rules installed and configured

raisjn:
i think remux needs to depend on toltec-bootstrap in order to get the udev rules installed and configured

PR #436, opened 2021-09-20
Update KOReader to v2021.09

Eeems:
No reMarkable specific fixes, but various fixes/improvments. You can see the full changelog here: https://github.com/koreader/koreader/releases/tag/v2021.09


matteodelabre (approved):
Thanks for the PR! Tested on rM2, working fine.

PR #439, opened 2021-09-25
Update ddvk-hacks to 24.04 (support for 2.9)

matteodelabre:
Tested on rM2.


Eeems (approved):
Works for me on my rM1

Issue #438, opened 2021-09-24
Launchers systemd services: Use alias instead of listing conflicts


matteodelabre:
In order to avoid having more than one launcher application started at boot, the existing launchers each declare a `Conflicts=` relationship with other launchers in their systemd units:

* https://github.com/Eeems/oxide/blob/master/assets/etc/systemd/system/tarnish.service
* https://github.com/toltec-dev/toltec/blob/stable/package/rmkit/remux.service
* https://github.com/toltec-dev/toltec/blob/stable/package/draft/draft.service

This approach works: if several launchers are enabled simultaneously, only one will actually start at boot (although it is not clear which one will start). It is not ideal because each launcher must list all other currently existing launchers in its service unit. If we were to add a new launcher, we’d need to update the service definitions of all other launchers.

I propose we use another mechanism: alias all launcher services to a common name. For example, add `Alias=launcher.service` in the `[Install]` section of each service. This has the effect of making `systemctl enable …` fail if any other launcher is already enabled, and therefore frees us from having to maintain those redundant `Conflicts=` lists. This approach is used on the desktop by display managers (gdm, lxdm, …); they all alias to `display-manager.service`.

Launcher authors (@Eeems @dixonary @raisjn), any thoughts on this? Would you want to have this upstream or only as patches in Toltec? I can open PRs to the respective repos if needed.

Eeems:
I like it. I'll happily accept a PR in my repo as well as here.

PR #441, opened 2021-09-25
koreader: Add service unit

### LINKED ISSUE #437: Include a Toltec-compatible 'koreader.service' unit with installation
### Current Workflow
Toltec is designed to work with a launcher and it does a wonderful job at automatically registering KOReader for use with oxide, remux or draft. That being said, users who would like to simply run KOReader at startup, must do a bit of extra work.

The [default KOReader systemd unit](https://github.com/koreader/koreader/blob/master/platform/remarkable/koreader.service) assumes a manual installation to the home folder, as evidenced by the line:

```vim
ExecStart=/home/root/koreader/koreader.sh
```

The current workflow is to manually change that path to `/opt/bin/koreader` and then copy the unit file to the `/etc/systemd/system` directory.

### Request
It would be ideal if Toltec could automatically create the `/etc/systemd/system/koreader.service` file that already complies with the Toltec directory structure. Leaving it up to the user whether to enable it or not. It would spare beginners some potential confusion and extra work — without interfering with the current function of launchers.

matteodelabre:
This PR adds a systemd service unit for starting KOReader at boot time (see #437). If KOReader fails to start, it falls back to Xochitl. It also updates the Xochitl service definition to alias to `launcher.service`; this is to prevent situations where both KOReader and Xochitl are enabled to start at boot time (see #438).


Eeems (changes requested):
The package should also have install/uninstall scripts handling letting the user know about the service, and properly stopping/disabling it on uninstall.

matteodelabre:
Good point, thanks! I added that.

Eeems (dismissed):
Changes look good, but I haven't tested.

matteodelabre:
Thanks for approving the PR—I had to push a new commit to fix the formatting, which auto-dismissed your review.

> Changes look good, but I haven't tested.

I’ve performed the following tests on my rM2: checked that enabling the service does start KOReader at boot, that it’s not possible to enable both KOReader and Xochitl at the same time, and that the configure() script suggests the appropriate command for enabling KOReader, depending on whether Xochitl is enabled.

I don’t think there is any rM1-specific things in this PR, so it should work fine on rM1.

Eeems:
It looks like while you can't enable both tarnish and koreader with this, you can start them both at the same time. Is that expected?

matteodelabre:
Yes, this is unfortunate, but I didn’t find a clean way to prevent this. This is not a regression from the `Conflicts=`-based approach though, since even conflicting services can be manually started at the same time.

Eeems (approved):

PR #443, opened 2021-09-25
Stable merge for week 38 of 2021

### LINKED ISSUE #409: bootstrap can overwrite existing installs when run after a system upgrade

### LINKED ISSUE #422: bootstrap fails on line 96 while fetching standalone wget due to a TLS error

matteodelabre:
Fixes:

* TLS issues with the bootstrap script (#422, #423)
* Bootstrap script could overwrite existing installs (#409, #423)

Updated packages:

* fbink, fbdepth, fbink-doom - 1.24.0-1 (#425)
* innernet - 1.4.1-2 (#420)
* koreader - 2021.09-1 (#436)
* remux - 0.1.9-5 (#431) (fixes 2.9 issues)
* toltec-bootstrap - 0.2.0-1 (#431)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,24 @@
+<!--
+
+Thank you for your interest in contributing to Toltec! Before submitting your
+pull request, please take a moment to read our contributing guidelines at
+<https://github.com/toltec-dev/toltec/blob/stable/docs/contributing.md>
+
+Most importantly, make sure to base your pull request on the **testing**
+branch (which is not the default branch). Pull requests to the stable
+branch cannot be accepted.
+
+If you’re proposing a new package please give us some details on:
+
+* What the package does
+* Whether you're the author of it
+* If the package was developed/tested for a specific reMarkable model
+
+If you’re updating an existing package, please give information on where the
+changelog can be found.
+
+A maintainer will reply to you shortly to get the package ready for testing.
+As soon as the package file looks good and it was sucessfully tested on both
+devices, we can add it! 🎊🎉🎊
+
+-->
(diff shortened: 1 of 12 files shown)


matteodelabre:
Tested on rM2, found no issues.

Eeems:
@matteodelabre What about Oxide 2.2.2 as it contains OS 2.9+ fixes as well?

matteodelabre:
Yes, I would have included it, it’s just that it was merged to testing after I opened this PR. There’s still time to add it to this PR if you wish. But we should merge this one as soon as possible, since it contains a fix for #422, which prevents users from installing Toltec.

Eeems:
Right, fair enough. Lets try to push this through now and open the next merge window PR immediately instead.

raisjn (approved):
tested fbink, koreader & remux on rM1. changes in general look fine to me

koreader gave some logs lines about no "librm2fb.client.so" being found (or something like this)

Eeems:
> koreader gave some logs lines about no "librm2fb.client.so" being found (or something like this)

Probably just it running binaries that don't support loading it, like Oxide does.

PR #446, opened 2021-09-27
Update rm2fb to start before launcher.service

Eeems:


matteodelabre:
Thanks for the PR! It looks good, although I’m not sure whether remux should be marked as conflicting with other launchers since it doesn’t show anything on screen unless triggered (thoughts @raisjn?). We could still force it to start after rm2fb by adding `After=rm2fb.service` to the remux service definition.

Eeems:
Remux does conflict with oxide as they both control applications with drastically different mechanisms. So having Oxide and remux run at the same time will have bad side-effects.

matteodelabre:
Yes you’re right, but remux doesn’t conflict with Xochitl, for example. Should we just leave the Conflicts list for the remux unit?

Eeems:
Good point. Although I believe remux assumes that xochitl isn't started before it? Not sure though.

matteodelabre:
If I understand remux correctly, it does not (but @ raisjn can correct me). It will detect that Xochitl (or any other app) is already running by scanning the process list.

raisjn:
remux doesnt conflict with xochitl on purpose (to reduce likelihood of bricking). if xochitl service is started, remux will stop it before launching xochitl. if xochitl is already launched (outside of system service), remux will properly manage it (iirc)

matteodelabre:
I reverted the changes for the remux unit then.

matteodelabre (approved):

PR #448, opened 2021-09-29
Create rmfakecloud-proxy package

matteodelabre:
This PR provides a new package intended to ease the setup of rmfakecloud on a device. [rmfakecloud](https://github.com/ddvk/rmfakecloud) is an alternative Xochitl sync server developed by ddvk that anyone can self-host. With this package, connecting to an rmfakecloud server only takes a few steps:

* `opkg install rmfakecloud-proxy`
* `rmfakecloudctl set-upstream https://...` (define the server to use)
* `rmfakecloudctl enable`

One can then run `rmfakecloudctl disable` (or uninstall the package) to revert back to the default reMarkable servers.

Technical details
-----------------

This package reuses most of its logic from the rmfakecloud-proxy installer <https://github.com/ddvk/rmfakecloud-proxy/blob/master/scripts/installer.sh>, but makes it more integrated with Toltec. The following steps happen when enabling rmfakecloud-proxy:

* A local certificate authority and a self-signed certificate are generated for the reMarkable cloud domain names (stored in `/opt/var/rmfakecloud-proxy`).
* The `rmfakecloud-proxy` service is enabled. This is a proxy server listening on `127.0.42.10` which forwards any requests it receives to the server selected by running `rmfakecloudctl set-upstream`.
* Entries are written to the `/etc/hosts` file to direct requests to the reMarkable cloud to the `rmfakecloud-proxy` server.

Additionally, if one is already connected to the reMarkable cloud in Xochitl when enabling rmfakecloud-proxy (or connected to rmfakecloud when disabling it), the script takes care of disconnecting them and marking their files as unsynced. This prevents Xochitl from believing that the files have disappeared from the server and therefore deleting them.

The “enabled” state survives package upgrades and system upgrades. This state is stored as a file in `/opt/etc/rmfakecloud-proxy/enabled` which exists if and only if the proxy is enabled.

Test plan
---------

Please make a [backup of your files](https://eeems.website/toltec/#2-backup) (in `~/.local/share/remarkable/xochitl`) before testing this.

To test this PR, you’ll need access to an rmfakecloud instance. I can give temporary access to my instance if needed.

The following scenarios should be tested:

* Installing and enabling rmfakecloud-proxy while previously being connected to the reMarkable cloud: you should get disconnected from the rM cloud, no files should be lost, and they should sync to the new server after connecting to your account.
* Installing and enabling rmfakecloud-proxy while not being connected to the rM cloud: your files should sync to the new server after connecting to your account.
* After enabling rmfakecloud-proxy, switching to a different rmfakecloud server with `rmfakecloud set-upstream`: you should get disconnected from the previous server and no files should be lost.
* Upgrading the system while being connected to rmfakecloud: the rmfakecloud connection should be re-enabled after running `toltecctl reenable`.
* Running `rmfakecloudctl disable` should remove entries from the `/etc/hosts` file, uninstall the self-signing CA and disconnect the user. No files should be lost.
* Removing the rmfakecloud-proxy package should disable rmfakecloud as before.


matteodelabre:
I’ve been testing this for a few days on my rM2 and haven’t had any issues. Testers wanted! (Please see the test plan above.)

matteodelabre:
Any thoughts on this @ddvk?

ddvk:
nicely done 👍

ddvk:
btw, there is a -c option now that accepts a yml config file

package/rmfakecloud-proxy/rmfakecloud-proxy.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=rmfakecloud reverse proxy
+Documentation=https://github.com/ddvk/rmfakecloud-proxy
+After=home.mount
```
Eeems: We probably should ensure that this is running before xochitl is when using the standard mechanism. It also means that xochitl will be stopped before this is on shutdown. This will avoid potential timing problems.
```suggestion
After=home.mount
Before=xochitl.service
```

Eeems: It might also make sense to have it start before `launcher.service`

package/rmfakecloud-proxy/rmfakecloudctl
```diff
@@ -0,0 +1,319 @@
...
+    local was_active
+
+    if systemctl --quiet is-active xochitl.service 2> /dev/null; then
+        systemctl stop xochitl
+        was_active=1
+    fi
+
+    if pgrep xochitl > /dev/null; then
+        killall xochitl
+        sleep 1s
```
Eeems: Wouldn't it make more sense to wait a slight bit longer, and to exit early if no xochitl processes exist?

package/rmfakecloud-proxy/rmfakecloudctl
```diff
@@ -0,0 +1,319 @@
...
+    fi
+
+    if pgrep xochitl > /dev/null; then
+        killall -9 xochitl
+        sleep 1s
+    fi
+
+    # Mark all files as not synced
+    grep sync "$xochitl_data_dir"/*.metadata -l \
```
Eeems, line 68: Do we ever expect `synced` to not exist in the file?

package/rmfakecloud-proxy/rmfakecloudctl
```diff
@@ -0,0 +1,319 @@
...
+
+    action="$1"
+    shift
+
+    case $action in
+        enable)
+            if is-enabled; then
+                echo "rmfakecloud-proxy is already enabled."
+            else
+                disconnect-cloud
```
Eeems: Should we trap various signals here that would cause an exit and disable?

package/rmfakecloud-proxy/rmfakecloud-proxy.service
matteodelabre: Good point, thanks! I also added `After=network.target`.

package/rmfakecloud-proxy/rmfakecloudctl
matteodelabre: I’m not sure, I took this line from ddvk’s install script

package/rmfakecloud-proxy/rmfakecloudctl
matteodelabre: Something like the solution proposed there? https://stackoverflow.com/a/41613532

Eeems: Yeah, I would probably use the answer below the one you linked
```bash
while kill -0 PIDS 2> /dev/null; do sleep 0.5; done;
```
And just modify it to have a timeout.

package/rmfakecloud-proxy/rmfakecloudctl
Eeems: I guess if `synced` was missing it would probably be interpreted as `false`, so it's probably fine to leave as is.

package/rmfakecloud-proxy/rmfakecloudctl
matteodelabre: You mean trapping script errors? Or do you have other cases in mind?

matteodelabre:
I'll have to investigate further but the preremove script of this package fails when invoked as part of `toltecctl uninstall`.

package/rmfakecloud-proxy/rmfakecloudctl
Eeems: Script errors, also ctrl-c or another interrupt signal.

Eeems:
> 
> 
> I'll have to investigate further but the preremove script of this package fails when invoked as part of `toltecctl uninstall`.

Should this be changed to a draft until you resolve this?

matteodelabre:
The `toltecctl uninstall` issue is due to the script relying on pgrep being available (as specified in `installdepends`) while it seems that the uninstall command does not honor dependencies and removes pgrep before this package. As a workaround, I’ve reimplemented a very basic pgrep inside the Bash script, but we may want to look into making the uninstall command properly manage dependencies in a separate issue.

matteodelabre:
I added back the dependency on pgrep. Provided that the updated uninstall script from #456 is used, this package now uninstalls without errors during `toltecctl uninstall`.

Eeems (changes requested):
Add traps for ctrl-c/other interrupt signals to avoid leaving things in a bad state.

matteodelabre:
Added an error handler to the `rmfakecloudctl enable` command that cleans up the install if it gets interrupted or encounters an error. To test, run the command and do `Ctrl-C` at any time during the install, or insert a failing command somewhere in the `install()` function of `/opt/bin/rmfakecloudctl`. Afterwards, check that the following has been cleaned up:

* `/usr/local/share/ca-certificates` should be empty
* `/opt/var/rmfakecloud-proxy` should be empty
* `/opt/etc/rmfakecloud-proxy` should be empty
* `/etc/hosts` should be free of cloud override lines
* `/etc/ssl/certs` should be free of symlinks to `/usr/local/share/ca-certificates/rmfakecloud-ca.crt`

(cc @Eeems)

matteodelabre:
I propose to add a guide to our website for how to use this package to connect to alternative sync servers. See toltec-dev/web#8.

Eeems:
Just tried to run through the first test and it looks like rmfakecloud-proxy.service fails to start after running `rmfakecloudctl enable` and there is no feedback to the user that it failed.

image

Steps that I took:

1. Installed package
2. rmfakecloudctl set-upstream ...
3. rmfakecloudctl enable
4. ctrl-c while it's running
5. rmfakecloudctl enable

Looks like the set-upstream configuration was wiped before enable was called.

After a disable, set-upstream, enable it worked. I was able to connect after generating a code, and all my documents started uploading!

rmfakecloudctl should also have a status method to check on the current upstream, if it's enabled, and if it's running.

Eeems (approved):
While I think the issue I found should be resolved, and rmfakecloudctl should be enhanced, I think it's worth getting this into testing first.

Eeems:
@matteodelabre poke?

matteodelabre:
Thanks for the review @Eeems! As you proposed I’ll go ahead and merge this, and will make a follow-up PR to address your points.

PR #450, opened 2021-10-02
[rmkit] upgrade remux with rM1 support on 2.9

raisjn:
this brings remux up to date with https://github.com/rmkit-dev/rmkit/commit/8254893999d7334fd14c162d5a15605dcff77ec5

major improvement is proper support for touch gestures in remux on rM1 after rebooting. the main issue was the display was resizing after remux started, so the touch gestures were using the wrong display size. holding the center button on rM1 would cause remux to restart (and fix this problem), so there is a workaround available.


Eeems:
Should we also add the dependency on toltec-bootstrap (https://github.com/toltec-dev/toltec/pull/431#issuecomment-932794304)?

raisjn:
I think so, but it would only happen for people like me who have been using toltec w. out bootstrap script

the path to test is: no udev rules, install launcher that depends on udev rules, see launcher hang while waiting to start

Eeems:
@raisjn looks like this branch is out of date, but I don't have permissions to update it?

matteodelabre (approved):
Tested on rM1 2.10.0 and rM2 2.10.1. Let’s handle the dependency on udev rules in a separate PR.

Issue #451, opened 2021-10-02
Toltec UX: New User


gbyl:
In order to provide the best UX possible, we should aim to ***make the least number of assumptions*** about the tech proficiency of a new toltec user. With that in mind, I have ran through a standard toltec installation:

### Issue 1 [fixed]
```
wget: bad address 'toltec-dev.org'
```
"Connect your device to Wi-Fi" should be explicitly mentioned in the README. As in "To automatically install Opkg, Entware and Toltec, connect your device to Wi-Fi and run the bootstrap script".
A new user may deem the above error to be toltec-related.

### Issue 2 [fixed]
```
Unexpected error on line bootstrap:96 in function wget-bootstrap
```
This is a well-known error that will surely be fixed soon. It is nevertheless rather disorienting for a new user.

### Issue 3 [fixed]
```
Please restart your SSH session or run 'exec bash -l' to use Toltec
```
The `1` and `l` characters can be almost identical in some default terminal fonts.
A new user may mistakenly type `exec bash -1` instead of `exec bash -l`. To avoid any ambiguity here, the long-form option should be preferred: `exec bash --login`

### Issue 4 [fixed]
```
INFO:  After each system upgrade, run 'toltecctl reenable' to re-enable Toltec
```
The help for `toltecctl` is excellent, but `toltecctl` itself should be featured prominently in the README.
A new user will likely utterly miss the fact that `toltecctl` is a fundamental component of managing a toltec install and has functions beyond reenabling.

### Issue 5 [fixed]
https://github.com/toltec-dev/toltec/pull/441 did not make it in time for https://github.com/toltec-dev/toltec/pull/443. The [new prompt](https://github.com/toltec-dev/toltec/issues/437#issuecomment-930256171) and service unit for KOReader will probably have to wait until the next merging window by the end of the month. In the meantime, is there a version of the bootstrap script that uses the testing branch by default? We may add a link to that instead in the KOReader wiki.
A new user will be unaware of the branch-switching function of toltecctl given Observation 4. They may also [miss a key step](https://github.com/toltec-dev/toltec/issues/437#issuecomment-930238513).

### Issue 6 [fixed]
It should be made clear that even after https://github.com/toltec-dev/toltec/pull/441, toltec was designed to integrate with a launcher application. This should also feature prominently in the README to avoid any kind of [workflow or terminology confusion](https://github.com/koreader/koreader/issues/8157#issuecomment-932801423).

raisjn:
[...]

PR #453, opened 2021-10-05
Address a few Toltec UX issues

gbyl:
This PR addresses UX issues (for new users) raised here: https://github.com/toltec-dev/toltec/issues/451


Eeems (approved):
lgtm

PR #455, opened 2021-10-05
Make Xochitl work even if rm2fb doesn’t start

matteodelabre:
The goal of this PR is to make Xochitl start on reMarkable 2 even in cases where the rm2fb server does not start. This happens for example if one disables rm2fb.service or if it’s ran on an unsupported system version. Making sure that Xochitl always starts can prevent soft-bricking for users who did not write down their SSH password and did not set up SSH key authentication.

The Xochitl wrapper script will now only set LD_PRELOAD if the rm2fb server is actually running (detected by checking the existence of `/dev/shm/swtfb.01`).

To implement this, I modified the script so that it sources files in `/opt/etc/xochitl.env.d` instead of only parsing them as key/value lists. This enables those files to run some logic (such as checking the existence of a file) before setting environment variables.

Test plan: On rM2, install the updated packages, enable Xochitl and disable rm2fb then reboot. Xochitl should start up anyway and the log line “rm2fb server is not running: starting without rm2fb client” should appear in `journalctl --unit xochitl`.


package/xochitl/xochitl
```diff
@@ -2,7 +2,11 @@
...
-echo "Loading custom environment variables from /opt/etc/xochitl.env.d"
-# shellcheck disable=SC2046
-export $(grep -vh '^#' /opt/etc/xochitl.env.d/* | /opt/bin/xargs -d '\n')
+for file in /opt/etc/xochitl.env.d/*; do
+    if [[ $file != /opt/etc/xochitl.env.d/README ]]; then
+        echo "Sourcing $file"
+        # shellcheck disable=SC1090
+        source "$file"
+    fi
+done
```
Eeems: I think this is probably a better solution overall. It would require the files to use the `.env` extension.
```suggestion
for file in /opt/etc/xochitl.env.d/*.env; do
    echo "Sourcing $file"
    # shellcheck disable=SC1090
    source "$file"
done
```

Eeems:
I can't test this since I don't have a rM2. @LinusCDE or @raisjn ?

matteodelabre:
I did more tests and unfortunately this patch is not enough to protect against Xochitl not starting if rm2fb is running on an unsupported system version. This is because the rm2fb server creates the `/dev/shm/swtfb.01` message queue before exiting with an error, so the test for that file is positive even when rm2fb fails.

Alternative solutions include fixing rm2fb to only instantiate the message queue after it has confirmed it’s running a supported version, or finding another way of checking whether the rm2fb server has started.

matteodelabre:
I think we can still merge this PR and a future release of rm2fb will change the server behavior to make it so that the message queue is created only if the server is running (see issue ddvk/remarkable2-framebuffer#76).

raisjn (approved):

PR #456, opened 2021-10-06
toltecctl: Honour dependencies when uninstalling

matteodelabre:
`toltecctl uninstall` will now try to uninstall packages in the reverse order of their dependencies, so that a package can use binaries it depends on during its removal scripts (see #448). Packages in a dependency cycle are removed in an unspecified order.

Test plan: Upgrade `toltec-bootstrap`, install some packages, then run `toltecctl uninstall`. You should see packages removed in reverse dependency order, so for example:

* `libc`, `libpthread`, `libgcc` should generally be the last removed packages
* graphical apps should get removed before `display`
* `toltec-bootstrap` should get removed before `coreutils-tsort`


raisjn (dismissed):
looks cool! is this much slower than normal opkg uninstall because its doing the listing of dependencies for each package?

matteodelabre:
Thanks for the review! From my tests, having installed the base Entware distribution + around 10 extra packages, the dependency ordering step took around 10 seconds. The `opkg depends` command can take up to 200–300 ms per invocation (because it’s unzipping and reading the full package list to memory each time), and makes up for much of this time since I call it once per installed package.

Your message prompted me to think about how to cut this time. It turns out `opkg depends` can list the dependencies of several packages, not just one, and even accepts wildcards. So by doing `opkg depends '*'` we get a complete list of dependencies with only one invocation of `opkg depends` (and only one full read of the package list).

```terminal
$ opkg depends '*'
terminfo depends on:
	libc
	libssp
	librt
	libpthread
libc depends on:
	libgcc
locales depends on:
	libc
	libssp
	librt
	libpthread
	grep
entware-upgrade depends on:
...
```

We should be just a simple awk script away from turning this into a dependency-ordered list of packages. I’m going to take a look into this and will report on the time saved by this approach.

EDIT: The following awk script and Bash pipeline produce the expected ordered list of packages in about 400 ms.

```awk
/^.* depends on:$/{
    from=$1;
    print from " " from;
}

/^\t/{
    print from " " $1;
}
```

```bash
opkg depends '*' | awk "$awk_script" | tsort
```

Eeems:
Looks good to me, what else is it waiting on?

matteodelabre:
I haven't replaced the package ordering code with the faster version yet. I'll take care of it tomorrow.

matteodelabre:
I added the new (faster) code for ordering packages. I executed the test plan again on rM2 and it’s working fine, there is no longer a noticeable delay during the package ordering step.

raisjn (approved):

PR #461, opened 2021-10-12
Update rm2fb to v0.0.8

Eeems:
Add 2.10 signature support for rm2fb, start including rm2fb.conf automatically so we can push updates without needing to get a new release of rm2fb.


Eeems:
It might make sense to remove signatures from the file for anything older than 2.8 since we don't really support that in toltec anymore.

matteodelabre:
I think it would be better to upstream this, for those who use rm2fb without Toltec.

matteodelabre:
Just saw your other PR ddvk/remarkable2-framebuffer#75 (thanks for working on this by the way!). What would be the rationale for duplicating the offsets data between the upstream repo and Toltec?

Eeems:
> 
> 
> Just saw your other PR [ddvk/remarkable2-framebuffer#75](https://github.com/ddvk/remarkable2-framebuffer/pull/75) (thanks for working on this by the way!). What would be the rationale for duplicating the offsets data between the upstream repo and Toltec?

As per the PR description:
> Add 2.10 signature support for rm2fb, start including rm2fb.conf automatically so we can push updates without needing to get a new release of rm2fb.

These signatures have been known for a while, and upstream hasn't been updated with them. I'd prefer to avoid that upstream reliance when we do have the ability to push out quick updates to a configuration file. Ideally we just get around to implementing https://github.com/ddvk/remarkable2-framebuffer/issues/69 and there is no need to update upstream or toltec for a signature database update.

matteodelabre:
Since @raisjn & I have write access to the [ddvk/remarkable2-framebuffer](https://github.com/ddvk/remarkable2-framebuffer) repo, it’s not harder to make a new rm2fb release than it is to update the Toltec package, so I’m not sure it’s worth the effort of maintaining two copies of the offsets datafile.

Eeems:
> 
> 
> Since @raisjn & I have write access to the [ddvk/remarkable2-framebuffer](https://github.com/ddvk/remarkable2-framebuffer) repo, it’s not harder to make a new rm2fb release than it is to update the Toltec package, so I’m not sure it’s worth the effort of maintaining two copies of the offsets datafile.

Yes you do, but my PR over there hasn't been merged yet, and this has had more progress ;)

matteodelabre:
Sorry but, while I see your point and agree that we need fast updates for rm2fb, I’m really not convinced by this approach. I think that introducing two different configuration files for Toltec users and for those who manually installed rm2fb will lead to confusion. ddvk/remarkable2-framebuffer#69 is a better solution and until it gets implemented we can rely on upstream releases.

@raisjn’s input on this would be valuable.

Eeems:
> I think that introducing two different configuration files for Toltec users and for those who manually installed rm2fb will lead to confusion.

I'm not sure I understand. The one configuration file I've introduced is part of the paths rm2fb searches for the file. I do agree that auto updates is a better long term solution, but until someone builds that we do need a way to allow the toltec package to move faster.

raisjn (approved):

matteodelabre (approved):

Issue #445, opened 2021-09-27
Warn users when installing Toltec on an unsupported OS version


matteodelabre:

LinusCDE:
That's a really good idea. How about using remarkable-framebuffers list of versions in a whitelist used for installation as well as reenabling (or downloading it).
If `/etc/version` has no match in this list, we should prevent installtion and reenabling by default and require setting an env to bypass this (e.g. `toltec_ignore_version_error=1`).

Eeems:
Instead of an environment variable to bypass, it might make more sense to just have a `--force` command line argument.

LinusCDE:
That is a good idea, but I think this shouldn't be too easy. If it were `add --force to it`, some users might simply bypass that check and then wonder why the device doesn't reboot properly.

Saying `set the env toltec_ignore_version_error to 1` should rule a good portion of those users preventing them from accidentally bricking their devices.

Eeems:
I think that will only work for a short while, until someone tells someone to just run it with `toltec_ignore_version_error=1` before the command and then everybody will start using it that way.

I think we should just not tell people to use `--force` if the reenable fails, and have a strong warning in the `--help` output.

LinusCDE:
Good point! :+1:

dominique-unruh:
> I think we should just not tell people to use `--force` if the reenable fails, and have a strong warning in the `--help` output.

Maybe it should have a more dangerous name. `--force-and-brick-my-device` or something. I guess most users will then at least check the documentation or Google for it.

Eeems:
The help output would have the documentation required in my proposal. You'll find that having more "dangerous" names doesn't actually keep users from using it without knowing what it actually does. As soon as one person posts a blog post, or reddit comment saying to use it, people will just start using it without reading up on what it does.

ZeKap:
It looks like no one did a PR for it, may I try to make one?

Eeems:
There is a linked PR for this already

PR #476, opened 2021-10-21
Update README to include warning from the website

jmrtsh:
Well I just soft-bricked my remarkable 2 because I followed the instructions here, which could have been avoided if the README contained the same warning as the Website.

I am pretty sure your "pull requests to stable cannot be accepted" rule does not apply to README updates, but if thats not the case then I'll create another one for testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@
 
 Toltec is a community-maintained repository of free software for [the reMarkable tablet](https://remarkable.com/).
 
+### Pre-Installation WARNING
+Toltec does not currently support OS 2.10 or newer on the reMarkable 2. You will likely soft-brick your device if you install before support is released. See [remarkable2-recovery](https://github.com/ddvk/remarkable2-recovery) for information on how to recover your device if you have done this.
+
 ### Install it
 
 Toltec works on top of the [Opkg](https://code.google.com/archive/p/opkg/) package manager and the [Entware](https://github.com/Entware/Entware) distribution, which are in widespread use in embedded devices.


LinusCDE:
Whether changes of the readme to stable are accepted I'm unsure.

I've installed toltec on my rM 2 with 2.10 support. As long as rm2fb is not installed, it shouldn't softbrick the device. The problem is, that for every release, rm2fb needs updated signatures to work properly or xochitl won't start. The issue is updating rm2fb (or just the signatures) fast enough and there are ongoing discussions about this in issue #445 and pr #461 on updating the signatures of rm2fb faster without waiting on a new release of the software. I think there were more discussions, but I'm unsure where or if just in discord).

In general I approve of adding this disclaimer for now. But I think we should have a more generalized disclaimer, that
 - warns for this very issue
 - shows where to check the currently supported xochitl version with rm2fb (though completion of the issue #445 might address this automatically)
 - urges people to write down their ssh password beforehand

Side note: rm2fb "only" bricks the selected launcher/xochitl from starting up if on an unsupported version and ssh should still work (at least over usb). PR #455 should resolve this issue in the future.

As said, I'm unsure whether changes to stable are accepted for the readme but find it logical as well. @Eeems or @matteodelabre, if this is okay, I would suggest merging this.

jmrtsh:
> Side note: rm2fb "only" bricks the selected launcher/xochitl from starting up if on an unsupported version and ssh should still work (at least over usb). PR #455 should resolve this issue in the future.

Yes, I fortunately managed to get back into it via ssh and uninstall toltec again, after the reboot it worked again :D

Eeems:
@LinusCDE the following will make it so we don't really need to have a warning anymore after upgrades:
- https://github.com/ddvk/remarkable2-framebuffer/issues/69
- https://github.com/ddvk/remarkable2-framebuffer/issues/76
- https://github.com/toltec-dev/toltec/pull/455

It's probably fine to add to the stable readme for now, but it'll have to be removed after the next stable merge.

@matteodelabre thoughts?

jmrtsh:
It might also be advantageous to only have one place where the installation process is detailled to avoid such inconsistencies in the future. Maybe the website can just link to the installation part of the README?

Eeems:
> 
> 
> It might also be advantageous to only have one place where the installation process is detailled to avoid such inconsistencies in the future. Maybe the website can just link to the installation part of the README?

I think it would make more sense to have the readme link to the website. That said, this isn't going to be an issue for much longer due to issues/prs that both @LinusCDE and I linked.

jmrtsh:
Well which way around the installation instructions are, is for you to decide, I also thought it could prevent similar issues in general.

matteodelabre:
I also had in mind of redirecting users to the website for install instructions. As a nice side effect it would also free us from having to maintain the bootstrap checksum manually in the readme. As a temporary solution I think it’s fine to merge this PR.

Eeems (approved):

Eeems:
I'll update the next stable merge to include removing the warning.

Eeems:
I just realized that we just pushed 2.10 support.. maybe we should remove the warning now and remove it from the website? @matteodelabre

PR #481, opened 2021-10-24
Update dotnet to 3.1.20

Eeems:
Update to the latest 3.1 version.
 We might want to look at also packaging dotnet 5 and allowing it to install alongside, but I'm not sure how best to manage it.


LinusCDE (approved):
Tested on rM 2.

Created a new default Hello World Console Application using the `dotnet` cli. After logging in again, the ICU error (language not set??) was fixed and `dotnet new "Console Application" --language "C#" -n hello_world` worked flawlessly. The application could then be compiled and executed using `dotnet run` without issues.

(Maybe add a hint in a future release that relogging might be needed.)

PR #484, opened 2021-10-25
Add doomarkable 0.4.0

LinusCDE:
Here it is!

Tested the package to build and run fine on the rM 1 with all major launchers.

While the rM 2 is supported it isn't really pleasant to play. (Not sure if it ever will.) I'll look whether I can do anything the coming weeks. Otherwise we could also consider marking it rm1-only.

The game runs at about 13 FPS. The recent brightness bump made the game a lot better to look at, but also increased artifacts a bit. There is certainly a lot of things to do (a bit more performance, better/multiple input schemes, todolist on readme), but this should be a solid game already.

- Keep in mind that even when the game doesn't seem to run, while the app is active, it sucks a lot of battery! I need to solve why doom keeps consuming cpu even when quit.)
- To run the game, you need a wad file. The easiest to get is from [here](https://doomwiki.org/wiki/DOOM1.WAD). Place it into the home directory and be sure to rename it to all uppercase. The app should note this however.

### DIFF
--- a/package/doomarkable/package
+++ b/package/doomarkable/package
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(doomarkable)
+pkgdesc="DOOM game"
+url=https://github.com/LinusCDE/doomarkable
+pkgver=0.4.0-1
+timestamp=2021-10-25T23:02Z
+section="games"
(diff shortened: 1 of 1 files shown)


Eeems (approved):
Works for me! It is barely playable at the current framerate. Likely some sort of filter on all the backgrounds to make them brighter but other stuff darker would help.

When saving your game you can't type in anything other than numbers, it would be nice to pull up a full keyboard when saving.

Quitting from the in game menu can't be done either due to the lack of a Y key.

PR #488, opened 2021-11-03
package: linux-mainline: Initial commit

alistair23:
This is the initial commit of the mainline kernel. This is based on
current Linux mainline with some extra reMarkable patches on top. This
doesn't yet have the same support as the vendor kernel, but does have
most of the core functionality.

You can see the diff with mainline here: https://mainline.space/

This simplifies the install process specified here:
https://remarkablewiki.com/tech/updatingkernel

Signed-off-by: Alistair Francis <alistair@alistair23.me>


Eeems:
@alistair23 could you list what the delta between the features of mainline and vendor are?

Also, no rm1 kernel? :(

alistair23:
So far everything I have tried works fine, so I don't think there are many deltas.

Someone else is welcome to work on the rm1 kernel. That's another massive effort

alistair23:
Any comments?

Eeems:
I can't test as I don't have a rM2.

matteodelabre:
I’d love to test this but I still haven’t managed to make the rM2 recovery work. I have all the required parts but no soldering iron at hand, yet some soldering seems to be required to make the connection stable enough. I’m on a tight budget right now so it’ll probably be a few months until I buy one. Anyone else available to test this?

Etn40ff:
I had a quick look at the package and I have a question: why did you decide not to perform the actions in configure and postremove?

Eeems:
> I had a quick look at the package and I have a question: why did you decide not to perform the actions in configure and postremove?

Because of the risk of bricking the device. That said, we probably should include some sort of helper script to automate the actions.

Etn40ff:
Why do you think that running the code in the package is any riskier than asking the user to run the same commands by hand?

Anyway I agree: a script would be nice to have.

Having said this, given the amount of things not working with a mainline kernel, I doubt this package will be used for other than development.

Eeems:
> Why do you think that running the code in the package is any riskier than asking the user to run the same commands by hand?

Because you have to choose to make the change. Doing it automatically could completely break things for users without them expecting it on upgrade or by just poking around.

That aside we tend to avoid making changes like this with toltec install scripts. You have to explicitly enable services when you install them. We aren't like debian, who enables services on install by default.

> Having said this, given the amount of things not working with a mainline kernel, I doubt this package will be used for other than development.

That list should shrink as time goes on.

Etn40ff:
Ok, I see the point.

I gave this package a spin and it works as advertised. The only thing that needs to be changed is line 45 because it mentions the wrong version. Maybe this should automatically point to pkgver.

package/linux-mainline/package
```diff
@@ -0,0 +1,49 @@
...
+    echo "To install them run and then reboot:"
+    echo "  mkdir -p /opt/boot-backup"
+    echo "  cp -rf /boot/* /opt/boot-backup/"
+    echo "  cp -rf /usr/boot/* /boot/"
+    echo "  /bin/sync"
+}
+
+postremove() {
+    echo "To restore to the original kernel, run and then reboot"
+    echo "  rm -rf /lib/modules/5.4.94/"
```
Eeems: This should be dynamic

alistair23:
> I had a quick look at the package and I have a question: why did you decide not to perform the actions in configure and postremove?

I was worried that someone would install the package without realising and then upgrade their kernel. By adding the manual install steps it's harder to install, but not too hard. As for removing I didn't want to delete modules from a running system unless the user really wanted to.

> Having said this, given the amount of things not working with a mainline kernel, I doubt this package will be used for other than development.

I'm using it daily, it does some a few issues but there aren't that many.

alistair23:
Just rebased and address the outdated string comment. Hopefully this is ready to merge now

alistair23:
Any more comments?

Etn40ff:
> Any more comments?

Sorry it is taking me ages to find some spare time to review this. I'll try to have a look at it today.

matteodelabre:
I think the list of known issues with the mainline kernel (as mentioned in the original post) should be shown to users in `configure()`.

Etn40ff:
Something is still not working with the uninstall instructions: it says 
`rm -rf /lib/modules/5.16.0/` but it should say `rm -rf /lib/modules/5.16.0-rc8/`

I am also having issues intalling this due to my partition being almost full. Can a bigger part of the package be offloaded to /opt?

Etn40ff:
Note to self: when this package does not install properly because of space constraints be careful about the backup: you might end up backupping mainline rather than upstream and then, when reverting, you end up with a brick due to mssing modules.

Etn40ff:
A follow-up comment due to my space constraints. Would it  be safer to install directly the kernel into `/boot` rather than `/usr/boot` and then ask the user to update symlinks? In this way if there is a failure cause by disk space it happens at install time and a simple `opkg remove` should fix the problem automagically.

Eeems:
It might be a good idea to add a utility to manage switching kernels and validating that backups are properly in place etc. `kernelctl` or something. That way it can handle failure more gracefully instead of assuming users will fully understand the implications of issues they encounter.

Etn40ff:
> It might be a good idea to add a utility to manage switching kernels and validating that backups are properly in place etc. `kernelctl` or something. That way it can handle failure more gracefully instead of assuming users will fully understand the implications of issues they encounter.

Since it looks unlikely that they will revert the change to `TRIM_UNUSED_KSYMS` (cf. https://github.com/reMarkable/linux/issues/8#issuecomment-1043664468) I am considering making another toltec package with an upstream kernel, `TRIM_UNUSED_KSYMS=n`, and maybe some exra modules like `uinput`. We may want to coordinate such a utility to deal with all aftermarket kernel toltec ships.

Etn40ff:
Here is a proposal for a script, say `kernelctl`, to manage aftermarket kernels.

`kernelctl` provides 4 commands:

- `kernelctl help`: the obvious thing,
- `kernelctl init`: to be run when it is installed and every time one updates the OS,
- `kernelctl list`: show a numbered list of available kernels (the current one being marked by a *),
- `kernelctl set number/name`: configure the system to boot the desired kernel.

Kernel packages in toltec will only install a tar.gz file inside a staging
directory, say `/opt/kernelctl`. Ideally the file should be named something
like `kernelname-version.tar.gz`. It should be possible to extract such a file
in / to have a working system booting kernelname. By this I mean that it should
contain `lib/modules/version/`, `boot/zImage`, and anything else that is
required respecting the correct paths.

`kernelctl init` should create `/opt/kernelctl`, a backup of the current
running kernel in it, say `upstream-version.tar.gz`, and write
`current_kernel=upstream-version` in some configuration file (I do not know
which could be a sane location for this file to be).

`kernel set number/name` should list the content of `current_kernel-version.tar.gz`
and use this list to delete the corresponding files, extract
`new_kernel-version.tar.gz` and set `current_kernel=new_kernel-version`,
optionally ask to reboot.

Uninstalling `kernelctl` should first run `kernelctl set default`, then remove `/opt/kernelctl`.

All kernel packages in toltec should depend on `kernelctl` and say that to enable the newly installed kernel one should use it.

Eeems:
> `/opt/kernelctl`

It would need to be stored elsewhere, `/opt` mirrors the standard folder structure. So `/opt/usr/share/kernelctl` would be a better location.

Etn40ff:
> > `/opt/kernelctl`
> 
> It would need to be stored elsewhere, `/opt` mirrors the standard folder structure. So `/opt/usr/share/kernelctl` would be a better location.

sounds good to me, any other comment/change?

Eeems:
Nothing jumps out to me at the moment.

alistair23:
Just addressed the comments on this PR, let me know if you think it should directly be installed in `/boot` instead.

A `kernelctl` tool would be useful to have as well

Etn40ff:
> A `kernelctl` tool would be useful to have as well

I am working on it, it should be available soon. Meanwhile could you please make so that this package uses it? 

Here is what needs to be done:

- this package has to depend on a package not yet in toltec that will be called `kernelctl`
- all the files needed to deploy your kernel need to be in a `.tar.bz2` file named something like `kernelname-kernelversion.tar.bz2` (for example the kernel shipped by remarkable will be saved in a file of the form `upstream-20220202133838.tar.bz2`)
- to boot your kernel it should suffice to extract `kernelname-kernelversion.tar.bz2`  in `/`, do not worry about any cleanup of other kernels
- this package should only install `kernelname-kernelversion.tar.bz2`  in `/opt/usr/share/kernelctl` and print a message directing users to `kernelctl` to boot the new kernel

Thanks

matteodelabre:
I was finally able to make the recovery steps work, so I’ll also give this kernel a try as soon as I can!

Etn40ff:
Here is a first attempt at kernelctl https://github.com/Etn40ff/toltec/tree/kernelctl
Please bear in mind that I have not yet tested it so there will be (plenty of) bugs.
Suggestions are welcome.

The interface is heavily inspired by `eselect kernel` that does similar things for gentoo.

I still need to provide an install script. I would like `kernelctl init` to be run by `toltecctl reenable` so I might have to refactor it accordingly.

Eeems:
@Etn40ff I'd recommend getting a PR opened so we can provide feedback there.

Etn40ff:
I will as soon as I have the package closer to being reviewed

Eeems:
#569 has been opened. We'll probably want to block this PR until it's done.

Etn40ff:
I am hijacking this PR to talk to @alistair23 

Did you figure out why is RDNIS not working? I just noticed that it does not work also in a test kernel that I built directly from the vendor sources. In theory it should work there.

alistair23:
I haven't figured out the RDNIS issue. That's strange it's not working from the vendor sources, have you changed any configs? Also I'm in the reMarkable Matrix room if you want to talk there

Eeems:
> I haven't figured out the RDNIS issue. That's strange it's not working from the vendor sources, have you changed any configs? Also I'm in the reMarkable Matrix room if you want to talk there

Which is linked to #general in discord.

Etn40ff:
I just realized that this kernel is twice as big as the default one: zImage is 12Mb compared to 6Mb . Do you have a guess on why is it so?

alistair23:
I would guess the size difference is related to the defconfig, this is just using the in-tree defconfig which enables lots of extra features

Etn40ff:
I see. Maybe later on (after getting all components to work) it could be useful to try to shrink the size: space is unfortunately very limited in the root partitions.

On a related note, in case you missed it I made a preliminary attempt to modify this package to use `kernelctl`: https://github.com/alistair23/toltec/pull/1

matteodelabre:
It works! 🎉
```console
$ uname -a
Linux reMarkable 5.18.0 #1 SMP Sat Jun 25 19:41:28 UTC 2022 armv7l GNU/Linux
```
Thanks @alistair23 for the excellent work, and sorry for the abysmally slow review process. From some initial tests, everything seems to work as expected, so I think we can merge this to testing.

I added some minor changes above (most notably moving the package to a dedicated kernel section). I’m working to change the `package()` a little bit to produce a kernel archive suitable for #569 and then we should be good for merge.

matteodelabre:
I added the change described in my previous comment, based on @Etn40ff’s work in https://github.com/alistair23/toltec/pull/1 (thanks!). The plan will be to merge this PR first (and leave in the manual post-install instructions for now), then use this package to test and finalize #569.

matteodelabre (approved):

Issue #9, opened 2020-09-05
[request] add packages for apps section in awesome-remarkable


raisjnn:
https://github.com/rehackable/awesome-remarkable

some missing packages:

* keywriter
* wikipedia
* rmvncclient
* pocket2rm

non listed packages that might be interesting:

* recrossable
* syncthing
* ???

Eeems:
- [ ] [Crazy Cow](https://github.com/machinelevel/sp425-crazy-cow) (MIT)
- [x] [rmWacomToMouse](https://github.com/LinusCDE/rmWacomToMouse) (MIT)
- [ ] [srvfb](https://github.com/merovius/srvfb) (Apache 2.0)
- [ ] [remarkable_news](https://github.com/evidlo/remarkable_news) (GPL-3.0)
- [ ] [remarkable-touchgestures](https://github.com/ddvk/remarkable-touchgestures) (MIT)
- [ ] [reStream](https://github.com/rien/reStream) (MIT)
- [x] [remarkable-splash](https://github.com/ddvk/remarkable-splash) (MIT)
- [ ] [1bitpaint](https://github.com/HiroshiTamura/1bitpaint_for_reMarkable) (MIT)
- [ ] [ReMarkable.NET](https://github.com/parzivail/ReMarkable.NET) (MIT)
- [ ] [Xochitl templates](https://remarkablewiki.com/tips/templates)
    - https://github.com/bliekp/remarkable-templates (CC0-1.0)
    - https://github.com/oisinmulvihill/reMarkableTemplates (GPL-3.0)

matteodelabre:
Thanks for the suggestions. Some of these apps lack a license that would allow us to redistribute it, I’ve marked them by editing your comments.

Eeems:
Time to open up issues asking for a license to be added :)

**Edit:** I've opened issues and edited the comment with the issues.

matteodelabre:
Thanks a lot @Eeems!

Eeems:
Looks like srvfb does have a license (Apache 2.0) at the bottom of the README, and in the header of every file.

LinusCDE:
- [x] rmWacomToMouse / rmServeWacomInput

I packaged [rmWacomToMouse](https://github.com/LinusCDE/rmWacomToMouse) yesterday. I named it after the server side component (the c version) rmServeWacomInput (as a package in all lowercase). I still need to update the readme though.

[...]

PR #490, opened 2021-11-04
package: mmc-utils: Initial commit

alistair23:
Add support for the mmc-utils package. This can be used to determine the life of the MMC, for example as described here: https://www.cnx-software.com/2019/08/16/wear-estimation-emmc-flash-memory/

This also completes part of: https://github.com/toltec-dev/toltec/issues/9

### DIFF
--- a/package/mmc-utils/0001-Makefile-Remove-Werror.patch
+++ b/package/mmc-utils/0001-Makefile-Remove-Werror.patch
@@ -1,0 +1,39 @@
+From 898bae517a42c1bf088e1c7b2bd34cd1fb086d22 Mon Sep 17 00:00:00 2001
+From: Alistair Francis <alistair@alistair23.me>
+Date: Wed, 3 Nov 2021 22:23:46 +1000
+Subject: [PATCH] Makefile: Remove -Werror
+
+Signed-off-by: Alistair Francis <alistair@alistair23.me>
+---
+ Makefile | 6 ++----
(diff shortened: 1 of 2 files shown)


Eeems (approved):
Installs and works for me on my rM1. Did you test on a rM2?

We probably should add some sort of helper script to parse the output and provide better health information. Most of it is vendor specific fields or just an overload of information to wrap your head around.

Eeems:
The good news is, my device is still between 0-10% usage:
```shell
reMarkable: ~/packages/ mmc extcsd read /dev/mmcblk1 | grep -i LIFE
eMMC Life Time Estimation A [EXT_CSD_DEVICE_LIFE_TIME_EST_TYP_A]: 0x01
eMMC Life Time Estimation B [EXT_CSD_DEVICE_LIFE_TIME_EST_TYP_B]: 0x01
reMarkable: ~/packages/ mmc extcsd read /dev/mmcblk1 | grep -i EOL
eMMC Pre EOL information [EXT_CSD_PRE_EOL_INFO]: 0x01
```

Maybe some patches to add the vendor specific field handling would also be in order.

alistair23:
I tested this on my rM2, so it works there.

I don't have plans to add scripts or vendor specific patches. Someone is welcome to add a script though. For vendor patches it's probably worth trying to send them upstream first.

I also see between 0-10% usage for my flash which is good :)

PR #492, opened 2021-11-16
Package netevent

Etn40ff:
This pull request is to package `netevent` https://github.com/Blub/netevent a
software to share input devices across linux machines. I am not the author of
this software. The package was tested on rm2.


package/netevent/package
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(netevent)
+pkgdesc="Input-Event device cloning utility"
+url="https://github.com/Blub/netevent"
+pkgver=68e648e-1
```
Eeems: By the looks of it, the 2.1 tag is the latest version and should be used instead? It looks like you are using unreleased code? It might be best to ask for a release tag to be added. Using git hashes for a version number will cause issues with upgrades not being detected as newer. See https://github.com/toltec-dev/toltec/blob/stable/docs/package-guide.md#writing-the-recipe

Eeems (changes requested):

package/netevent/package
Etn40ff: 2.1 does not incorporate some recent changes that are needed to pair this with vnsee. I'll try to ask for a release tag.

Eeems:
Current waiting on https://github.com/Blub/netevent/issues/23 to get a new release tag to work with from upstream.

@Etn40ff another option would be to just use `2.2~beta` or something for now until an official release tag is added?

Etn40ff:
> @Etn40ff another option would be to just use `2.2~beta` or something for now until an official release tag is added?

It looks like @Blub is busy at the moment so I went ahead and used a beta tag. 
I will update the package once an upstream release is available.

Eeems (dismissed):
Works for me on my rM1!

package/netevent/package
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(netevent)
+pkgdesc="Input-Event device cloning utility"
+url="https://github.com/Blub/netevent"
+pkgver=2.2-beta-1
```
matteodelabre: opkg considers `2.2-beta-1` newer than `2.2-1`, which may be an issue if this package is updated to a non-beta version number later on. `2.2-1~beta` sorts lower than `2.2-1`. To check version ordering, you can use `opkg compare-versions`:
```console
$ opkg compare-versions 2.2-beta-1 "<<" 2.2-1 && echo "lower" || echo "not lower"
not lower
$ opkg compare-versions 2.2-1~beta "<<" 2.2-1 && echo "lower" || echo "not lower"
lower
```

```suggestion
pkgver=2.2-1~beta
```

matteodelabre (changes requested):

package/netevent/package
Etn40ff: I spoke too early: 2.2.1 is out

matteodelabre:
I’m trying to test this on rM2 but it tells me `uinput` is missing. Is there something I’m missing or does this need a dependency?

Etn40ff:
You need `uinput` on your host machine.

matteodelabre:
Oh, I thought this was a way to share e.g. my keyboard towards the reMarkable, not the other way around.

Etn40ff:
You can but then you would need the uinput kernel module

matteodelabre:
Do I need to set up an SSH daemon on my host to make this work or is there a simpler option?

Etn40ff:
I use ssh. 

The manual reads:
```OUTPUT_SPEC can currently be either a file/fifo, a command to pipe to when prefixed with exec:, or the name of a unix or abstract socket when using unix:/path or unix:@abstractName```

**Comment by Eeems** [comment]
@matteodelabre poke?

**Comment by matteodelabre** [comment]
Sorry, it’s not clear to me what steps I should take to test this.

**Comment by Etn40ff** [comment]
> Sorry, it’s not clear to me what steps I should take to test this.

Here is my use case, you may want to test something along these lines.

You will need a host machine `host` running linux with `uinput` enabled in the kernel, `netevent` installed, an ssh server, and a user `user` with privileges to write to `/dev/uinput`.

ssh to your reMarkable then create a file, say `/tmp/netevent-setup.ne2`, containing the following (edit user@host accordingly and if you are not on rm2 change the device paths)
```
device add pen /dev/input/by-path/platform-30a20000.i2c-event-mouse
#device add touch /dev/input/by-path/platform-30a40000.i2c-event
output add remote exec:ssh user@host netevent create
use remote
grab-devices on
write-events on
```

finally run 

```
netevent daemon -s /tmp/netevent-setup.ne2 /tmp/netevent-socket.sock
```

on `host` you should see a Wacom tablet among the input devices.

Bonus points, you can uncomment the line about the touchscreen but I was not able to make it work nicely yet. It requires some minor dbus hacks.

**Comment by matteodelabre** [comment]
Thanks for the detailed instructions! This seems to be working well, I can see the following kernel log line after starting the daemon:
```
input: Wacom I2C Digitizer as /devices/virtual/input/input43
```

Unfortunately, and this is probably a bug in netevent, sway rejects the newly created input with the following error:
```
[ERROR] [wlr] [libinput] event25 - Wacom I2C Digitizer: libinput bug: missing tablet capabilities: resolution. Ignoring this device.
```

The package in itself looks fine though. Thanks for your contribution!

**Review by matteodelabre — APPROVED** [review]

**Comment by Etn40ff** [comment]
> Unfortunately, and this is probably a bug in netevent, sway rejects the newly created input with the following error:
> 
> ```
> [ERROR] [wlr] [libinput] event25 - Wacom I2C Digitizer: libinput bug: missing tablet capabilities: resolution. Ignoring this device.
> ```

I do not think it is related to wayland but to the digitizer itself: apparently it is not a digitizer well supported on standard linux.
Writing the following to `/etc/udev/hwdb.d/60-remarkable-wacom.hwdb` will make it work:

```
evdev:input:b0018v056Ap0000*
# LIBINPUT_DEVICE_GROUP=f865e87b
 ID_INPUT=1
 ID_INPUT_TABLET=1
 ID_INPUT_JOYSTICK=0
 EVDEV_ABS_00=::20
 EVDEV_ABS_01=::20
```

PS: the commented out line is for making touch arbitration work when using `netevent` to share both touchscreen and digitizer.

EDIT: thanks for the review

PR #495, opened 2021-11-19
Add /opt/etc directory to xochitl oxide application

murchu27:
This directory needs to be visible for rM2 users who have a custom `/opt/etc/rm2fb.conf` - usually due to running a tablet version that rm2fb that does not support yet.

See Eeems/oxide#231. Originally fixes Eeems/oxide#228.

### DIFF
--- a/package/xochitl/package
+++ b/package/xochitl/package
@@ -5,8 +5,8 @@
 pkgnames=(xochitl)
 pkgdesc="Read documents and take notes"
 url=https://remarkable.com
-pkgver=0.0.0-9
-timestamp=2021-09-25T16:36Z
+pkgver=0.0.0-10
+timestamp=2021-11-20T23:31Z
 section="readers"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=MIT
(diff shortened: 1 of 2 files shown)


Eeems (dismissed):
lgtm. Anybody with a rm2 want to test?

matteodelabre:
I can take a look at it tonight.

matteodelabre:
@Eeems Is there a way to allow access to a single file rather than a whole directory? Allowing access to the whole `/opt/etc` directory just for one file seems too wide. Not a blocker if it’s not possible, though.

Also, I think we should also allow access to the other locations the rm2fb.conf file can be read from, namely `/usr/share`, `/opt/share`, and `/etc`.

Eeems:
It uses bind mounts, which can work, but it can have some issues iirc. Like it doesn't like if the file doesn't exist.

Eeems:
> Also, I think we should also allow access to the other locations the rm2fb.conf file can be read from, namely `/usr/share`, `/opt/share`, and `/etc`.

Only if we document that users should place it there. I think if we just document `/opt/etc` it should be fine. `/etc` and `/usr/share` are already part of the list, so we would only be talking about `/opt/share` which really shouldn't be edited by users and only populated by packages.

Eeems (approved):

Eeems (changes requested):
Looks like `/opt/etc/xochitl.env.d` is missing from the new version.

Eeems (approved):
Oh I see, it was replaced by `/opt/etc`

PR #497, opened 2021-11-20
Stable merge for week 46 of 2021

Eeems:
New packages:

- doomarkable
- mmc-utils

Updated packages:

- whiteboard-hypercard - 0.3.4
- vnsee - 0.4.1
- aspnet-runtime, aspnet-targetting-pack, dotnet-profile, dotnet-host, dotnet-sdk, dotnet-targetting-pack, netstandard-targetting-pack - 3.1.20
- display, rm2fb-client 0.0.9

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,7 +18,7 @@ If you’re updating an existing package, please give information on where the
 changelog can be found.
 
 A maintainer will reply to you shortly to get the package ready for testing.
-As soon as the package file looks good and it was sucessfully tested on both
+As soon as the package file looks good and it was successfully tested on both
 devices, we can add it! 🎊🎉🎊
 
 -->
--- a/package/display/package
+++ b/package/display/package
@@ -8,7 +8,7 @@ timestamp=2021-08-23T09:53:41Z
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
 url="https://github.com/ddvk/remarkable2-framebuffer"
-pkgver=1:0.0.8-2
+pkgver=1:0.0.9-1
 _release="${pkgver%-*}"
 _release="v${_release#*:}"
 _libver=1.0.1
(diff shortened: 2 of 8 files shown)


LinusCDE (approved):
Tested on rm 2. Installs without any problem.

2 points though (which I consider non-critical):

### Segfault for rm1 packages (on rM 2)

While this will not happen in practice (since remote repos are used), for some reason attempting to install a rm1 only package straight segfaults instead of complaining about missing/wrong arch as it used to.

Not sure if there is a weird change in the format.

I **highly recommend testing the rM 1 dummy packages on a rM 1** (rm2fb, display) just to be sure. There doesn't seem to be any backtrace of value.



### Rm2fb doesn't include the newest release or config

We should start prioritizing shipping at least the most bleeding edge possible rm2fb in future stable merges.

Aside from that it seems to work well on rM 2.

Eeems:
> While this will not happen in practice (since remote repos are used), for some reason attempting to install a rm1 only package straight segfaults instead of complaining about missing/wrong arch as it used to.

Did you try removing the rm2 version of the package first?

> We should start prioritizing shipping at least the most bleeding edge possible rm2fb in future stable merges.

That's next weeks merge, which is why I'm pushing to have this merged today.

LinusCDE:
> Did you try removing the rm2 version of the package first?

I didn't consider it necessary. I just tested on rM 1 and the applied in reverse. Segfaults on rm2-only packages.

I removed rm2fb-client from the rm1 kinda and after that attempting the rm2 package, opkg stated it was up-to-date instead. So probably some weird opkg behavior/crash. Never encountered this before though.

> That's next weeks merge, which is why I'm pushing to have this merged today.

Nice!

Eeems:
I'm also able to replicate the segmentation fault on my rm1 when trying to install rm2 packages. @matteodelabre any ideas? We might need to open an issue with upstream if it's a newer version of opkg that's introduced this.

PR #499, opened 2021-11-21
Fix bootstrap script dep error during toltec-bootstrap install

matteodelabre:
Following #456, a dependency on coreutils-tsort was added to the toltec-bootstrap package. That dependency is only needed during uninstall and is used to sort the packages in reverse dependency order.

The bootstrap script does an initial temporary install of toltec-bootstrap in order to use the toltecctl definitions. It does not expect the package to have dependencies. As a consequence, the script fails to perform installs in its current state.

This fix makes the script ignore dependencies for the temporary install. The dependencies do get installed at the end of the install procedure. I also added a warning above the dependency list in toltec-bootstrap to inform future contributors that those dependencies will not be honored in the initial temporary install of toltec-bootstrap.

Sorry that I missed this when writing #456!

**To test:** simply use the bootstrap script from this PR to make a new Toltec install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ To automatically install Opkg, Entware and Toltec, make sure your device is conn
 
 ```sh
 wget http://toltec-dev.org/bootstrap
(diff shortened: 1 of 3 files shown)


matteodelabre:
Tested on rM2.

raisjn (approved):
accept to unblock

PR #501, opened 2021-11-26
[rmkit] update remux to 0.12-1 (custom power management)

raisjn:
* enable per app power management (default: on, xochitl: off)

tested on rM2 by starting remux, then looking for "POWER MANAGEMENT" line when switching apps


matteodelabre (approved):
Testing successful on rM2, version 2.10.1.332

PR #503, opened 2021-11-27
[rmkit] update remux to 0.13-1

raisjn:
easy diff. this fixes an extra sleep(3) remux had and reverts to sleep(1), see https://github.com/rmkit-dev/rmkit/commit/b810e68e18eeb33b70b6dd37c35f0d6b78e82cc4. 

(this caused it to feel like the remarkable is "slow" to respond to wake event when pressing power from a non-xochitl app)


Eeems (approved):
lgtm

PR #505, opened 2021-11-29
Update xochitl.oxide to support templatectl

Eeems:


pr0fsmith:
Tested.  All is well.

package/xochitl/package
```diff
@@ -5,7 +5,7 @@
 pkgnames=(xochitl)
 pkgdesc="Read documents and take notes"
 url=https://remarkable.com
-pkgver=0.0.0-10
+pkgver=0.0.0-11
 timestamp=2021-11-20T23:31Z
```
LinusCDE:
```suggestion
timestamp=2021-11-29T22:06Z
```

LinusCDE (dismissed):
Tested on rM 2 and works fine. Could add a custom template with templatectl and it showed up when launched through oxide.

You just forgot to update the timestamp. I based the time on your first commit that changed the version.

Eeems:
@LinusCDE updated

raisjn (approved):
based on two people saying it works

PR #510, opened 2021-12-09
Add hint for when the temporary wget or opkg files exist

Eeems:


Eeems:
@matteodelabre poke?

matteodelabre:
Sorry, I was a little bit busy those last few days. This looks good, thanks! A better long term solution would be to store our temporary binaries (wget & opkg) in a `mktemp`-ed directory so that we know for sure they don’t conflict with existing files. I’ll see if I can make a follow-up PR for that soon.

matteodelabre (approved):

PR #512, opened 2021-12-15
Stable merge for week 50 of 2021

Eeems:
Updated packages:
 - xochitl - 0.0.0-11


matteodelabre:
Thanks for creating the PR! Shouldn’t we hold back `rmfakecloud-proxy` until https://github.com/toltec-dev/toltec/pull/448#issuecomment-986156400 is resolved?

Eeems:
Hmm, maybe. I think that's up to you. We probably should put together some extra documentation for the package that could at least just mention the known issues until we resolve them.

matteodelabre:
Yeah, I would prefer waiting for the next merge PR for rmfakecloud-proxy. I opened #513 with the required fixes. I removed the package from this merge PR.

Eeems:
> Yeah, I would prefer waiting for the next merge PR for rmfakecloud-proxy. I opened #513 with the required fixes. I removed the package from this merge PR.

Sounds good! Have you had a chance to test? We can probably merge this PR as soon as we want since the original change is old enough.

matteodelabre:
I just tested writing garbage offsets in `/opt/etc/rm2fb.conf` and Xochitl picked them up while launching from the Oxide chroot. So it seems to work!

matteodelabre:
Oh but that was not the actual thing to test right, this change seems to be about templates rather. What bug does this fix?

Eeems:
> Oh but that was not the actual thing to test right, this change seems to be about templates rather. What bug does this fix?

See the source PR: https://github.com/toltec-dev/toltec/pull/505
templatectl templates were not showing up.

matteodelabre:
Right, thanks for clearing that up! Just tested with some templates and everything seems to be working well.

matteodelabre (approved):

PR #514, opened 2021-12-16
Update ddvk-hacks to 29.01-1

Eeems:


Eeems:
I plan on updating to 2.11 soon, so I'll be able to test on a rM1, but someone else will need to test on a rM2.

Eeems:
Works for me on my rM1. @matteodelabre, @raisjn, or @LinusCDE can you check the rM2?

matteodelabre:
There seems to be a missing file, here’s what I get on rM2 2.11.0.442:
```
Installing ddvk-hacks (29.01-1) to root...
Configuring ddvk-hacks.

Device: reMarkable 2
Xochitl version: 2.11.0.442
Patch version: 29.2.02

Patching Xochitl
bspatch: fopen(/opt/share/ddvk-hacks/patch_29.2.02): No such file or directory
```

Eeems:
We really need a way to automate creation and testing of this package :(

matteodelabre:
> We really need a way to automate creation and testing of this package :(

Yes! The `package()` part can probably be automated to go look for the highest patch version number in each directory. Extracting the metadata used in `configure()` (version number, version timestamp, Xochitl original hash) will probably not be as easy though.

matteodelabre (approved):
Working on rM2 2.11.0.442

Eeems:
> Extracting the metadata used in `configure()` (version number, version timestamp, Xochitl original hash) will probably not be as easy though.

Since you just need to parse it out of patch.sh, it's probably not that difficult.

matteodelabre:
> Since you just need to parse it out of patch.sh, it's probably not that difficult.

Mmh. Parsing shell scripts can be a challenge sometimes. But assuming the patch.sh format stays fixed it should be doable indeed.

PR #517, opened 2021-12-20
Update KOReader to v2021.12

gbyl:
Remarkable specific changes:

- Fix: Middle button launch for reMarkable 1 [koreader-base#1428](https://github.com/koreader/koreader-base/pull/1428) and [koreader#8487](https://github.com/koreader/koreader/pull/8487)

[Release notes](https://github.com/koreader/koreader/releases/tag/v2021.12)


Eeems:
Since we don't include button-listen in our package, that wont actually be included in the package we provide.

gbyl:
@Eeems Good point. I noticed that the timestamp was not updated for https://github.com/toltec-dev/toltec/pull/496. How did you generate it?

Eeems:
> @Eeems Good point. I noticed that the timestamp was not updated for https://github.com/toltec-dev/toltec/pull/496. How did you generate it?

Grabbed it from the KOReader GitHub release

gbyl:
That makes sense. I was confused for a second as a different timestamp is listed in https://build.koreader.rocks/download/stable/v2021.12/

Eeems (approved):

PR #519, opened 2021-12-24
[rmkit] update remux and minesweeper

raisjn:
* add `launch` command to remux, use like: `echo "launch xochitl" > /run/remux.api`
* fix pixmap display in minesweeper so icons show up

### DIFF
--- a/package/rmkit/package
+++ b/package/rmkit/package
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 pkgnames=(bufshot genie harmony iago lamp mines nao remux simple)
-timestamp=2021-11-27T07:33:10Z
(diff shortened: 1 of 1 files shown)


matteodelabre (approved):
Looks good! The minesweeper definitely looks better with the icons :-)

matteodelabre:
Random suggestion: Maybe include a small command to invoke the remux API, to make it more discoverable, at some point in the future?

raisjn:
> Random suggestion: Maybe include a small command to invoke the remux API, to make it more discoverable, at some point in the future?

oh, like `remux-api launch xochitl` or `remux-launch xochitl`? makes sense, thanks for the idea :-D i added the launch command in preparation for evidlo's ask on re-starting xochitl from oxide/remux via API.

PR #530, opened 2021-12-28
Force rm2fb-client to depend on a specific version of display

Eeems:
If people try to just update rm2fb-client without updating display as well, this should force it to pull the updated display package, or raise an error instead.


matteodelabre:
Sounds reasonable. What about the reverse dependency (display -> rm2fb-client)?

Eeems:
> Sounds reasonable. What about the reverse dependency (display -> rm2fb-client)?

Technically not required. Applications can interact with rm2fb without rm2fb-client.

matteodelabre:
But the dependency is already there.

Eeems:
> But the dependency is already there.

Fair enough, in that case we should specify it. Should we also start specifying minimum versions for the xochitl version?

matteodelabre:
Note to self: This needs to be backported to [toltec-dev/build](https://github.com/toltec-dev/build).

matteodelabre:
> Technically not required. Applications can interact with rm2fb without rm2fb-client.

By the way, you have a point there. E.g., VNSee does not need rm2fb-client. The catch is that rm2fb-client also includes the shim for Xochitl, and that needs to be installed anytime display is installed, otherwise we may have two concurrent swtcons running. Perhaps the better option would be to split the Xochitl shim from rm2fb-client, but this is probably out of scope for this PR.

scripts/toltec/version.py
```diff
@@ -145,21 +145,13 @@ def __init__(
...
-                if enum_kind.value == dependency[:colon]:
-                    kind = enum_kind
-                    dependency = dependency[colon + 1 :]
-                    break
-            else:
-                raise InvalidDependencyError(
-                    f"Unknown dependency type \
-'{dependency[:colon]}'"
-                )
+        for enum_kind in DependencyKind:
```
Eeems, line 150: No more raising errors?

matteodelabre: Since there is an ambiguity between dependency types specifiers and epochs, I first check whether any known dependency type is present and otherwise let it fall through as an epoch. If some invalid dependency type is given, it’ll be processed as an epoch, which will also raise an error since epochs need to be numeric.

Eeems: Alright, that's good enough for me. You'll need to re-approve the changes.

matteodelabre: Great! I still need to test that the install process goes smoothly before approving. I’ll do this later in the day.

matteodelabre:
Shall we also increase the package revision number?

matteodelabre:
Some tests: 

#### Manual install of `.ipk` files
Trying to manually `opkg install` a `.ipk` for either `rm2fb-client` or `display` alone now fails with the following error:
```
 * pkg_hash_check_unresolved: cannot find dependency display (= 1:0.0.12-2) for rm2fb-client
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for rm2fb-client:
 *      display (= 1:0.0.12-2)
 * opkg_install_cmd: Cannot install package rm2fb-client.
```
(if both `.ipk`s are specified in the same `opkg install` command, the install is successful)

#### Install by name
Installing either `rm2fb-client` or `display` by name (or through a package that depends on `display`) works as expected.

#### Upgrades
Running `opkg upgrade` upgrades both `rm2fb-client` and `display` to the same version, as before. Running either `opkg upgrade rm2fb-client` or `opkg upgrade display` triggers the upgrade for both packages regardless of which one was specified.

matteodelabre (approved):

PR #532, opened 2022-01-05
Fix remarkable-shutdown calls and add toltec-base package

Eeems:
On the rM1 remarkable-shutdown and remarkable-reboot don't actually display anything anymore on my system. In order to resolve this, and ensure that shutdown and reboot images are properly displayed, I've replaced the calls with remarkable-splash.

This will have to be tested on both the rM1 and rM2.


matteodelabre:
I can confirm I see the same issue on rM1, version 2.11.0.442. However remarkable-shutdown is working fine on rM2 (same version), so I would say that this is likely just an oversight from rM. I’m a bit hesitant to include such a fix in Toltec. Perhaps a better route would be to notify the rM devs of this and get this fixed?

Eeems:
> I can confirm I see the same issue on rM1, version 2.11.0.442. However remarkable-shutdown is working fine on rM2 (same version), so I would say that this is likely just an oversight from rM. I’m a bit hesitant to include such a fix in Toltec. Perhaps a better route would be to notify the rM devs of this and get this fixed?

Next time we see them on we can try to let them know I guess. That said, it's broken right now, so we might as well fix it for our users in a way that works across OS versions. This also gives us options if they ever decide to move the shutdown/reboot image display logic inside xochitl.

matteodelabre:
You make a good point. Are you sure this belongs to the `xochitl` package though? This will make using remarkable-splash mandatory for most users. Would it make more sense to add this to the `remarkable-splash` package itself? Or in a new package?

Eeems:
> You make a good point. Are you sure this belongs to the `xochitl` package though? This will make using remarkable-splash mandatory for most users. Would it make more sense to add this to the `remarkable-splash` package itself? Or in a new package?

Good point. Perhaps a separate fix package similar to the wifi suspend fix?

Eeems:
@matteodelabre poke?

raisjn:
a new package sounds good, let me know if you need review

matteodelabre:
On version 2.12 remarkable-shutdown is still a no-op, so it does look like a deliberate decision from rM.

Eeems:
Okay, I've updated it to use its own package. I've also created a meta package so that we can make it easier to specify packages that need to be installed on the rM1 or rM2. I've added it as a dependency of the xochitl package so that as people upgrade they'll also get it. Eventually we can drop that dependency. I also tweaked toltec-bootstrap to handle installing it.

package/open-remarkable-shutdown/remarkable-reboot-override.conf
```diff
@@ -0,0 +1,6 @@
+[Unit]
+After=launcher.service
+
+[Service]
+ExecStart=
+ExecStart=/opt/bin/remarkable-splash /usr/share/remarkable/rebooting.png
```
matteodelabre: In the stock remarkable-reboot.service definition, the ExecStart entry is prefixed with a dash so that errors from the splash process [are ignored](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=). Should we also do that here?
```suggestion
ExecStart=-/opt/bin/remarkable-splash /usr/share/remarkable/rebooting.png
```

package/open-remarkable-shutdown/remarkable-shutdown-override.conf
```diff
@@ -0,0 +1,6 @@
+[Unit]
+After=launcher.service
+
+[Service]
+ExecStart=
+ExecStart=/opt/bin/remarkable-splash
```
matteodelabre: Same as above.
```suggestion
ExecStart=-/opt/bin/remarkable-splash
```

package/open-remarkable-shutdown/remarkable-reboot-override.conf
```diff
@@ -0,0 +1,6 @@
+[Unit]
+After=launcher.service
```
matteodelabre: Making sure that the splash only comes after the configured launcher service dies seems like a fix that would be important to bring to the rM2 too.

package/open-remarkable-shutdown/package
```diff
@@ -0,0 +1,33 @@
...
+pkgnames=(open-remarkable-shutdown)
+pkgdesc="Use remarkable-splash to display shutdown and reboot images"
+url=https://github.com/ddvk/remarkable-splash
+pkgver=1.0-1
+timestamp=2022-02-28T00:12Z
+section="utils"
+maintainer="Eeems <eeems@eeems.email>"
+license=MIT
+installdepends=(remarkable-splash)
+installdepends_rm2=(display)
```
matteodelabre: Does this package need to depend on `display`, given that `remarkable-splash` already does, and given that this package doesn’t directly use the display?

package/open-remarkable-shutdown/package
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+archs=(rm1 rm2)
+pkgnames=(open-remarkable-shutdown)
```
matteodelabre, line 5: I’m not sure about the package name. Since this will be installed by default when installing Toltec, how about `toltec-shutdown`?

package/remarkable-splash/package
```diff
@@ -10,8 +10,8 @@ timestamp=2019-12-31T10:07Z
 section="utils"
 maintainer="Eeems <eeems@eeems.email>"
 license=MIT
-installdepends=(display)
 flags=(patch_rm2fb)
+installdepends=(display)
```
matteodelabre: Looks like this change is a no-op.

package/toltec-meta/package
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+archs=(rm1 rm2)
+pkgnames=(toltec-meta)
```
matteodelabre: Since this describes the base set of packages for Toltec, how about `toltec-base` as a name? `toltec-meta` is correct but seems a bit too generic to me, as a meta package can really be about anything.

package/xochitl/package
```diff
@@ -5,12 +5,12 @@
...
 pkgdesc="Read documents and take notes"
 url=https://remarkable.com
-pkgver=0.0.0-11
+pkgver=0.0.0-12
 timestamp=2021-11-29T22:06Z
 section="readers"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=MIT
-installdepends=(findutils)
+installdepends=(findutils toltec-meta)
```
matteodelabre: I think it would make more sense to add this transitional dependency to `toltec-bootstrap`, as the `xochitl` package was not previously installed by default (only as a dependency of some launchers).

package/toltec-meta/package
```diff
@@ -0,0 +1,23 @@
...
+pkgdesc="Toltec meta packages to ensure that the required base packages get installed"
+url=https://toltec-dev.org/
+pkgver=1.0-1
+timestamp=2022-02-28T00:12Z
+section="utils"
+maintainer="Eeems <eeems@eeems.email>"
+license=MIT
+installdepends=(toltec-bootstrap toltec-completion xochitl)
```
matteodelabre: If we’re changing `xochitl` to be installed by default, would it make sense to remove it from the installdepends arrays of `display`, `draft`, `reboot-guard`, `tarnish`, and `remarkable-stylus`?

matteodelabre:
Should we also add `wget-ssl` to the set of base packages?

package/open-remarkable-shutdown/remarkable-reboot-override.conf
Eeems: Yes, good catch

package/open-remarkable-shutdown/remarkable-reboot-override.conf
Eeems: True, perhaps it should be added to the xochitl package so that the rM2 gets it by default, and this just overrides it if they want to install it.

package/open-remarkable-shutdown/package
Eeems: Good catch, you would be correct.

package/open-remarkable-shutdown/package
Eeems: It's only installed by default on the rM1 to work around a bug. When the bug goes away we can stop installing it by default. I personally like the name as it's an open source alternative to their shutdown utility.

package/remarkable-splash/package
Eeems: Yeah it was leftover from when I used to have the package part of this one. I can revert it fully.

package/toltec-meta/package
Eeems: Sure! That would be more inline with what entware and others do.

package/xochitl/package
Eeems: While that is true, there have been several users who don't have toltec-bootstrap installed as they installed toltec before it was its own package. This would get more coverage. How about both?

package/toltec-meta/package
Eeems: No, even if something is installed by default. They should still be marked as depends by anything that uses them. That way if you try to uninstall a default package you get proper dependency warnings, and can get it back automatically if you install something that uses it again. Being explicit is not a bad thing here.

Eeems: That said, perhaps it shouldn't be installed by default?

Eeems:
> Should we also add `wget-ssl` to the set of base packages?

Perhaps, although we'd also need to change how bootstrap handles its additional dependencies:

```bash
[[ ! -d /opt/etc/ssl/certs ]] && additional_packages+=(ca-certificates)
[[ ! -f /opt/bin/wget ]] && additional_packages+=(wget-ssl)
```

package/open-remarkable-shutdown/remarkable-reboot-override.conf
matteodelabre: Good idea! I’ll move this to the xochitl package. Perhaps that package would be due for a rename at some point in the future.

package/toltec-meta/package
matteodelabre: That’s a good point. I don’t think we should have it installed by default.

package/xochitl/package
matteodelabre: That sounds good! I added a note above the dependency in both recipes to give context for future readers, and to remind ourselves to remove those dependencies after a while.

matteodelabre:
> Perhaps, although we'd also need to change how bootstrap handles its additional dependencies

Good point, let’s not complicate this PR further then, we can handle this separately.

matteodelabre:
The changes look good, we’ll only need some testing to validate everything works as expected.

package/open-remarkable-shutdown/package
```diff
@@ -0,0 +1,31 @@
...
+    SKIP
+)
+
+package() {
+    install -D -m 644 -t "$pkgdir"/etc/systemd/system/remarkable-shutdown.service.d "$srcdir"/remarkable-shutdown-override.conf
+    install -D -m 644 -t "$pkgdir"/etc/systemd/system/remarkable-reboot.service.d "$srcdir"/remarkable-reboot-override.conf
+}
+
+configure() {
+    systemctl daemon-reload
```
matteodelabre, line 30: I think we also need `systemctl daemon-reload` in postremove

matteodelabre:
There seems to be a bug from our build system: the `open-remarkable-shutdown` dependency gets added to the rM2 version of the base package.

Eeems:
We should probably fix that before merging.

matteodelabre (approved):
Looks good to me!

Tested on rM1 and rM2: on upgrade, toltec-base is automatically installed (on both devices). rM2 picks up rm2-suspend-fix and rM1 picks up open-remarkable-shutdown. On rM1, reboot and shutdown screens are shown properly after `shutdown now` and `reboot`.

matteodelabre:
On rM1, I also tested it with Oxide, making sure the reboot/shutdown images are shown after Oxide quits.

PR #534, opened 2022-01-10
Update plato to 0.9.25

LinusCDE:
Updates plato to the newest version while also finally merging the config fixes into an official release. More [here](https://github.com/LinusCDE/plato/releases/tag/0.9.24-rm-release-11).

Notes regarding the patches: It seems that for some reason, rust nightly from Oct 2021 had some weird issue where even with explicit declaration of the unstable feature `proc_macro_is_available`, it still wont work properly. Hence I vendored the dependencies to just patch out the call for now. This will be fixed as soon as a new [toolchain](https://github.com/toltec-dev/toolchain/) release automatically updates the rust nightly version.

Test cases:
 - Just check that it runs (the issues with reverting configs should be gone).


LinusCDE:
Updated to 0.9.25.

Did basic testing of the standalone version on both devices. For changes see the [release notes](https://github.com/LinusCDE/plato/releases/tag/0.9.25-rm-release-12).

raisjn (approved):
installs for me on rM1, also tested it can open an epub. it's quite fast and snappy!

PR #535, opened 2022-01-12
Stable merge for week 2 of 2022

Eeems:
New Packages:
- rmfakecloud-proxy
- netevent

Updated Packages:
- ddvk-hacks, rm2fb-client - 29.01
- display - 0.0.13
- KOReader - 2021.12.1
- oxide, erode, fret, rot, tarnish, decay, corrupt, anxiety - 2.3
- bufshot - 0.1.1
- harmony - 0.1.4
- mines - 0.1.3
- remux - 0.1.14

### DIFF
--- a/package/ddvk-hacks/package
+++ b/package/ddvk-hacks/package
@@ -6,15 +6,15 @@ archs=(rm1 rm2)
 pkgnames=(ddvk-hacks)
 pkgdesc="Enhance Xochitl with additional features"
 url=https://github.com/ddvk/remarkable-hacks
-pkgver=28.01-1
-timestamp=2021-11-17T16:32:30Z
+pkgver=29.01-1
+timestamp=2021-12-15T19:55:50Z
 section="readers"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=MIT
 flags=(nostrip)
 
-source=(https://github.com/ddvk/remarkable-hacks/archive/10727fe79cb23c5c35ce7a4809cd41b52e6f9384.zip)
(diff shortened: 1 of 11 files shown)


LinusCDE:
All packages tested and run fine on the rM 2.

Only exception is launching xochitl from oxide. This is most likely due to the state of my device though (differing xochitl.oxide file and having used a later version on it before). So should work fine on other devices.

Btw, also really nice to have this new proxy for rmfakecloud. Kinda stopped using it since even running a handmade script that did the host key injection (didn't use the go proxy) was too tedious for me at some point. Once I added a locally-reachable dedicated domain. Worked fine then.  
Only point would be that an unreachable IP makes certain steps hang seemingly forever. The package should probably lower the connect timeout to 3 or 5s for such cases or add some response after that time. The systemd log tipped me off that i initially used my vpn-only domain (have not setup wg on the rm yet).

LinusCDE (approved):
Previous comment was meant as a review :laughing:

Eeems:
@LinusCDE So we just need a final test on the rM1. I've been running rmfakecloud-proxy as my daily driver since it was merged into testing with no issues.

raisjn:
> Only exception is launching xochitl from oxide. This is most likely due to the state of my device though (differing xochitl.oxide file and having used a later version on it before). So should work fine on other devices.

Someone should also verify oxide is working as expected before merge

Etn40ff:
I tested netevent, ddvk-hacks, rm2fb-client, and display on RM2, they work as intended.

LinusCDE (approved):
Tested the packages on my rM 1 as well. Oxide also launched xochitl perfectly here. I would say this is good to go.

PR #541, opened 2022-01-23
Minor opkg completion update

LinusCDE:
Pretty small update just removing one comment.

Did look at some scripts and found that it contained a todo I already pondered and deemed as not useful (linked in the commit description).

I also checked whether the globally disabled spellchecks were now unnecessary (maybe some refactoring) but they are still good.

Not sure about the comments/descriptions of the variables, but since I didn't find any documentation in the [https://github.com/scop/bash-completion], these are probably still helpful for anyone taking a look at it (checked source code as well as explicit docs).

### DIFF
--- a/package/toltec-completion/package
+++ b/package/toltec-completion/package
@@ -5,8 +5,8 @@
 pkgnames=(toltec-completion)
 pkgdesc="Expands bash-completion with functions for toltec-specific commands"
 url=https://github.com/toltec-dev/toltec
-pkgver=0.3.0-1
-timestamp=2021-07-29T22:55Z
+pkgver=0.3.1-1
+timestamp=2022-01-23T23:29Z
 section="utils"
 maintainer="Linus K. <linus@cosmos-ink.net>"
 license=MIT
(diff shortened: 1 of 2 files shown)


package/toltec-completion/package
```diff
@@ -5,8 +5,8 @@
 pkgnames=(toltec-completion)
 pkgdesc="Expands bash-completion with functions for toltec-specific commands"
 url=https://github.com/toltec-dev/toltec
-pkgver=0.3.0-1
-timestamp=2021-07-29T22:55Z
+pkgver=0.3.1-1
```
Eeems, line 8: Since you are just removing a comment, I would probably just increment the package release version, so `0.3.0-2`

LinusCDE: I had already pondered this decision and am somewhat against it, since strictly speaking "the source code" was updated/patched here and not the packaging. But I can also understand your point since the functionality of the software doesn't change with this.

Feel free to decide on your own whether to apply this suggestion.

Eeems: Technically it's up to you since you maintain this package :P

Eeems (approved):

PR #543, opened 2022-01-29
Add package for Folly (z-machine interpreter)

bkirwi:
:wave: This is a new package for the interactive fiction interpreter I've been working on, https://github.com/bkirwi/folly

I've developed it entirely on RM2, though @Eeems was kind enough to test on RM1 a while back. I've checked that the package installs on my tablet, and I've spent quite a lot of time running the software itself, but let me know if there's more specific testing I should take on.

Note that this autocreates the `~/folly/` directory, and adds a few files to it, if it doesn't already exist. Happy to configure or change that if that path seems inappropriate.

I've also made a small quality-of-life change to `builder.py`, printing the hashes that don't match, since that makes spotting errors or updating the file somewhat easier. (And gitignoring `.venv`, since that's not checked in.)


raisjn:
looks awesome, can't wait to try it out! i think setting the FOLLY_ROOT to somewhere in /opt/ makes sense

raisjn (changes requested):
back on your plate for rM1 compatibility

LinusCDE (commented):
On the rM 1, the pen keeps drawing when not touching anymore. So not really working on that model.

This may be what was found on discord already and will probably get fixed by https://github.com/canselcik/libremarkable/pull/97 .

bkirwi:
Yeah, apologies, should have updated this PR when that was found! Agree that the PR you linked is intended to fix... I should get an update in this PR within a day or so.

LinusCDE (changes requested):
Package works well on the rM 1 now! Nice tutorial!

- Please remove the majority of comments from the package file.

Next time it would also be nice to have the `.env` and `builder.py` changes be a separate PR. But I don't mind it this time.

LinusCDE (dismissed):
Nice! Merging..

LinusCDE:
Ah. Seems that raisjn's review is still outstanding.

> i think setting the FOLLY_ROOT to somewhere in /opt/ makes sense

@raisjn, are you considering it okay to merge despite default stories being installed to  `/home/root/folley`? I don't really have a strong opinion regarding this. Your call.

bkirwi:
> Package works well on the rM 1 now! Nice tutorial!

Thanks, and glad to hear it!

> Please remove the majority of comments from the package file.

Done!

> Next time it would also be nice to have the `.env` and `builder.py` changes be a separate PR. But I don't mind it this time.

Noted!

bkirwi:
> Ah. Seems that raisjn's review is still outstanding.

Ah yes, right, sorry I missed it! I do feel a bit weird about using `/opt` for user data, including saves and things. Another alternative I was considering was to use `$HOME/.local/share` like xochitl does. But I notice plato uses a top-level directory in home, so it looks like there may not be a standard yet?

LinusCDE:
Yeah. There is no real standard yet. I put saves/configs of my games into `/home/root/.config/<name>`. Of note is that I never have any of these files in my package, so they're either created by the app or an configure script.

As for plato, I preferred this folder to make it easier for users to find. But I only create it in the configure script and don't have the package maintain that folder (or files in it) directly. Since you have default files, I'm not sure whats best to proceed here. Maybe put these files into a default location (like `/opt/usr/share/folley`) and let the configure script do a symlink to the current location the first time.

Opkg seems to be not that particular about stuff especially if new folders are creates, but having default fails technically be managed by the package manager, might cause some problems if people just delete them later.

Eeems:
> Yeah. There is no real standard yet.

Technically the XDG standard does exist, and we just use `/opt` as the root instead of `/` for any paths outside of `/home` in that standard.

LinusCDE:
> Technically the XDG standard does exist, and we just use /opt as the root instead of / for any paths outside of /home in that standard.

Yeah. But this is usually for configurations. And `/home/root` is easy to find by users while also preventing data loss, should they completely uninstall toltec and therefore whipe everything in `/opt`. And some apps use `/opt/share/` while others use `/opt/usr/share`, so not quite 100% mapped to `/opt`.

Eeems:
> Yeah. But this is usually for configurations.

The XDG standard also covers data storage. https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

> If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used. 

The ArchWiki has a good write-up on it: https://wiki.archlinux.org/title/XDG_Base_Directory#User_directories

LinusCDE:
For end-user friendly ness (who don't necessarily know to look in some hidden subfolder), having it obvious findable when using a file manager is helpful though. My main concern is whether it is good from a packaging standpoint to have files lying around that a user thinks he should delete and the package manager thinks, it alone should manage them.

bkirwi:
> Technically the XDG standard does exist [...]

For sure -- I just meant that it wasn't consistently followed. But it sounds like that's the way to go here... I'll follow up with that change.

> For end-user friendly ness (who don't necessarily know to look in some hidden subfolder), having it obvious findable when using a file manager is helpful though.

Definitely my original motive for the choice! But I appreciate that not everyone wants their home directory cluttered, and I'm optimistic that it won't be too much of an issue in practise.

Thanks everyone for weighing in!

raisjn:
can you clarify what directory for saves you went with?

bkirwi:
> can you clarify what directory for saves you went with?

For sure - all user data (including saves and any other files that the app creates) are now stored under `~/.local/share/folly`.

raisjn (approved):
looks good

PR #552, opened 2022-02-11
Update rm2fb to v0.0.14

Eeems:
* Added support for version 2.11.0.435
* Added support for version 2.12.1.527

PR #545, opened 2022-02-01
Update innernet to 1.5.3

Eeems:
[Release notes](https://github.com/tonarino/innernet/releases/tag/v1.5.3)

PR #542, opened 2022-01-27
[rmkit] update to all applications, add new apps: dumbskull, rpncalc and wordlet

raisjn:
rmkit:

* rmkit went through a large refactoring, so we update all apps
* fix keyboard layout (swapped O and I keys originally)

dumbskull:

* add new game, "Dumbskull", which is a themed solitaire

genie:

* add palm rejection to genie (thakns Ekalfwons)

iago:

* add settings dialog (@aeneas) that allows snapping

mines:

* fix spurious crash when playing multiple games in a row

remux:

* add dialog_height/dialog_width options to remux.conf
* better memory analysis (of all processes in group, not just main process)
* rm1: remove usb check that keeps it powered indefinitely

rpncalc:

* add new app, rpncalc
* add better spacing and layout
* add push after operation behavior (next input is on new stack element)

wordlet:

* add new game, "Wordlet", which is a wordle clone

PR #556, opened 2022-02-18
Stable merge for week 6 of 2022

### LINKED ISSUE #554: Support for Remarkable 2 - 2.12.1.527
Remarkable 2 shipped 2.12 last week, the current build being 2.12.1.527.

It would be great to have an update to Toltec to support this.

raisjn:
New Packages:

* Add new rmkit apps: dumbskull, rpncalc and wordlet (#542)
* Add package for Folly (z-machine interpreter) (#543)

Updated Packages:

* Update KOReader to v2022.01 (#517)
* Update innernet to 1.5.3 (#545)
* Update to all rmkit applications (#542)
* Update plato to 0.9.25 (#534)
* Update rm2fb to v0.0.14 (#552)

Misc:

* Remove outdated todo from opkg completion (#541)


LinusCDE (dismissed):
Tested packages successfully on both models.

~~Sidenote: Since 2.12.1.527 is not supported yet. I had to add [this](https://github.com/ddvk/remarkable2-framebuffer/commit/1bccb9e9a7d4505800c2444e69b8fdab555b7ecf) to `/opt/etc/rm2fb.conf` to make it work.~~

raisjn:
thanks so much for testing!! 

this should contain rm2fb support for latest xochitl (2.12.1.527), if it doesn't we should update this pr

LinusCDE:
> this should contain rm2fb support for latest xochitl (2.12.1.527), if it doesn't we should update this pr

oops. you're right. not sure why but I had configure errors when updating and though they might have missed and it worked afterwards. Might have not applied rm2 specific updates though. Not entirely sure, but removing the config and rebooting works.

matteodelabre:
When starting plato from remux, the app is correctly launched and shows up both on screen and on htop, but it does not appear as started in the remux menu (no memory usage is visible). Switching back and forth between other apps and plato through remux does work. If I tap the blank square beside “Plato” in the menu, where the memory usage would normally appear, Plato is not killed, but the Plato screen is henceforth not recalled properly when switching to another app and back to Plato.

I’m curious if anyone else can reproduce that bug? Does anyone know whether it’s been reported elsewhere already (I couldn’t find any reference from a quick search)?

matteodelabre:
`innernet`’s timestamp was not updated in #545 (correct timestamp is `2022-01-31T20:08:43Z`).

matteodelabre:
`rmkit`’s timestamp says 2021 instead of 2022.

matteodelabre (approved):

matteodelabre:
Tested on rM2 2.12.

PR #560, opened 2022-02-21
Hotfixes for stable merge of week 6

matteodelabre:
* Update innernet timestamp
* Fix rmkit timestamp for apps updated in #542
* Patch the Plato wrapper script to make the process name be
  `/opt/bin/plato`, so that it can be properly recognized by remux


matteodelabre:
Tested the patch for Plato on rM2 2.12, this does make Plato appear correctly in the remux menu and enables it to be killed properly.

raisjn (approved):

PR #561, opened 2022-02-26
rmfakecloud-proxy: Set status check timeout to 5s

### LINKED ISSUE #536: Add connect timeout for rmfakecloudctl
As found here: https://github.com/toltec-dev/toltec/pull/535#issuecomment-1012598588

When an IP is unreachable, wget tried to seemingly forever connect. There should be some sensible connect timeout (3-5s) specified to prevent this.

matteodelabre:
Running `rmfakecloudctl status` with an unreachable server set as upstream would previously hang indefinitely. This PR sets the wget timeout to 5 seconds with no retry. It also changes the `status` command output to give feedback to the user before attempting to reach the server.

Fixes #536.


LinusCDE (approved):
All changes work great!

matteodelabre:
Thanks for the review!

PR #564, opened 2022-02-28
folly: Remove hardcoded paths and values in recipe

matteodelabre:
No substantial changes in this PR, simply some changes in the recipe for `folly` (added in #543) to replace some hardcoded paths to toolchain locations and the hardcoded GCC version with values from the environment (populated by default in the toolchain Docker images).

This makes the recipe more robust to changes in the toolchain images. For example in the 2.3 toolchain update the GCC version was bumped from 10.2.0 to 10.3.0, which made the recipe fail as it was searching for 10.2.0 includes.

I’ve tested that the recipe still builds, installed folly on my rM2, and went through part of the tutorial to make sure that it still works. Pinging @bkirwi: does this look good to you?

### DIFF
--- a/package/folly/package
+++ b/package/folly/package
@@ -5,7 +5,7 @@
 pkgnames=(folly)
 pkgdesc="Z-machine interpreter for interactive fiction"
 url="https://github.com/bkirwi/folly"
-pkgver=0.0.1-1
+pkgver=0.0.1-2
 timestamp=2022-01-29T15:35:52-05:00
 section=games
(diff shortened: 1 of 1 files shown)


bkirwi:
This looks good - I think I follow everything well enough to continue maintaining it, and your testing sounds quite sufficient. Thanks for taking this on!

bkirwi (approved):

raisjn (approved):

matteodelabre:
Thanks!

PR #565, opened 2022-03-01
keywriter: Update to latest version

### LINKED ISSUE #544: Update keywriter to latest version
Would be great to have this rm keywriter fork with smaller font size for edit mode:
https://github.com/danschrage/remarkable-keywriter

There have been few more people who wanted to have a writing experience with smaller font size.

matteodelabre:
* Reduce font size in edit mode
* Rotate the buffer using Ctrl-Left/Ctrl-Right in view mode
* Add Ctrl-Q shortcut to quit the app
* Restore cursor position when switching between modes

Fixes #544

### DIFF
--- a/package/keywriter/package
+++ b/package/keywriter/package
@@ -5,24 +5,24 @@
 pkgnames=(keywriter)
 pkgdesc="Markdown-enabled free writing app"
 url=https://github.com/dps/remarkable-keywriter
-pkgver=0.1.0-3
-timestamp=2019-07-13T06:27Z
+pkgver=0.2.0~ddc9e73-1
+timestamp=2022-02-09T07:05:03Z
 section="writing"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=MIT
 installdepends=(display)
 flags=(patch_rm2fb)
 
-image=qt:v2.1
+image=qt:v2.3
 _sundown=37728fb2d7137ff7c37d0a474cb827a8d6d846d8
 source=(
-    https://github.com/dps/remarkable-keywriter/archive/c77ec3f65d9ff769f3f5dc85cc91abbf05aa163f.zip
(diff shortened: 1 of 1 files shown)


LinusCDE (approved):
Tested on the rM 1. Works well!

Not sure if this is intentional, but the update also changed the app to be in portrait orientation.

Edit: Ah, yeah, with rotating the buffer this just seems to be the new default but can easily be changed with CTRL+Right Arrow as described.

CTRL+Q also works.

package/keywriter/package
```diff
@@ -5,24 +5,24 @@
 pkgnames=(keywriter)
 pkgdesc="Markdown-enabled free writing app"
 url=https://github.com/dps/remarkable-keywriter
-pkgver=0.1.0-3
-timestamp=2019-07-13T06:27Z
+pkgver=0.2~20-gddc9e73-1
```
LinusCDE, line 8:
```suggestion
pkgver=0.2~20-ddc9e732-1
```

I think the commit sha1 short has a type here.

LinusCDE (changes requested):
Overlooked this. Seems there is a typo.

package/keywriter/package
matteodelabre: I’ll admit it’s a bit strange, but this is the format generated by running `git describe --tags` in the keywriter repo. Not sure why git adds the `g` prefix here.

LinusCDE: Ah, cool, didn't even know this was [a thing](https://stackoverflow.com/questions/23939214/why-does-git-describe-prefix-the-commit-id-with-the-letter-g). Learned something new I guess.

LinusCDE (approved):

matteodelabre:
Thanks for the review!

PR #567, opened 2022-03-05
xochitl: Fix missing launcher.service alias

### LINKED ISSUE #558: Enabled launcher state is lost after upgrade and reenable
In normal circumstances, when enabling a launcher app to start at boot, a symlink is created by systemd from `/etc/systemd/system/launcher.service` to the said service. This helps prevent having multiple launchers start at boot and compete for display access, and is used by Toltec’s helper `how-to-enable` to inform users that they need to disable another launcher before enabling a new one.

After upgrades, the systemd state gets overwritten as it’s located on the system partition. In particular, Xochitl reverts to its enabled state and the `launcher.service` symlink is not recreated, even after `toltecctl reenable` is run. This causes a state where both Xochitl and a launcher can be enabled simultaneously, and post-install messages after installing a launcher are incorrect.

matteodelabre:
This PR contains some fixes for the xochitl package:

* On configure, force creating the launcher.service symlink if Xochitl was previously enabled (closes #558). After removal, force cleaning the launcher.service if Xochitl is enabled.
* Remove unused dependency on `findutils`.
* Fix the following warning in `xochitl` wrapper script that would appear in Xochitl logs when there are no files in /opt/etc/xochitl.env.d (such as by default on rM1):

```
/opt/bin/xochitl: line 8: /opt/etc/xochitl.env.d/*.env: No such file or directory
```

### DIFF
--- a/package/xochitl/package
+++ b/package/xochitl/package
@@ -5,8 +5,8 @@
 pkgnames=(xochitl)
 pkgdesc="Read documents and take notes"
 url=https://remarkable.com
-pkgver=0.0.0-12
-timestamp=2021-11-29T22:06Z
+pkgver=0.0.0-13
+timestamp=2022-03-05T23:26Z
 section="readers"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=MIT
@@ -14,7 +14,7 @@ license=MIT
 # below as a transitional dependency to ensure that it gets installed for
 # existing users (see toltec-dev/toltec#532 for context). This dependency
 # should be removed once the package has reached enough users
-installdepends=(findutils toltec-base)
+installdepends=(toltec-base)
 
 source=(
     xochitl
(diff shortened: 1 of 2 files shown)


raisjn:
how does this play with remux?

Eeems:
@raisjn good point, remux assumes that the service is running right? Perhaps xochitl shouldn't alias to launcher, and instead oxide and draft should just explicitly conflict with xochitl?

raisjn:
remux will stop xochitl service if it is running and launch xochitl directly, so it doesn't technically require xochitl to be running. i was just wondering if there will be any cases where this is a problem since it is fine to have remux and xochitl enabled at the same time. will this PR essentially cause remux and xochitl to conflict?

matteodelabre:
This PR doesn’t change any relationships between the services—xochitl and remux services don’t conflict, since [remux is not aliased to `launcher.service`](https://github.com/toltec-dev/toltec/blob/stable/package/rmkit/remux.service). This PR is only to address the fact that the xochitl service is left in an inconsistent state after `toltecctl reenable` (with `xochitl.service` enabled but `launcher.service` not enabled). After this PR, it will still be fine to enable both xochitl and remux at the same time (but it won’t be to enable xochitl and tarnish, for example).

raisjn (approved):
image

PR #569, opened 2022-03-14
Kernelctl

Etn40ff:
This is a utility to manage aftermarket kernels.


Eeems:
Does it make sense to force bz2 compression as the only option? It might make more sense to allow multiple compression algorithms to be used, like xz, gz, zstd.

Etn40ff:
I do not see the benefit in allowing different compression formats: there are no preexisting kernels that we need to accommodate for so we might as well ask for a specific format. Moreover dealing with multiple formats would add a small complexity to the script.

Eeems:
Fair enough. The next question is why bz2 instead of one of the other formats?

Etn40ff:
Higher compression rate compared to gz. I do not really have a preference.

Eeems:
If the compression ratio is what we want, and we aren't concerned about decompression speed, xz would be the better option. If we care about decompression speed while still having a good compression ratio, zstd would be better.

Etn40ff:
I am not sure compression rate is such a concern: the backup of the vendor kernel is currently around 5.6M. 
Compression time is indeed quite slow with bz2.
zstd is not currently available in the official firmware but we could depend on the toltec version.

Etn40ff:
Here is a testing kernel you can use to check that kernelctl works. It is for RM2. It is basically the same as the current upstream kernel with `uinput` enabled and `TRIM_UNUSED_KSYMS=n`.
edit: the kernel here had an issue with RDINS

GH is just dumb: I can attach a `.gz` but not a `.bz2` so please just rename it to `vendor_patched-20220315.tar.bz2` before using it.

PS: if there is enough interest in having the upstream kernel with few modifications in toltec I might easily be persuaded to make a package.

Eeems:
@Etn40ff I'm on a rM1 so I can't test, but it looks like the automated checks are complaining about the script not following the style guide. You'll need to resolve that before it can be merged.

Etn40ff:
~~It should be fixed now.~~ Oops, now the linter complains. This will take longer, I'll have a look at it tomorrow.

As for the file format, should I switch to zstd or xz?

Kernelctl should in theory work as well on a rm1. Actually it need testing there but I do not own a rm1. Could you do it? Do you need me to provide a testing kernel for you? Should I make a toltec package with the modifications I use on my rm2 (uevent, tun/tap, and `TRIM_UNUSED_KSYMS=n`)?

Etn40ff:
Automated checks should be fine now

Etn40ff:
I am unhappy of the current status:

Currently this script passes the list produced by `tar tjf current_kernel_backup.tar.bz2`  to `rm -rf` to clear up before installing the new kernel. The flags `-rf` are needed because a directory might be passed to rm before it has been emptied and it would give an error.

Obviously this is an easy recipe for disaster: if the kernel backup is not properly formatted one could delete something that should not be deleted. (I just erased the whole `/lib` this way and I was not a happy camper.) Any suggestion on a better way to delete all files and directories (if empty) from a list?

Etn40ff:
I might have answered my own question but better solutions are welcome

package/kernelctl/kernelctl
```diff
@@ -0,0 +1,191 @@
...
+    if [[ ! -e "$current_kernel" ]]; then
+        rear -r -d '' msg <<- EOM
+			${bf}There is no copy of the current running kernel in the staging
+			area, you might have accidentally deleted it!${sf}
+			Try making a backup first
+		EOM
+        echo -e "$msg"
+        exit 1
+    fi
+    tar tjf "$current_kernel" | sort -r | xargs rm -d 2> /dev/null
```
Eeems: Do we care if `tar tjf "$current_kernel"` doesn't return any values?
```suggestion
    tar tjf "$current_kernel" | sort -r | xargs -r rm -d 2> /dev/null
```

Eeems: Also, what does `rm -d` do? I'm not seeing it in any help [manual](https://www.busybox.net/downloads/BusyBox.html)/output.

Etn40ff: > Do we care if `tar tjf "$current_kernel"` doesn't return any values?

No but I guess an extra safety will not hurt. A am also considering removing `2> /dev/null`, this will help to identify badly formatted tarballs when reviewing packages.

Etn40ff: > Also, what does `rm -d` do? I'm not seeing it in any help [manual](https://www.busybox.net/downloads/BusyBox.html)/output.

It should remove empty directories but not full one. But as you point out this is not implemented in busybox. Ouch, I'll fix it later.

Etn40ff:
Here is a better kernel for testing on RM2: I messed up the previous one.

This is vanilla + uinput + tun - TRIM_UNUSED_KSYMS
[vendor_patched-20220329.tar.bz2.zip](https://github.com/toltec-dev/toltec/files/8366783/vendor_patched-20220329.tar.bz2.zip)

package/kernelctl/kernelctl
```diff
@@ -0,0 +1,191 @@
...
+    echo -e "  ${bf}$(< $kernelctl_status)${sf}"
+}
+
+# change the kernel that will boot next time
+set() {
+    new_kernel_name=$(to_kernel_name "$1")
+    new_kernel="$kernelctl_dir/${new_kernel_name}.tar.bz2"
+    current_kernel="$kernelctl_dir/$(< $kernelctl_status).tar.bz2"
+    if [[ ! -e "$current_kernel" ]]; then
+        rear -r -d '' msg <<- EOM
```
gbyl:
```suggestion
        read -r -d '' msg <<- EOM
```

alistair23:
Any update on this? I'd love to get https://github.com/toltec-dev/toltec/pull/488 merged

Etn40ff:
> Any update on this? I'd love to get #488 merged

I pinged toltec's team over Discord few days ago, they said that they wanted to focus on getting a stable release out first. That happened already so, hopefully, they will have some spare time soon.

alistair23:
Any updates?

luvanir:
Tested on RM1 with linux-stracciatella. kernelctl works as advertised. Thank you!

package/kernelctl/kernelctl
```diff
@@ -0,0 +1,219 @@
...
+        read -r -d '' msg <<- EOM
+			${bf}WARNING: failed to extract the new kernel!${sf}
+			Maybe there is not enough space left on the root partition?
+			Attempting to revert to the current running kernel.
+		EOM
+        echo -e "$msg"
+        if switch "$new_kernel_file" "$current_kernel_file"; then
+            /bin/sync
+            read -r -d '' msg <<- EOM
+				Successfully reverted to the current rnning kernel.
```
gbyl:
```suggestion
				Successfully reverted to the current running kernel.
```

package/kernelctl/kernelctl
```diff
@@ -0,0 +1,219 @@
...
+				Please check that everything is in order before rebooting.
+			EOM
+            echo -e "$msg"
+        else
+            read -r -d '' msg <<- EOM
+				Unable to revert to a working configuration.
+				If you do not know how to proceed please seek assistence
```
gbyl:
```suggestion
				If you do not know how to proceed, please seek assistance
```

package/kernelctl/kernelctl
```diff
@@ -0,0 +1,219 @@
...
+				in reMarkable's server on Discord.
```
gbyl:
```suggestion
				in reMarkable's Discord server: https://discord.gg/ATqQGfu
```

package/kernelctl/kernelctl
```diff
@@ -0,0 +1,219 @@
...
+    fi
+    if switch "$current_kernel_file" "$new_kernel_file"; then
+        rm -f "$current_kernel_file"
+        ln "$new_kernel_file" "$current_kernel_file"
+        /bin/sync
+        echo "Reboot system to use your newly installed kernel."
+    else
+        read -r -d '' msg <<- EOM
+			${bf}WARNING: failed to extract the new kernel!${sf}
+			Maybe there is not enough space left on the root partition?
```
gbyl:
```suggestion
			This may be due to a lack of space in the root partition.
```

gbyl (changes requested):

package/kernelctl/kernelctl
Eeems: It's the community discord server, it's not official.

Etn40ff:
```suggestion
				in the reMarkable's community Discord server: https://discord.gg/ATqQGfu
```

package/kernelctl/kernelctl
Etn40ff: Space may not be the only issue making this fail, no?

gbyl: Good point! I adjusted the warning to take that into consideration as well.

Eeems (approved):
Works for me on my rM1, further changes can be new PRs

PR #572, opened 2022-03-17
[rmkit] update harmony and remux

raisjn:
harmony: support layers
remux: use less memory for storing app snapshots

Test Plan:

harmony:

* load harmony, play with layers: add layer, rename, show, hide, export, etc. export png, import png, etc

remux:

* load remux, swap between different apps and verify the screen looks correct


matteodelabre (approved):
Tested on rM2 2.12.2.

PR #575, opened 2022-03-24
Update KOReader to v2022.03.1

Eeems:
[Release notes](https://github.com/koreader/koreader/releases/tag/v2022.03.1)
Don't use KOReader suspend/screensaver/autosuspend when Oxide is running

### DIFF
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -5,7 +5,7 @@
 pkgnames=(koreader)
 pkgdesc="Ebook reader supporting PDF, DjVu, EPUB, FB2 and many more formats"
 url=https://github.com/koreader/koreader
-pkgver=2022.03-1
+pkgver=2022.03.1-1
 timestamp=2022-03-20T08:31:59Z
 section="readers"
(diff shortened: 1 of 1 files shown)


matteodelabre:
@Eeems On rM1, when KOReader is started from Oxide, after putting the device to sleep and resuming it, or after pausing KOReader and then going back to it from Oxide, there seems to be a screen depth issue when the last screen is recalled (fixes itself after interacting with KOReader).

Eeems:
Oh, it might be a bug with Oxide since it's not aware of screen depth and just changes it as part of pause/resume/start/stop events.

matteodelabre:
I see. This should not be blocking for merging this PR then.

matteodelabre (commented):
Tested on rM1 and rM2 2.12.3, with a focus on testing that the KOReader sleep screen does not show up when KOReader is started from Oxide, and that it is triggered when KOReader is not started from Oxide.

matteodelabre (approved):

PR #581, opened 2022-04-02
Linux stracciatella

Etn40ff:
This is my take on the default kernel for both rm1 and rm2.

Sources are exactly those provided by RemarkableAS. 

Configs differs from theirs as follows:

- Removed version name modifications
- TRIM_UNUSED_KSYMS=n
- UINPUT=y
- TUN=y

It requeres `kernelctl` and it should be a safe kernel to test its
functionalities.


Etn40ff:
Closed: it was based on stable rather then testing.
Sorry for the noise

PR #582, opened 2022-04-02
Stracciatella kernel

Etn40ff:
This is my take on the default kernel for both rm1 and rm2.

Sources are exactly those provided by RemarkableAS.

Configs differs from theirs as follows:
- Removed version name modifications
- UINPUT=m
- TUN=m
- PERF_EVENTS=y

It requeres kernelctl and it should be a safe kernel to test its
functionalities.

### DIFF
--- a/package/linux-stracciatella/package
+++ b/package/linux-stracciatella/package
@@ -7,45 +7,50 @@ pkgnames=(linux-stracciatella)
 pkgdesc="RemarkableAS's vanilla kernel with a few extra flakes"
 url=https://github.com/Etn40ff/linux-remarkable
 pkgver=5.4.70-1
-timestamp=2022-04-02T18:36:09+02:00
-section="devel"
+timestamp=2022-06-26T23:50:04+02:00
+section="kernel"
 maintainer="Salvatore Stella <etn45p4m@gmail.com>"
+makedepends=(build:flex build:bison build:libssl-dev build:bc build:lzop build:libgmp-dev build:libmpc-dev build:kmod)
 license=GPL-2.0-only
 flags=(nostrip)
 installdepends=(kernelctl)
-makedepends=(build:flex build:bison build:libssl-dev build:bc build:lzop build:libgmp-dev build:libmpc-dev build:kmod)
-image=base:v2.2
-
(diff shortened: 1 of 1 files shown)


package/linux-stracciatella/package
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+# Copyright (c) 2022 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+archs=(rm1 rm2)
+pkgnames=(linux-stracciatella)
+pkgdesc="RemarkableAS's vanilla kernel with few extra flakes"
```
gbyl:
```suggestion
pkgdesc="RemarkableAS's vanilla kernel with a few extra flakes"
```

package/linux-stracciatella/package
```diff
@@ -0,0 +1,51 @@
...
+
+package() {
+    pushd "$srcdir"/out
+    tar cpjf ../stracciatella-"$pkgver".tar.bz2 lib/modules/* boot/*
+    popd
+    install -D -m 644 -t "$pkgdir"/opt/usr/share/kernelctl "$srcdir"/stracciatella-"$pkgver".tar.bz2
+}
+
+configure() {
+    echo -e "Please kernelctl to select the kernel to boot."
```
gbyl:
```suggestion
    echo -e "Please use kernelctl to select the kernel to boot."
```

gbyl (changes requested):

Etn40ff:
This kernel requires firmware blobs that were introduced in version 2.9. Now we print a warning when installing on an older version.

luvanir:
RM1 owner here. linux-stracciatella works wonders! The only caveat is that you need to have at least version 2.9 of the remarkable OS before you switch away from the vanilla kernel. Dear maintainers, let me know if there are any tests you would like me to run before merging. Thank you for the help troubleshooting, Salvatore.

Eeems:
@Etn40ff did you test this on a rM2?

Eeems:
Tested on my rM1 as well and this works. I'd like someone to confirm that they tested on a rM2 and that it works.

Etn40ff:
> @Etn40ff did you test this on a rM2?

It has been my daily drive since April.

Eeems (approved):

PR #599, opened 2022-06-24
Update KOReader to 2022.06

Eeems:
[Release notes](https://github.com/koreader/koreader/releases/tag/v2022.06)

- Fix reMarkable crash bug v2022.05: event overwrite with new time module


LinusCDE (approved):
Works fine and fixes the crash with rM 2 (tested on both models with FW 2.14.1.866).

It would be pretty useful to have this in the next stable release (#601) as well, if you don't mind.

PR #600, opened 2022-06-26
Packages for the Linux perf profiler

matteodelabre:
This package provides the `perf` Linux profiler, which interfaces with the kernel perf events subsystem and tracing features. See <https://perf.wiki.kernel.org/> for details.

Using `perf` requires a kernel with `CONFIG_PERF_EVENTS` enabled, not enabled on the default reMarkable kernel. Use the `linux-mainline` package, for example.


Eeems (dismissed):
Tested on my rM1 with the Stracciatella kernel and it works

Eeems:
@matteodelabre did you test on a rM2 with linux-mainline? If so, this can probably be merged.

matteodelabre:
Since this package is built from @alistair23’s sources, I think it makes more sense to move this as a split package inside the existing `linux-mainline` recipe. It will ensure we keep both in sync.

matteodelabre:
Tested on rM2 (2.15.1.1189) with linux-mainline (6.0.0) and linux-stracciatella (5.4.70).

alistair23 (dismissed):

package/linux-mainline/package
```diff
@@ -3,14 +3,11 @@
 # SPDX-License-Identifier: MIT
 
 archs=(rm2)
-pkgnames=(linux-mainline)
-pkgdesc="reMarkable 2 kernel based on the mainline kernel"
-url=https://www.kernel.org
+pkgnames=(linux-mainline linux-mainline-perf)
 pkgver=6.0.0-1
```
Etn40ff: version bump?

package/linux-stracciatella/package
```diff
@@ -3,17 +3,14 @@
 # SPDX-License-Identifier: MIT
 
 archs=(rm1 rm2)
-pkgnames=(linux-stracciatella)
-pkgdesc="RemarkableAS's vanilla kernel with a few extra flakes"
-url=https://github.com/Etn40ff/linux-remarkable
+pkgnames=(linux-stracciatella linux-stracciatella-perf)
 pkgver=5.4.70-2
```
Etn40ff: version bump?

package/linux-mainline/package
matteodelabre: I don't think it's necessary to do a version bump, since the built `linux-mainline` and `linux-stracciatella` packages are unchanged, and the `*-perf` packages are new.

Etn40ff: I thought that we used the last part in the version number to track changes to the build system. Either way works for me.

matteodelabre: We should use version bumps whenever we need users to receive a new version of a package. The last part is useful when there are changes to the package but the upstream version has not changed.

package/linux-mainline/package
```diff
@@ -21,35 +18,54 @@ sha256sums=(1583d553380a656ecaac8dab89558051e9165983381adc9c3d52fb8ac0ce19fd)
...
-    echo "The new kernel files have been copied, but not installed."
-    echo "Please use kernelctl to select the kernel to boot."
-    echo
-    echo "Known issues with the mainline kernel:"
-    echo " - No support for low power mode (suspend uses more power then it should)"
-    echo " - WiFi sometimes is off on boot (can be turned on again though)"
-    echo " - GUI shutdown in Oxide doesn't work"
-    echo " - Wacom stylus doesn't work in Xochitl (works everywhere else though)"
-    echo " - No OTG control support"
+linux-mainline-perf() {
```
Eeems, line 74: Shouldn't this conflict with `linux-stracciatella-perf` since they provide the same files?

Eeems (changes requested):

package/linux-mainline/package
matteodelabre: Good point.

alistair23 (approved):

Eeems:
@matteodelabre I don't think this was waiting on anything further, right? Was this fully tested? I'm good to merge it if it's good to go.

Eeems:
Oh, looks like the build is broken when fetching sources.

Eeems (dismissed):

PR #598, opened 2022-06-23
ddvk-hacks: Update to v34.01 (support for OS 2.14)

matteodelabre:
Tested on rM1 and rM2 2.14.0.861

PR #593, opened 2022-06-06
Bump ddvk 2.13.0.758

Etn40ff:
This bumps ddvk's hacks to rm1: 2.12.3.606 and rm2: 2.13.0.758.

For rm1 version 32.1.02 has been superseeded by 32.1.3.

I am still on 2.12.3.606 so I can't test this PR.

PR #592, opened 2022-06-06
Bump to the new rm2fb release

bkirwi:
Just a version bump to pick up: https://github.com/ddvk/remarkable2-framebuffer/releases/tag/v0.0.17

Re: testing - I've been running with these values in rm2fb.conf for some time with no issues.

PR #588, opened 2022-05-06
[rmkit] update remux and harmony

raisjn:
remux: palm filtering config option. defaults to off
harmony: layer dialog improvements, better HW support for pressure/tilt

test plan:

* start remux, verify "ENABLED PALM EVENT FILTERS" is not in the initial output. 
* harmony: play with layers, import/export, etc

PR #601, opened 2022-06-27
Stable merge for week 26 of 2022

Eeems:
### New packages
- `linux-mainline` - 5.18.0-1 (#488)
  - Mainline linux kernel for the rM2 only

### Updated packages

- `koreader` - 2022.06-1 (#590 #599)
  - Fix reMarkable crash bug v2022.05: event overwrite with new time module
- `display` `rm2fb-client` - 1:0.0.20-1 (#592 #597 #604)
  - Added support for OS 2.13.0.758
  - Added support for OS 2.14.0.861
  - Added support for OS 2.14.1.866
- `ddvk-hacks` - 35.01-1 (#593 #598 #605)
  - Support OS 2.12.3.606 for rM1
  - Support OS 2.13.0.758 for rM2
  - Support OS 2.14.0.861 for both rM1 and rM2
  - Supprot OS 2.14.1.866 for both rM1 and rM2
- `folly` - 0.0.1-3 (#587)
- `remux` - 0.2.3-1 (#588)
  - Make palm filtering a config option
- `harmony` - 0.2.1-1 (#558)
  - More layer dialog improvements
  - Sort import dialog by file date
  - Refactor pressure / tilt events to use hardware values
- `wireguard` - 1.0.20210219-3 (#488)
  - Move to kernel section
- `fuse` - 1.0.0-4 (#488)
  - Move to kernel section


raisjn:
thanks for opening! can you also include the PRs that matteo accepted last week? 

* https://github.com/toltec-dev/toltec/commit/f17336fa607809d02084f2b7c87e47feb1a60730
* https://github.com/toltec-dev/toltec/commit/de887138ec78c5748c79f8febf5e6402f171ccbf
* https://github.com/toltec-dev/toltec/commit/b671b5af88d42557aebb77e4835130f1cb7cbbb2

Eeems:
@raisjn Since they were only merged 5 days ago, I was holding off on this first stable merge. @matteodelabre any concerns with pulling those forward as well?

matteodelabre:
Thanks for the PR. Yes, I think it would be worthwhile to include the commits that raisjn mentioned.

Eeems:
@matteodelabre @LinusCDE @raisjn poke

LinusCDE (changes requested):
Tested most packages on both models except:
 - The Linux related packlages (`linux-mainline`, `fuse`, `wireguard`) as I don't want to dive into that rn. Maybe later
 - ddvk-hacks (I think I already had this version installed prior and updated to 35.x from PR #605 for this update for testing and because I got this fw as well).

Regressions:
 - ⚠️ harmony@rm2: Hides the toolbar for some brushes (e.g. "Paint brush", which causes hitching when touching the pen. The rM 1 is not affected. More an inconvenience, I don't deem that problematic.
 - ❗ koreader@rm2: Crashed when touched by either pen or finger -> Unusable ([Screenshot of crash report](https://transfer.cosmos-ink.net/gUGhJB/192.168.2.png)). Runs fine on the rM 1. Seems like #599 would fix that. Will test that one next and we should probably include that to fix this issue.

I manually added the entry for the rm2fb to make it run on the 2.14.1.866 FW. We really need some solution that can just update the `/opt/etc/rm2fb.conf` file with newer entries to reduce delays and the need to release and test package a bunch until users can have the improvement.

Eeems:
There is a PR for 2.14.1.866 rm2fb support already open to testing. I agree about just editing the file instead of waiting for a new release. There are issues open on the rm2fb repo with ideas to simplify things, they just need to be implemented.

LinusCDE:
> There is a PR for 2.14.1.866 rm2fb support already open to testing. [...]

Right. Just noticed that. Tested and merged it.

> [...] I agree about just editing the file instead of waiting for a new release. There are issues open on the rm2fb repo with ideas to simplify things, they just need to be implemented.

Yeah. Just wanted to point it out again. At this point it seems toltec on the rM 2 is only useful when being 1-2 fw releases behind because of the delayed releases. Or one knows how to get the update advance with said file. But this is probably the minority. Maybe I'll find time for it next. I think, we should just include the procedure in toltecctl / bootstrap and either parse the config.cpp file or pull an extra released standalone version of it.

Eeems:
@LinusCDE I've pulled in the latest commits for 2.14.1.866 support

LinusCDE (dismissed):
Retested, koreader works now fine and ddvk and rm2fb work out-of-the box.

Nice work, @Eeems !

I didn't really follow the kernel and wireguard things. If someone can confirm those to work, this should be good to go.

Eeems:
@matteodelabre would you be able to test the final pieces?

matteodelabre:
Sure, I can take a look! Sorry for the delay.

matteodelabre:
(Note: I fixed the `display` version number in your original PR comment.)

Tested on rM1 and rM2, 2.14.0 and 2.14.1.

* `linux-mainline` currently instructs users to make a backup of the stock kernel files to the `/opt/boot-backup` directory, but this directory will get removed by `toltecctl uninstall`. We should change this to recommend backing up to a folder under `/home/root` instead. (#606)
* Can confirm the issue mentioned by @LinusCDE with `harmony` on rM2. This makes it a bit difficult to draw (ping @raisjn)
* `fuse` and `wireguard` do not require testing as they are just being moved to a different section.

Everything else is working correctly as far as I can tell.

Eeems:
Fuse and wireguard will require testing with the mainline kernel and the latest OS kernel at some point to make sure they work properly. They've been broken for a while but the latest OS kernel should start working again.

Etn40ff:
> Fuse and wireguard will require testing with the mainline kernel and the latest OS kernel at some point to make sure they work properly. They've been broken for a while but the latest OS kernel should start working again.

Fuse may give issues with linux-mainline since it already has `CONFIG_FUSE_FS=y`. 

Maybe we should re-think the whole idea of making packages with single modules
because it may lead to a lot of confusion whit alternative kernels. Moreover installing things in /lib/modules is definitely going to make `kernelctl` much more complicated than already is.

matteodelabre:
It should be possible to make package variations for each kernel that Toltec provides, and for example mark `fuse` as conflicting with the mainline kernel/depending on the stock kernel.

Etn40ff:
> It should be possible to make package variations for each kernel that Toltec provides, and for example mark `fuse` as conflicting with the mainline kernel/depending on the stock kernel.

Not really a straightforward path: suppose you install linux-mainline but do not enable it, or possibly temporarily decide to switch to the upstream kernel: you will have to install/uninstall the package just to switch. Moreover, how is `kernelctl` going to know which files in `/lib/modules/` belong to the kernel and which to the package? I mean, it is possible but not in a clean way.

Eeems:
I'm not sure the conflict will mater too much since modprobe would still load fuse properly since it's included in the kernel no"?

Etn40ff:
I never tried adding an off-tree kernel module on top of a kernel that already provides it, I do not know how it would react. Possibly just with a warning.

matteodelabre (approved):
Looks good to me. Thanks again @Eeems for your work on this!

PR #602, opened 2022-06-28
Remove parent directories

Etn40ff:
Remove some parent directories from the tarball so that kernelctl will not attempt to delete them when switching kernels.

### DIFF
--- a/package/ddvk-hacks/package
+++ b/package/ddvk-hacks/package
@@ -6,15 +6,15 @@ archs=(rm1 rm2)
 pkgnames=(ddvk-hacks)
 pkgdesc="Enhance Xochitl with additional features"
 url=https://github.com/ddvk/remarkable-hacks
-pkgver=33.02-1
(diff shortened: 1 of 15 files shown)


matteodelabre:
Maybe `kernelctl` should not delete directories after all? Directories don’t use significant space anyway, and other things might get installed by other packages into those directories, as mentioned in #601.

Etn40ff:
> Maybe `kernelctl` should not delete directories after all? Directories don’t use significant space anyway, and other things might get installed by other packages into those directories, as mentioned in #601.

Ideally I wold agree but, unfortunately, we have pretty stringent disk space constraints to work with and having even two sets of kernel modules consumes a sensible amount of space. For instance linux-stracciatella, which basically is upstream, uses 2.7MB in /lib/modules while linux-mainline uses 7.4MB for a total just above 10MB.

Eeems:
@Etn40ff it looks like there are no outstanding changes in this PR, I'm closing it.

PR #603, opened 2022-07-02
Update Oxide to v2.4

Eeems:
[Release notes](https://github.com/Eeems-Org/oxide/releases/tag/v2.4)

- Opt-in crash reporting
- Created a shared library to provide an easier API for applications to use
- Fixed issue where launcher will crash if a notification is removed with the API instead of through the UI
- Fixed issue where the task switcher background will disappear if you close an application
- Fixed issue where applications will be imported even though the file isn't executable, and thus would fail to launch

### DIFF
--- a/package/oxide/package
+++ b/package/oxide/package
@@ -9,7 +9,7 @@ maintainer="Eeems <eeems@eeems.email>"
 url=https://oxide.eeems.codes
 license=MIT
 flags=(patch_rm2fb)
-image=qt:v2.1
+image=qt:v2.3
 source=("https://github.com/Eeems-Org/oxide/archive/e6c62a6860b52cd1cbd47e21377f8d1ecf72bb0a.tar.gz")
 sha256sums=(a29cf455d5f66fee4ee67722c6f1d20627930920286398932c3d5c813d58ebee)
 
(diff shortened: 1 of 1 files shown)


LinusCDE (changes requested):
Installation fails because of version mismatch (top is rM 1 and bottom rM 2):

grafik

The packages all require `libsentry` with exactly version 0.4.13 but the actual package in this PR is version 0.4.17. I ensured to install the `libsentry` first to prevent installation order issues in this regard.

Eeems:
@LinusCDE That's what I get for not installing the packages on my device, since I'm using development builds constantly. That should have fixed it.

LinusCDE (approved):
Tested and works on both version.

Sorry for the late re-review. Opkg acted up and somehow cached the fact that the erode 2.4-1 file was of version 2.3-*? Didn't manage to find the proper cache. A reinstall of toltec fixed it and oxide seems to run fine now. :+1:

LinusCDE:
@LinusCDE That's what I get for not installing the packages on my device, since I'm using development builds constantly. That should have fixed it.

No problem. Is fixed. I usually do a last ipkg test before publishing a PR. But as my updates don't change stuff in the packages, I now tend to just use the generated ipkgs from the PR, test-install them and then add them to a release.

PR #604, opened 2022-07-06
Update rm2fb to v0.0.20

Eeems:
Added support for 2.14.1.866


LinusCDE (approved):
Works fine!

Tested on rM 2 without a `/opt/etc/rm2fb.conf` file (asuming rM 1 packages to be dummies as usual).

Would be super nice to squeeze this along with #605 into the next stable release to support the fw fully.

PR #605, opened 2022-07-09
Bump ddvk to 35.01

Etn40ff:
I could not test this because I do not have 2.14.1.866 yet


LinusCDE (approved):
Tested on both models with the newest fw and works fine!

PR #606, opened 2022-07-12
linux-mainline: Improve backup instructions

matteodelabre:
The manual install/uninstall instructions for the `linux-mainline` package will be superseded by kernelctl (#569), but in the meantime I believe it would be safer to instruct users to make their backup outside of the Toltec root, since otherwise it will get deleted upon uninstall.


Etn40ff:
LGTM but I have no reviewer status so someone else should approve

Eeems (approved):
lgtm

PR #607, opened 2022-07-16
Update rmfakecloudctl to support dev server

Eeems:
If you have `ServiceEnvironment` set to `Dev`, the current rmfakecloud-proxy configuration generated by rmfakecloudctl will not work. You'll need to add`webapp-legacy-dev.cloud.remarkable.engineering` to the proxy list.


Eeems:
Hmm, it looks like this allows you to connect, but you get stuck in a loop.
image
image

matteodelabre:
Any logs on the server side when this loop occurs?

Eeems:
The second image is the server side logs, it just loops through that.

StarNumber12046:
also adding dev.tectonic.remarkable.com and dev.internal.cloud.remarkable.com to the /etc/hosts and CA cert seems to make it function properly.

StarNumber12046:
#932 would add full support

Eeems:
@StarNumber12046 I've merged your changes, but you mentioned that there are still issues with it?

StarNumber12046:
> @StarNumber12046 I've merged your changes, but you mentioned that there are still issues with it?

It seems like, in some cases (only once for about a minute today), it got stuck on periodically refreshing the  decicetoken because of dev.ping.remarkable.com/v2/events (which is not exactly part of the cloud)

Eeems:
> > @StarNumber12046 I've merged your changes, but you mentioned that there are still issues with it?
> 
> It seems like, in some cases (only once for about a minute today), it got stuck on periodically refreshing the decicetoken because of dev.ping.remarkable.com/v2/events (which is not exactly part of the cloud)

That should still be part of rmfakecloud. I would recommend opening an issue on that repository to get it resolved.

PR #608, opened 2022-07-22
rmfm: Add version 1.3.0

TheLastZombie:
Since I couldn't find one, I wrote a (relatively) simple file manager called rmFM. It can be used for viewing, moving, copying, renaming and deleting files and directories.

I tested it on my reMarkable 2 (including the built IPK file). Since it uses Node.js and rmkit's simple app script, it should work on the rM 1 as well, but I can't test it since I don't own one.


Eeems (approved):
Looks good to me, and it works well on my rM1

PR #609, opened 2022-07-31
Update KOReader to v2022.07

Eeems:
There are no reMarkable specific changes.

[Release notes](https://github.com/koreader/koreader/releases/tag/v2022.07)


LinusCDE (approved):
Runs fine on the rM 2. I don't imagine any changes that could only negatively affect the rM 1 here.

PR #612, opened 2022-08-03
Add package: lf

gbyl:
I've been using [lf](https://github.com/gokcehan/lf) for a while and thought others might find it useful as well. Tested on rM1 and rM2.

### DIFF
--- a/package/lf/package
+++ b/package/lf/package
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# Copyright (c) 2022 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(lf)
+pkgdesc="Terminal file manager"
+url=https://github.com/gokcehan/lf
+pkgver=r27-1
+timestamp=2022-04-02T09:40Z
+section="utils"
+maintainer="gbyl <gbyl@users.noreply.github.com>"
+license=MIT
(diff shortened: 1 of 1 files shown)


package/lf/package
```diff
@@ -0,0 +1,25 @@
...
+section="utils"
+maintainer="gbyl <gbyl@users.noreply.github.com>"
+license=MIT
+
+image=golang:v2.3
+source=("https://github.com/gokcehan/lf/archive/refs/tags/r27.zip")
+sha256sums=(a4f7b3ada4aa1348b7f102374d8580b6992977f7e84053aa04ef6aadb69dc205)
+
+build() {
+    # export GOARCH=arm64
```
Eeems: If this isn't needed it should be removed

Eeems (changes requested):

Eeems:
image
I get this error when attempting to run it on my rM1.

package/lf/package
gbyl: I mistankenly pushed an earlier version of this recipe. I'll submit the fix now.

gbyl:
Thank you for the review @Eeems. I forgot to push my last commit. It should work now.

Eeems (approved):
Looks good to me, it works on my rM1

PR #613, opened 2022-08-03
Update yaft to v0.0.7 and add rM1 input patch

gbyl:
I saw this was requested in the discord server. Tested on rM1 and rM2.


LinusCDE (approved):
Nice addition! Tested package on both models and indeed noticed, that the old version didn't work on the rM 1 and is now fixed with this. :+1:

PR #614, opened 2022-08-06
Add package: ncdu

gbyl:
The C-based [ncdu](https://dev.yorhel.nl/ncdu) version distributed via entware has stalled for months. This PR brings the latest fixes and optimizations of the LTS version to toltec. Tested on rM1 and rM2.

### DIFF
--- a/package/ncdu/package
+++ b/package/ncdu/package
@@ -29,3 +29,7 @@ package() {
 configure() {
     ln -s /lib/libncursesw.so.5 /lib/libncurses.so.5
 }
+
+postremove() {
+    rm /lib/libncurses.so.5
+}
--- a/package/ncdu/package
+++ b/package/ncdu/package
@@ -18,18 +18,11 @@ sha256sums=(810745a8ed1ab3788c87d3aea4cc1a14edf6ee226f764bcc383e024ba56adbf1)
 
 build() {
     export CC=arm-linux-gnueabihf-gcc
+    export LDFLAGS="-static"
     ./configure --host=armv7
     make
 }
 
 package() {
     install -D -m 755 "$srcdir"/ncdu "$pkgdir"/opt/bin/ncdu
-}
-
-configure() {
-    ln -s /lib/libncursesw.so.5 /lib/libncurses.so.5
-}
-
-postremove() {
-    rm /lib/libncurses.so.5
 }
--- a/package/ncdu/package
+++ b/package/ncdu/package
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Copyright (c) 2022 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(ncdu)
+pkgdesc="Disk usage analyzer with an ncurses interface"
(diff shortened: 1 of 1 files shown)


LinusCDE (commented):
Tested to work on the rM 1.

On the rM 2 it fails to launch for me due to this error:

grafik

The entware provided package was 1.16-1 if I remember correctly. After removing this version (1.17-1 built in GitHub CI), a `opkg install ncdu` actually installed the same version for me.

grafik

Not sure how this happens, but seems that entware has the newest version, but doesn't advertise it?!? Furtermore this version launches fine on my rM 2. So maybe this PR has some configurations missing?

It seems that an update to 1.17-1 may happen eventually and the orignal package at least for me, doesn't seem to be awefully out-of-date. I guess if we ever fail to update more frequent then [entware](https://bin.entware.net/armv7sf-k3.2/Packages.html), we could just remove our version and have update fall back to entware again.

The package file seems fine to me (apart for somehow not launching on my specific rM 2). I'm just not sure whether it is worth the hassle of mainting a minor version update. Judging from the [website](https://dev.yorhel.nl/ncdu), this has seemingly minor improvements over 1.16.

grafik

The newest version actually seems to be 2.1.2 which seems to now use the zig programming language.

@gbyl 
- Would it be likely that the 1.X series will get updated in the future given that it seems to be pretty separate to the new 2.X versions?
- How do they differ and why not go with 2.X?
- Do you have any idea why it might not work for me on the rM 2?

@matteodelabre @Eeems Do any of you have an idea, why the package only falls back to 1.17-1 after manually installing this build? Maybe we can fix something here, or the most recent version might be available in a bit anyway.

Not sure how to go with this package.

**Edit:** I think I used ncdu in the past already and quite like the package. I would love to see it added. However since entware seemingly already has this release, the weirdness with rM 2 as well as the entware package of this version kinda existing and 2.X being the actual latest version, I'm not sure how to proceed with this.

gbyl:
>Do any of you have an idea, why the package only falls back to 1.17-1 after manually installing this build? Maybe we can fix something here, or the most recent version might be available in a bit anyway.

It appears that in-between my submitting this PR and your review, Entware actually updated ncdu (along with many other packages). I just confirmed this by running the usual `opkg update` and `opkg upgrade`.

>Would it be likely that the 1.X series will get updated in the future given that it seems to be pretty separate to the new 2.X versions?

It's hard to say, but the 1.X series should be usable for a very long time. There are also interesting alternatives inspired by ncdu, such as [gdu](https://github.com/dundee/gdu).

>How do they differ and why not go with 2.X?

I couldn't get the Zig-based 2.X series to compile and run on the rM. How they differ is detailed [here](https://dev.yorhel.nl/doc/ncdu2).

>Do you have any idea why it might not work for me on the rM 2?

I can't say for sure. I may have customized my device in such a way that it happens to work. I will try on a fresh install next time.

>Not sure how to go with this package.

Given the Entware update, there's no point in duplicating efforts. I'll go ahead and close this PR. Thank you for the thorough review!

PR #628, opened 2022-09-10
display: Update to v0.0.23 (support for 2.14.3.1005)

fidanoski:

### DIFF
--- a/package/display/package
+++ b/package/display/package
@@ -23,7 +23,7 @@ source=(
     rm2fb-preload.env
 )
 sha256sums=(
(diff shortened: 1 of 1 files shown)


fidanoski:
I had wrong sha256 sum the first time, because I based it on my previous commit instead of original commit. Updated that and rebased so that it's not polluted with commits. Clean and ready for review now.

fidanoski:
Tested on RM2 version: 2.14.3.1005

raisjn (approved):

PR #630, opened 2022-09-26
linux-stracciatella: merge latest changes to upstream

Etn40ff:
Just keep linux-stracciatella synced to rmMarkable/linux


Eeems:
This works on my rM1. Can you confirm that you tested on a rM2? I'm concerned about how close to full the root partition is after installing this, it would actually be full if I hadn't [moved my logs to the data directory](https://discord.com/channels/385916768696139794/385921709649297409/968207357437943830). There are also old copies of the vanilla kernel that would be nice for kernelctl to clear up in the future.

Eeems:
image

Eeems (approved):

PR #631, opened 2022-09-28
add python package based on prebuilt tgz

SomethingGeneric:
Hi! I hope this an acceptable addition to toltec, rather than entware.
Obviously I'm not the author of Python :sweat_smile: 
This should:tm: work for both devices.

### DIFF
--- a/package/python/package
+++ b/package/python/package
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(python)
+pkgdesc="The Python programming language"
+url="https://python.org"
+pkgver=3.10.7
+_relver=0.0.3.3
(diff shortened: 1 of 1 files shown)


SomethingGeneric:
I apologize also if there are rules for wether to build in-tree with docker, but containers are far above me :sweat:

Eeems:
We do prefer building in-tree. There is a tool you can use to help with testing package recipes called toltecmk that you can install from pypi. You can find more info in the toltec-dev/build repo.

You can also look at other recipes in the repo for examples on how to do various builds.

Eeems:
Oh, and what would the reason be to have python in toltec instead of using the one that comes from the entware repositories?

SomethingGeneric:
> Oh, and what would the reason be to have python in toltec instead of using the one that comes from the entware repositories?

Mfw I didn't see that entware has python3. I guess the GUI package manager didn't fuzzy search correctly. Nevermind 😅

PR #632, opened 2022-09-28
Make a backup at install time only if it is not there yet

Etn40ff:
Currently if one installs kernelctl twice with the same version of codex `configure` will fail. 

This PR fixes this behaviour

### DIFF
--- a/package/kernelctl/package
+++ b/package/kernelctl/package
@@ -5,7 +5,7 @@
 pkgnames=(kernelctl)
 pkgdesc="Manage aftermarket kernels"
 url=https://toltec-dev.org/
-pkgver=0.1-1
+pkgver=0.2-1
 timestamp=2022-03-14T00:00Z
 section="utils"
 maintainer="Salvatore Stella <etn45p4m@gmail.com>"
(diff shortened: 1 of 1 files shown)


Eeems:
Just an FYI, but the build will not create new packages if the version number doesn't change.

Eeems (changes requested):
image

Eeems:
Running the script manually produces the following:
image

package/kernelctl/package
```diff
@@ -27,5 +27,15 @@ package() {
 }
 
 configure() {
-    kernelctl backup vanilla
+    if [[ ! -e "/opt/usr/share/kernelctl/vanilla-$(< /etc/version)" ]]; then
```
Eeems:
```suggestion
    if [[ "$(kernelctl list | tail -n +2 | awk '{print $2}' | grep "vanilla-$(< /etc/version)")" == "" ]]; then
```

Etn40ff: LGTM, sorry for the omission

Eeems (approved):

PR #636, opened 2022-10-01
kernelctl: add prune

Etn40ff:
Add a prune function to kernelctl

### DIFF
--- a/package/kernelctl/package
+++ b/package/kernelctl/package
@@ -5,7 +5,7 @@
 pkgnames=(kernelctl)
 pkgdesc="Manage aftermarket kernels"
 url=https://toltec-dev.org/
-pkgver=0.1-2
+pkgver=0.1-3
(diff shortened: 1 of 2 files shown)


package/kernelctl/package
```diff
@@ -27,7 +27,7 @@ package() {
 }
 
 configure() {
-    if [[ "$(kernelctl list | tail -n +2 | awk '{print $2}' | grep "vanilla-$(< /etc/version)")" == "" ]]; then
+    if [[ ! -e /opt/usr/share/kernelctl/vanilla-"$(< /etc/version)".tar.bz2 ]]; then
```
Eeems: I personally would prefer to use kernelctl so that the package doesn't have to change if you start using multiple archive versions instead of just `.tar.bz2`

Etn40ff: OK

Eeems (approved):
Worked for me!

Issue #553, opened 2022-02-15
Check if user has autoupdate turned on


pbouffard:
Not sure if it's exposed somehow in the OS but given that an auto update of the RM2 OS can lead to suddenly losing the ability to use unofficial software, it would be great if on install/update of toltec that it check if the user has auto-update turned on and give a warning or perhaps even turn it off if that's possible.

If not at least a very conspicuous message somewhere in the install flow to direct the user to do this would go a long way.

matteodelabre:
Thanks for the suggestion. I believe auto-updates are enabled if and only if the `update-engine` systemd service is enabled, so it should be possible to check this in a script. We could even suggest the `systemctl mask update-engine` way of permanently disabling auto-updates.

matteodelabre:
Solved by #641.

PR #641, opened 2022-11-04
Disable auto update on reenable/install

Eeems:
- Disable `update-engine.service` on install/reenable
- Decouple `sync.service` and `xochitl.service` as this means that either sync doesn't work, or xochitl starts when it is disabled.
  - Accomplished by adding a `manual-sync.service` and masking `sync.service`
  - `/opt/bin/xochitl` will also run `sync --service` in the background manually if running from a chroot and there is no sync service running
- Add `kill-xochitl` command as oxide stop events do not allow arguments


Eeems:
Untested

package/kernelctl/package
```diff
@@ -5,7 +5,7 @@
 pkgnames=(kernelctl)
 pkgdesc="Manage aftermarket kernels"
 url=https://toltec-dev.org/
-pkgver=0.1-2
+pkgver=0.1-4
```
Eeems, line 8: There is another PR with `-3` that will likely get merged before this one is.

Eeems:
I can't test the `toltec-base` install because it's failing to fetch `open-remarkable-shutdown`. Even when I try to install that from a ipk file, it still fails. Not sure what is happening.

Eeems:
Okay, I've reinstalled toltec and this seems to have resolved it. This seems to work for me now, but someone will need to test/confirm on a rM2.

LinusCDE (dismissed):
Works fine

Eeems:
I guess sync should be unmasked on uninstall?

LinusCDE:
Wouldn't hurt I guess. Best to restore the original state on uninstall.

LinusCDE (approved):
Works fine. :+1:

matteodelabre:
@Eeems Did this resolve #553?

Eeems:
Most likely

PR #644, opened 2022-11-08
Basic code changes

Peter015:
Pretty simple, combined an if statement into one line and a with statement into one line.


Eeems:
Since there isn't anything functionally different, and this doesn't do anything other than cause code churn, I'm not sure if I want to accept this PR. If it was alongside something more meaningful in the same area, sure.
@matteodelabre ?

matteodelabre:
I concur. I appreciate you taking time to contribute @Peter015, but we would favor functional changes over purely stylistic ones.

PR #653, opened 2022-11-20
Update all actions to latest

Eeems:
This will get rid of all the warnings on builds we have.


matteodelabre:
Looks good to me. From a quick glance at the changelogs, there does not seem to be breaking changes in these actions, apart from using Node 16 instead of 12.

matteodelabre (approved):

PR #664, opened 2023-02-06
Add package: webinterface-wifi

rM-self-serve:
Add package: webinterface-wifi

### DIFF
--- a/package/webinterface-wifi/package
+++ b/package/webinterface-wifi/package
@@ -35,23 +35,20 @@ package() {
 
 configure() {
     systemctl daemon-reload
-
-    if ! is-enabled "$pkgname.service"; then
-        printf "\nRun the following command to use $pkgname\n"
-        printf "systemctl start $pkgname\n\n"
-        printf "To automatically start the application after restarting, run:\n"
-        printf "systemctl enable $pkgname\n\n"
-    fi
+    
+    echo ""
+    echo "Run the following command to use $pkgname"
+    how-to-enable "$pkgname.service"
 }
 
 preremove() {
-    if is-active $pkgname; then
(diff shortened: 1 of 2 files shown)


package/webinterface-wifi/package
```diff
@@ -0,0 +1,60 @@
...
+
+configure() {
+    systemctl daemon-reload
+
+    if ! is-enabled "$pkgname.service"; then
+        printf "\nRun the following command to use $pkgname\n"
+        printf "systemctl start $pkgname\n\n"
+        printf "To automatically start the application after restarting, run:\n"
+        printf "systemctl enable $pkgname\n\n"
+    fi
```
Eeems:
```suggestion
    how-to-enable "$pkgname.service"
```

Eeems (approved):
Works for me. It would be good to add some sort of application to start/stop the service in the future. Right now users will have to use yaft to start/stop to temporarily use it.
Having a SSID whitelist would also be an option, that or putting a password in front of it so it can stay on all the time?

rM-self-serve:
Thanks again for the help! I definitely plan to add whitelist/blacklist functionality. Going to look into some existing packages for the best way to handle a config file and will add some command line options to configure it. I'm thinking of defaulting to a whitelist containing only the SSID of the wifi network that webinterface-wifi is installed/configured on. I would also prefer the application update the config rather than the service file when changing a port.

Eeems:
I really recommend following the XDG standards. Config files should either be in `/opt/etc` or `~/.config`. It would probably be better to put it in `/opt/etc` since it's installed as a service that doesn't run for a specific user session, and it's meant to be modified by the user. I'd also recommend adding a small cli for modifying the settings live without having the restart the service.

PR #665, opened 2023-02-10
Add package: neofetch

rM-self-serve:
This should be tested on the rM2,  I have tested it on the rM1 versions 2.10, 2.15, and 3.1. The OS information and battery read out are the primary concerns.


Eeems:
Works fine for me on my rM1, but I'll hold off on approving and merging until we can get someone to verify on a rM2.

shaunsingh:
Works fine on rM2 as well over both ssh and yaft terminal

Eeems (approved):

PR #681, opened 2023-04-27
Update KOReader to v2023.04

Eeems:
There are no reMarkable specific changes.

See the [release notes](https://github.com/koreader/koreader/releases/tag/v2023.04) for more information.

### DIFF
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -44,6 +44,11 @@ package() {
 }
 
 configure() {
+    # This file can cause issues with startup when moving from 2022.08 to another version
+    if [ -f /opt/koreader/cache/fontlist/fontinfo.dat ]; then
+        rm /opt/koreader/cache/fontlist/fontinfo.dat
+    }
+
     systemctl daemon-reload
 
     if ! is-enabled "$pkgname.service"; then
--- a/package/koreader/package
+++ b/package/koreader/package
@@ -5,8 +5,8 @@
 pkgnames=(koreader)
 pkgdesc="Ebook reader supporting PDF, DjVu, EPUB, FB2 and many more formats"
 url=https://github.com/koreader/koreader
-pkgver=2023.03-1
-timestamp=2023-03-20T11:38:17Z
+pkgver=2023.04-1
+timestamp=2023-04-27T20:53:54Z
 section="readers"
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=AGPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


LinusCDE (commented):
Runs fine on the rM 2, but on the rM 1 I received the following crash:

grafik

I made sure to update toltec on both devices before but didn't update the firmware. The rM is currently on firmware 2.15.1.1189 which is probably pretty old. If that is the case and it runs on modern firmwares well, feel free to merge anyway @Eeems .

Eeems:
Based on https://github.com/koreader/koreader/issues/9771 it looks like your filesystem has some corruption that needs to be cleaned up. Try deleting the cached data files and see if it will work

LinusCDE:
Deleting `/opt/koreader/cache/fontlist/fontinfo.dat` did indeed solve the issue. Seems indeed like some corruption that causes zstd to try to allocate an insane amount of memory (far more than installed ram).

LinusCDE:
Fyi just tested whether it's random corruption or not.

For that I downgraded Koreader to a release from roughly September 2022 and copied back the file (`/opt/koreader/cache/fontlist/fontinfo.dat`, which I only moved it to `/tmp` instead of deleting it) and started that.

```bash
opkg install koreader_2022.08-1_rmall.ipk --force-downgrade
```
https://transfer.cosmos-ink.net/KQCerPfqRa/koreader_2022.08-1_rmall.ipk

That version started fine. When then upgrading the the newest release, it started crashing again. The file has been last changed on Aug 31 2022 according to ls
```
reMarkable: ~pkgs/pkgs1/ ls -lah /tmp/fontinfo.dat
-rw-rw-rw-    1 root     root       34.4K Aug 31  2022 /tmp/fontinfo.dat
reMarkable: ~pkgs/pkgs1/
```

So this is likely not some corruption or unexpected crash which aborted writing the file, but the inability of never koreader versions to properly read the (older version?) of this file.

Below the fontinfo.dat that crashes, as well as a different one where there was no issue for comparison.

- File in question that causes crashes on a recent version: https://transfer.cosmos-ink.net/ofHeNebpbN/fontinfo.dat
- Similar file of other device model, which worked fine: https://transfer.cosmos-ink.net/BNTt5jaVgk/fontinfo.dat (the file appears to have been replaced / upgraded automatically when running the newer version)

LinusCDE (approved):
With the new workaround, upgrading from 2022-08 to this release (with the old problematic file from before) works fine.

Thanks for investigating and fixing it @Eeems! :+1:

PR #682, opened 2023-04-29
Update rmFM host

TheLastZombie:
I'm migrating to a self-hosted [Forgejo](https://forgejo.org/) instance, so the URLs are changing.

Edit: ~~Doing some more maintenance during the migration, should be finished by this evening.~~ Should be stable now.


Eeems:
@TheLastZombie it looks like you opened this against the stable branch. Could you re-open this based off of the testing branch. I'm closing the PR for now.

PR #688, opened 2023-05-08
Update restream to 1.2

Eeems:
[Release notes](https://github.com/rien/reStream/releases/tag/1.2.0)


LinusCDE (approved):
Installed and tested on both models and seem to work fine (firmware on both might be very outdated though). :+1:

PR #691, opened 2023-05-09
kernelctl: reset to default kernel when uninstalling

Etn40ff:
As discussed on Discord. 
Not my favourite solution.

### DIFF
--- a/package/kernelctl/package
+++ b/package/kernelctl/package
@@ -31,3 +31,10 @@ configure() {
         kernelctl backup vanilla
     fi
 }
+
+preremove() {
+    if [[ "$(kernelctl show | tail -n 1 | grep "vanilla-$(< /etc/version)")" == "" ]]; then
+        echo "Restoring default kernel"
+        kernelctl set default
+    fi
+}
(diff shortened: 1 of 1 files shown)


Eeems:
Welp I somehow managed to brick my rM1 testing this. I'm going to guess that somehow the kernel backup I had was broken somehow from one of the upgrades. I'll recover and try again.

Eeems:
The bad news is that I couldn't fix it by copying over the original kernel files. Which means I'm not entirely what broke it. The good news is, I just reflashed the OS, and I'm upgrading again to 2.15.

Eeems:
Alright, this isn't great
image

Eeems:
Normal workflow of `opkg remove linux-stracciatella kernelctl` with stracciatella as my active kernel works though. I think it just needs to be slightly smarter to straight up refuse if the kernel backup is missing. It should also `exit 1` if the restore fails for any other reason. The user can always use force flags with `opkg` if they really don't care for some reason.

Eeems (changes requested):
As mentioned, let's properly catch failures, and detect if the vanilla kernel is missing for some reason.

Eeems (approved):

PR #692, opened 2023-05-10
Provide better output on toltecctl uninstall

Eeems:
- Provide output right at the end if a removal fails due to a preremove script. That way, the user can double-check the output to see if there is something bad. For example, kernelctl failed to revert to the stock kernel.
- Make changes to toltecctl to get closer to allowing the installation to be somewhere other than `/opt`, not that entware supports that yet.
- Make toltecctl uninstall handle unexpected states like partial uninstall a lot better.
- Backup toltecctl on uninstall to continue allowing users to uninstall if something goes wrong.
- Fix issues with installdepends not properly being set for several packages that caused them to be removed after their dependencies were removed, thus breaking uninstall.


rM-self-serve:
I created a test package that will fail during preremove.
Do you have any tips on restoring /opt after uninstalling?

Eeems:
> I created a test package that will fail during preremove.
> Do you have any tips on restoring /opt after uninstalling? 

Uninstall removes /opt, which is actually /home/root/.entware. you will have to install again with bootstrap. If you want to avoid doing so, you can copy the .entware folder before doing the uninstall, rename the copy to .entware after uninstalling and then re-enable.

Jayy001 (dismissed):

Eeems:
image
image

I think this will need to be updated to remove toltec packages first, and then it might even be safe to just delete things without removing entware packages, as they never touch the root partition.

Eeems:
It also probably highlights that launcherctl needs to depend on whatever provides `/opt/bin/find`.

Eeems:
image
launcherctl is being removed after xochitl, which means it is failing to run its uninstall and enable xochitl.

package/toltec-bootstrap/toltecctl
```diff
@@ -460,7 +463,7 @@ reinstall-root() {
         if [[ -v "on_root_packages[$pkgname]" ]]; then
             reinstall_packages[$pkgname]=1
         fi
-    done < <(gunzip -c /opt/var/opkg-lists/* | grep "^Package:" | awk '{print $2}')
+    done < <(gunzip -c $toltec_dest/var/opkg-lists/* | grep "^Package:" | awk '{print $2}')
```
Havner: I don't think anyone would install toltec in a path with whitechars, but it's always wise to put paths from env variables inside ""

package/toltec-bootstrap/toltecctl
```diff
@@ -541,9 +545,9 @@ generate-opkg-conf() {
 # then run \`toltecctl generate-opkg-conf\` to regenerate this file
 
 dest root /
-dest ram /opt/tmp
-lists_dir ext /opt/var/opkg-lists
-option tmp_dir /opt/tmp
+dest ram $toltec_dest/tmp
```
Havner, line 550: same in those 3 lines and few cases below

Eeems: These lines are being written to a config file, so wrapping them inside `"` would not do what is expected.

package/toltec-bootstrap/toltecctl
Eeems: I'm surprised our linter missed this, it catches all the other ones.

package/toltec-bootstrap/toltecctl
Havner: I'm not sure what you mean, those are arguments to a binary or more likely some function (dest/lists_dir/etc). The "" would not be passed there, they would simply make sure that a param is a single param and whitespace wouldn't split it ($2/$3/etc)

package/toltec-bootstrap/toltecctl
```diff
@@ -465,7 +465,7 @@ reinstall-root() {
         if [[ -v "on_root_packages[$pkgname]" ]]; then
             reinstall_packages[$pkgname]=1
         fi
-    done < <(gunzip -c $toltec_dest/var/opkg-lists/* | grep "^Package:" | awk '{print $2}')
+    done < <(gunzip -c "$toltec_dest"/var/opkg-lists/* | grep "^Package:" | awk '{print $2}')
```
Havner, line 468: And some cases below ;-)

Havner: Line 692/693, and possibly others. Too tired to catch all now.

Eeems: Just did a full pass, they should all be covered now.

package/toltec-bootstrap/toltecctl
```diff
@@ -806,6 +808,14 @@ uninstall() {
     }
     set -o functrace
     trap 'handle-error ${LINENO}' ERR
+    handle-exit() {
+        if ! $completed && $success && $clean; then
```
Havner: completed vs complete, wrong variable names

package/toltec-bootstrap/toltecctl
Havner: Not sure what's happened here, you didn't reply, marked as resolved yet did not change it. So either you assumed I'm wrong or you forgot something.

package/toltec-bootstrap/toltecctl
Eeems: You seem to be reviewing an outdated version, this is `$complete` for me when I look

Havner: Ok, I'm reviewing patches in order, not a full diff. I'm more used to a workflow where a PR is force updated instead of incremental patches and a squash afterwards.

Havner: Also I think there is operator order issue here:

```
$ export VAR1=true
$ export VAR2=true
$ export VAR3=true
$ if ! $VAR1 && $VAR2 && $VAR3; then echo failed; fi
$ export VAR2=false
$ if ! $VAR1 && $VAR2 && $VAR3; then echo failed; fi
$ if ! ($VAR1 && $VAR2 && $VAR3); then echo failed; fi
failed
```

Eeems: Since they are declared as local it could be that they aren't working as expected. Although when I tested this originally it was working as I expected.

Havner: I don't think local makes any difference here, `!` has higher priority than `&&` hence it applies only to VAR1 and from the look of it you wanted to apply it to the whole expression (V1 && V2 && V3)

Eeems: I would have wrapped it in `(` and `)` if I was trying to apply it to all of them, I'm checking if not complete, and success and clean. success/clean are true by default and get changed to false when something else handles reporting an error.

Havner: Ah, ok then. From reading the code it looked to me like you are looking for a failure scenario (e.g. at least one of those vars is false).

Eeems: In this case, I'm only wanting to handle the error if the error handler is triggered and we did not complete the uninstall, and nothing set success or clean to false due to their own error handling.

It is less than ideal, thus the extra comments on the variable declarations. I couldn't really come up with a cleaner way to handle this with the limitations bash has on error handling.

Havner: I would have to analyze this fully to check the logic completely. I'm too tired now.

What I did find though is that you make a dep between `clean` and `success` so checking both in this line might be at least redundant.

https://github.com/toltec-dev/toltec/blob/Eeems-patch-10/package/toltec-bootstrap/toltecctl#L885

But I'm not able to check the logic fully now.

PR #687, opened 2023-05-08
Update fbink, fbdepth, and fbink-doom to 1.25.0

Eeems:
[Release notes](https://github.com/NiLuJe/FBInk/releases/tag/v1.25.0)

PR #686, opened 2023-05-08
Don't enable rm2fb if it fails to start

Eeems:
This is to try to keep the device from ending up in a bad state when users install it on an OS version that isn't supported.

PR #670, opened 2023-03-12
[rmkit] update rmkit, harmony, simple, remux and rpncalc

raisjn:
harmony: fix individual layer bug
remux: suspend API
rpncalc: more hp48 like
simple: add range widget, add canvas widget, make images clickable

PR #693, opened 2023-05-13
Stable merge for week 19 of 2023

Eeems:
### Updated Packages
- `koreader` - 2023.04-1 (#672 #681)
- `display`, `rm2fb-client` - 0.0.31-1 (#680 #686)
  - If rm2fb fails to start on install, disable it to avoid soft-bricking the device
  - Add support for the following OS versions:
    - 3.2.2.1581
    - 3.2.3.1595
- `wordlet` - 0.0.2-2 (#670)
  - Swap M and N on keyboard
  - Fix image thumbnail code
- `harmony` - 0.2.2-1 (#670)
  - Fix exporting individual layers
- `simple` - 0.2.0-2 (#670)
  - Add "range" input widget
  - Add "canvas" widget
  - Less debug output
  - Make images clickable
  - Move 'timeout' to its own thread instead of jamming task queue
  - Ignore empty lines
- `remux` - 0.2.4-1 (#670)
  - Add "suspend" to remux FIFO api
- `rpncalc` - 0.0.3-2 (#670)
  - Make it more like the hp48
- `toltec-base` - 1.2-1 (#685)
  -  Only disable sys-subsystem-net-devices-usb1.device if on a reMarkable 1. On a reMarkable 2, this makes SSH over usb no longer available.
  - Make toltec-base properly clean itself up on uninstall
- `kernelctl` - 0.1-6 (#691)
  - Reset to the default kernel on uninstall so that `toltecctl uninstall` will return the device to a stock state
- `restream` - 1.2-1 (#688)
- `fbink`, `fbdepth`, and `fbink-doom` - 1.25.0-1 (#687)

**Note:** This doesn't change what OS version that toltec supports, as full support still requires `ddvk-hacks` to be updated, and for proper testing of various other packages.

### DIFF
--- a/package/display/package
+++ b/package/display/package
@@ -4,11 +4,11 @@
 
 archs=(rm1 rm2)
 pkgnames=(display rm2fb-client)
-timestamp=2023-02-20T15:49:05Z
+timestamp=2023-04-16T20:53:38Z
 maintainer="raisjn <of.raisjn@gmail.com>"
 license=MIT
 url="https://github.com/ddvk/remarkable2-framebuffer"
-pkgver=1:0.0.30-1
+pkgver=1:0.0.31-1
 _release="${pkgver%-*}"
 _release="v${_release#*:}"
 _libver=1.0.1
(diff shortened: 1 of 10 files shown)


LinusCDE (dismissed):
Tested to be working on the rM 2 :+1:

Eeems:
We are doing some further testing on uninstalls and managed to break something. I'm working on a break fix now that we'll include before merging this.

LinusCDE (approved):

PR #695, opened 2023-05-27
Update rmFM host

TheLastZombie:
Essentially the same PR as #682, except against the `testing` branch instead of `stable`.

From #682:

> I'm migrating to a self-hosted [Forgejo](https://forgejo.org/) instance, so the URLs are changing.


Eeems (approved):
Untested. Since it builds, and the sha256sum is the same for the source, I'm making what should be a reasonable assumption that nothing changed and the testing we did previously is good enough.

PR #697, opened 2023-06-02
Oxide: depend on rm2fb on rm2 only

Eeems:


raisjn (approved):
looks good

Jayy001:
Just tested on 3.2.3, can confirm that when disabling rm2fb, tarnish will fail and xochitl will automatically be started. Happy to merge.

Just a quick note, when using with remux - it seems that it automatically started after the tarnish fail (launcher for xochitl). Might be worth looking in the future.

Eeems:
> Just a quick note, when using with remux - it seems that it automatically started after the tarnish fail (launcher for xochitl). Might be worth looking in the future.

You shouldn't be using oxide at the same time as remux, their service files are marked as conflicting with each other.

PR #698, opened 2023-06-05
Update oxide to v2.6

Eeems:
* Removed `notify-send` package and replaced it with `oxide-utils`
* Updated `erode`, `fret`, `oxide`, `rot`, `tarnish`, `decay`, `corrupt`, `anxiety`, `liboxide` to v2.6
* Added `inject_evdev` package

[Release notes](https://github.com/Eeems-Org/oxide/releases/tag/v2.6)

PR #700, opened 2023-06-07
Call update-desktop-database to refresh Oxide's application cache on install

Eeems:
This makes sure that applications will automatically show up in oxide when they only provide a draft file.

This is a good candidate for install hooks that packages can install to avoid having to repackage everything with new scripts.


Eeems:
This is likely failing because I'm targeting the `oxide` branch, which is for #698, so it's trying to build every single package.

Eeems:
When #698 is merged, this can be marked as ready.

Eeems:
https://github.com/toltec-dev/toltec/blob/9c1aa92aaa747b6afbf45e5c9124a6fbaf9ea427/scripts/toltec/builder.py#L549-L553
TODO:

- [x] Update this hook to also handle draft apps
- [x] Remove update-desktop-database calls from all package recipes
- [x] Update documentation to talk about this hook

Eeems:
I've chosen not to resolve the issue with whiteboard-hypercard in this PR and to leave it in a less complete state to be resolved later.
https://github.com/toltec-dev/toltec/issues/729

Eeems:
The folly build is broken now as well, I'm not going to resolve it in this PR
https://github.com/toltec-dev/toltec/issues/748

Jayy001 (dismissed):
Tested on RM2 3.3.2.1666 with no issues - automatically adds application

Eeems:
It looks like the place we download KOReader from is down at the moment, which is blocking this.

Eeems:
Opened https://github.com/koreader/koreader/issues/11146 to alert the koreader team about the site being down as it's been all day and it's still down, and I didn't see another issue opened for it.

Jayy001 (approved):

PR #702, opened 2023-06-09
Update zoneinfo-utils to 2023c

Eeems:
- Update zoneinfo-utils to 2023c-1
- Update workflow runners to use ubuntu-22.04


Eeems:
```
[   ERROR] toltec.builder: zoneinfo-utils [rmall] (zoneinfo-utils): ./zic.x86  -d '/home/runner/work/toltec/toltec/build/package/zoneinfo-utils/rmall/pkg/zoneinfo-utils/usr/share/zoneinfo'  tzdata.zi
[   ERROR] toltec.builder: zoneinfo-utils [rmall] (zoneinfo-utils): ./zic.x86: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./zic.x86)
```
When I try to run this locally, it works fine. I'm also building `zic.x86` as part of the packaging, so I don't see how it can't find libc since it's built using it. This also used to work when this package was initially created.

@matteodelabre any insight?

Eeems:
`package()` is run on the host, while `build()` is run in the toolchain container. This means that the host doesn't have the right version of glibc, or it's in a different location.

Eeems:
Tested and works on my new rM2 :sunglasses:

raisjn (approved):

PR #703, opened 2023-06-28
Add ReTerm

Eeems:


bkirwi (dismissed):
Installed correctly. I can't test the keyboard since I don't own one, but it appears to start up just fine.

I notice

```
term=killall -9 dotnet
```

in the installed draft file, which is pretty sketchy... any other dotnet executable on the tablet would also be killed. cc @i-am-shodan - probably wise to narrow this down!

Eeems:
> Installed correctly. I can't test the keyboard since I don't own one, but it appears to start up just fine.
> 
> I notice
> 
> ```
> term=killall -9 dotnet
> ```
> 
> in the installed draft file, which is pretty sketchy... any other dotnet executable on the tablet would also be killed. cc @i-am-shodan - probably wise to narrow this down!

Oh right, that needs to change. This is compiled directly into a standalone executable for the package, so it's not running as dotnet. I'll get that changed.

bkirwi (approved):

PR #704, opened 2023-06-29
Better bootstrap check-install

Eeems:
I just had to help someone through reinstalling toltec after a library got corrupted when installing something with nao. One thing I identified is that our current solution for bootstrap checking to see if things are in a partial install state wasn't properly checking to see if /opt is bind mounted still or not. This changes that and provides more context on what it detects when it detects a partial installation.


scripts/bootstrap/bootstrap
```diff
@@ -61,17 +61,37 @@ exists-non-empty() {
     [[ -d $1 ]] && files="$(ls -A -- "$1")" && [[ -n $files ]]
 }
 
+already-installed-message() {
+    log "Toltec is already installed or partially installed"
+    log "Toltec is already installed or partially installed"
```
bkirwi: Duplicate?

bkirwi (commented):
Otherwise seems reasonable and useful, though I'm not very familiar with bootstrap yet...

scripts/bootstrap/bootstrap
Eeems: Hmm, something must have gone wrong when I copy pasted it.

bkirwi (dismissed):

Eeems:
I just ran this to test it again. It's working as expected. I also added a small guard that verifies that it's being run with `bash`, and not with `sh`, `ash` or another shell. This only really helps when testing changes, and you are running bootstrap manually and forget that it isn't compatible with `sh`.

Jayy001 (approved):

Eeems:
Example outputs:
image
image

PR #706, opened 2023-07-12
tailscale-systemd package

kiwiz:
*What*
A package with the necessary systemd service configuration to get Tailscale working on rM devices. Requires the base `tailscale` package.

*Authorship*
Service file comes directly from https://remarkable.guide/tailscale.html

*Specificity*
Should work on all

### DIFF
--- a/package/tailscale-systemd/package
+++ b/package/tailscale-systemd/package
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # Copyright (c) 2020 The Toltec Contributors
-# SPDX-License-Identifier: BSD-3-Clause
+# SPDX-License-Identifier: MIT
 
 pkgnames=(tailscale-systemd)
 pkgdesc="SystemD configuration for tailscale"
--- a/package/tailscale-systemd/package
+++ b/package/tailscale-systemd/package
@@ -22,3 +22,25 @@ sha256sums=(
 package() {
     install -D -m 644 -t "$pkgdir"/etc/systemd/system "$srcdir"/tailscaled.service
 }
+
+configure() {
+    systemctl daemon-reload
+    systemctl enable tailscaled
+    systemctl start tailscaled
+}
+
+preremove() {
+    if is-active tailscaled; then
+        echo "Stopping tailscaled"
+        systemctl stop tailscaled
+    fi
+
+    if is-enabled tailscaled; then
+        echo "Disabling tailscaled"
+        systemctl disable tailscaled
+    fi
+}
+
+postremove() {
+    systemctl daemon-reload
+}
--- a/package/tailscale-systemd/package
+++ b/package/tailscale-systemd/package
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(tailscale-systemd)
+pkgdesc="SystemD configuration for tailscale"
+url=https://tailscale.com
+pkgver=0.0.1
+section="devel"
+timestamp=2023-07-12T00:00Z
+maintainer="Kai <z@kwi.li>"
+license=MIT
+installdepends=(tailscale)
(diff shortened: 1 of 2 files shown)


package/tailscale-systemd/package
```diff
@@ -0,0 +1,17 @@
...
+url=https://tailscale.com
+pkgver=0.0.1
+section="devel"
+timestamp=2023-07-12T00:00Z
+maintainer="Kai <z@kwi.li>"
+license=MIT
+installdepends=(tailscale)
+
+package() {
+    install -D -m 644 -t "$srcdir"/tailscaled.service /lib/systemd/system/tailscaled.service
```
Eeems: We've been putting service files under `/etc/systemd/system` as a rule in toltec.

package/tailscale-systemd/tailscaled.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Tailscale node agent
+Documentation=https://tailscale.com/kb/
+Wants=network-pre.target
+After=network-pre.target
+
+[Service]
+Environment="HOME=/home/root"
+ExecStartPre=/opt/bin/tailscaled --cleanup
+ExecStart=/opt/bin/tailscaled --state=/var/lib/tailscale/tailscaled.state --socket=/run/tailscale/tailscaled.sock --tun=userspace-networking --socks5-server=localhost:1055 --outbound-http-proxy-listen=localhost:1055
```
Eeems: Try to place things under `/opt` if possible. So `/var/lib/tailscale/tailscaled.state` should be `/opt/var/lib/tailscale/tailscaled.state`. it's fine to use `/run` as it's tmpfs.

Eeems (changes requested):

kiwiz:
* [ ] Test on device

kiwiz:
Having trouble running `toltecmk` on my host. Looks like it might be due to https://github.com/toltec-dev/build/pull/31, but I don't see anything that would be interpretted as a C-quoted string.

Eeems:
> Having trouble running `toltecmk` on my host. Looks like it might be due to [toltec-dev/build#31](https://github.com/toltec-dev/build/pull/31), but I don't see anything that would be interpretted as a C-quoted string.

It's based on the version of bash you are using underneath or something IIRC, so you won't see it in your package recipe.

package/tailscale-systemd/package
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # Copyright (c) 2020 The Toltec Contributors
-# SPDX-License-Identifier: MIT
+# SPDX-License-Identifier: BSD-3-Clause
```
Eeems: This is the license for this file, not the underlying software it's packaging. It must be MIT.

kiwiz:
Tested on my rm2. Install + usage works fine. However, `is-active` and `is-enabled` are undefined when uninstalling. Do those need to be sourced somehow?

Eeems:
> Tested on my rm2. Install + usage works fine. However, `is-active` and `is-enabled` are undefined when uninstalling. Do those need to be sourced somehow?

When built through toltec, install-lib is available, but toltecmk doesn't include it.

I work around that when building with toltecmk by embedding them when a package isn't part of the main toltec workflow: https://github.com/Eeems-Org/oxide/blob/master/package#L17-L53

kiwiz:
Ack. Tested locally with the relevant functions and it looks good.

Eeems:
> Ack. Tested locally with the relevant functions and it looks good.

Have you tested with the packages generated by the actions on this PR?

kiwiz:
Oh, I have not! Will do that now.

kiwiz:
Confirmed that:
1) Package installs
2) Tailscaled comes up
3) Uninstall cleans up service

Eeems (approved):
I haven't tested anything as I'm not setup to be able to test tailscale at all, but if it works for you that's good enough for me right now.

PR #707, opened 2023-07-19
Make sure xochtil shows up properly when run

Eeems:
This make sure that xochitl properly shows up as `/usr/bin/xochtil` when you run it using the wrapper script.


TheSunCat:
I installed the `xochitl_0.0.0-15_rmall.ipk` package from the artifacts, and ran `xochitl` from the ssh prompt (which resolves to `/opt/bin/xochitl` according to `which xochitl`). I then checked with `ps | grep xo` which shows the correct path in my new process (the second one):
```
root@reMarkable:~# ps | grep xo
 6858 root      270m S    /usr/bin/xochitl
13930 root      251m S    /usr/bin/xochitl
```

Jayy001 (approved):

PR #709, opened 2023-07-24
Add valgrind

Eeems:
Entware already contains valgrind, but it's not really useful for things built targeting the reMarkable instead of generic entware. It also appears to be missing some tools.

This does mean that we have a name conflict with the packages. I'm tempted to rename it to `toltec-valgrind` or `remarkable-valgrind` and mark it as conflicting with valgrind. Thoughts?

### DIFF
--- a/package/valgrind/package
+++ b/package/valgrind/package
@@ -12,15 +12,14 @@ image=base:v2.3
 makedepends=(build:gawk build:automake)
 
 build() {
-    ./autogen.sh
-    ./configure --prefix=/opt --host=arm-linux-gnueabihf
-    sed -i -e "s#armv7#arm#g" configure
+    export CC=arm-linux-gnueabihf-gcc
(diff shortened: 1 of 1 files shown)


Eeems:
Well that's odd, when building locally with toltecmk it doesn't fail. Not sure why the exact same image fails here.

Eeems:
Okay, next issure. ld-linux-armhf.so.3 is stripped so we don't have debug symbols to work with.

Eeems:
It looks like the toolchain contains unstripped versions of the libraries in the arm sysroot. I'm not really sure if it's worth trying to provide them in a package somehow? It would be possible to automate installing the toolchain, and then grabbing the files from the sysroot and stuffing them into a package that bind mounts them into a `.debug` folder in `/lib` and `/usr/lib` or something, whatever valgrind would need to work.

Is it worth it?

PR #713, opened 2023-07-31
Add zerotier-one

Eeems:


Etn40ff:
I think that zerotier is already available in entrare but it misses the systemd integration.

Eeems:
> I think that zerotier is already available in entrare but it misses the systemd integration.

It looks like you are correct. I'll have to check to see what all they include, but it is a couple of minor versions behind. They probably don't have the manpages or selftest packages either.

Eeems:
https://github.com/Entware/entware-packages/tree/master/net/zerotier

It looks like they have the same config and a couple of other small patches that I should take a closer look at. They also have picked an odd name for the package. Most distributions call it zerotier-one since that's what the software is called.

I'll have to adjust my package to take into account that the entware one exists, and add some of the patches they use.

Eeems:
Tested on my rM1 and it's working. The selftest succeeds as well:
image

Eeems:
The selftest also passes on my rM2, as well as running it. A lack of tun in the kernel I'm on is keeping it from fully working, but I'll resolve that when I can test a custom kernel.

Eeems:
Tested with `linux-stracciatella` and this works as expected.

raisjn (approved):

PR #715, opened 2023-07-31
Add terminfo dependency to yaft and add /opt/etc/profile to /etc/prof…

### LINKED ISSUE #714: yaft package doesn't use terminfo file(s) in `/opt/share/terminfo`
This leads to TUI programs like `vim` not working properly.

Currently, the files are dropped into `/opt/share/terminfo/y/yaft-256color`. The file should either be copied or symlinked into `/etc/terminfo/y`.

Relevant `yaft` issue: https://github.com/timower/rM2-stuff/issues/15

Eeems:
This does the following:

- Adds `/opt/etc/profile` to `/etc/profile.d` so that it automatically loads the environment variables provided by the entware source
- Adds `terminfo` as a dependency of yaft

We may want to add some extra default environment variables to `/etc/systemd/system.conf.d` in the future as well.


kiwiz:
Tested via the following steps:

* Ran
    ```bash
    ln -sf /opt/etc/profile /etc/profile.d/toltec.sh
    sed -i \
        -e 's|^export TERM=xterm|if \[ -z "$TERM" \];then export TERM=xterm;fi|' \
        /opt/etc/profile
    ```
* Restarted `yaft` and invoked `bash -l`
* Verified that `TERMINFO` was set and `TERM=yaft-256color`
* Ran `vim` and verified that it rendered correctly

timower:
>   Fetching Compiled Descriptions
>       The  ncurses  library  searches	for  terminal  descriptions in several
>       places.	It uses	only the first description found.  The library	has  a
>       compiled-in  list  of places to search which can	be overridden by envi-
>       ronment variables.  Before starting to search, ncurses  eliminates  du-
>       plicates	in its search list.
>
>       o   If  the  environment	variable TERMINFO is set, it is	interpreted as
>	   the pathname	of a directory containing the compiled description you
>	   are working on.  Only that directory	is searched.
>
>       o   If  TERMINFO	is not set, ncurses will instead look in the directory
>	   $HOME/.terminfo for a compiled description.
>
>       o   Next, if the	environment variable  TERMINFO_DIRS  is	 set,  ncurses
>	   will	 interpret  the	 contents of that variable as a	list of	colon-
>	   separated directories (or database files) to	be searched.
>
>	   An empty directory name (i.e., if the variable begins or ends  with
>	   a  colon, or	contains adjacent colons) is interpreted as the	system
>	   location /usr/share/misc/terminfo.
>
>       o   Finally, ncurses searches these compiled-in locations:
>
>	   o   a list of directories (@TERMINFO_DIRS@),	and
>
>	   o   the system terminfo  directory,	/usr/share/misc/terminfo  (the
>	       compiled-in default).

Should I / we just install the terminfo files to `$HOME/.terminfo`

Eeems:
> 
> Should I / we just install the terminfo files to `$HOME/.terminfo`

Not for toltec/entware packages. They are meant to go under `/opt`, and we were just missing expected environment variables that the entware side of the install expects.

timower (approved):
I'll also add the `-l` upstream

bkirwi (approved):

Eeems:
@bkirwi did you happen to test this? I think the only testing was by myself on my rM1

bkirwi:
Nope! I'm personally comfortable with your and @kiwiz' testing, since this is not one of the areas where I know the RM2 to be different, but I'd completely understand if you want to hold out for an additional test. (Though I personally won't have the chance to do it until probably September.)

Eeems:
@kiwiz would you be willing to try installing the ipk to make sure it works as expected?

kiwiz:
Installed both `ipk` files and confirmed:

* `TERM=yaft-256color`
* `vim` runs correctly

PR #718, opened 2023-08-03
Add support for rm-sync in xochitl package

### LINKED ISSUE #717: 3.5.1.1798 onwards renames sync to rm-sync
The following will need to be updated to handle both names:
https://github.com/toltec-dev/toltec/blob/9c1aa92aaa747b6afbf45e5c9124a6fbaf9ea427/package/xochitl/package#L68-L75
https://github.com/toltec-dev/toltec/blob/9c1aa92aaa747b6afbf45e5c9124a6fbaf9ea427/package/xochitl/manual-sync.service#L16

timower:
Recent xochitl versions renamed the sync service and binary to rm-sync. This updates the xochitl package to support both.


package/xochitl/package
```diff
@@ -70,8 +70,15 @@ configure() {
     if [[ "x$(systemctl is-enabled sync.service)" != "xmasked" ]]; then
         systemctl mask sync.service
     fi
-    if ! is-active manual-sync.service; then
-        systemctl enable --now manual-sync.service
+
+    if [[ "x$(systemctl is-enabled rm-sync.service)" != "xmasked" ]]; then
```
Eeems: image
This will need to first check to see if the unit exists before checking if it's masked.

Eeems (changes requested):
image
image

A preupgrade script will need to stop and disable the old manual-sync.service

package/xochitl/package
timower: We don't accept 'proceeding anyway'?

Eeems: > We don't accept 'proceeding anyway'?

No, as it adds noise that the end user doesn't need to care about, and there will be users who open tickets because of it.

Eeems:
@timower poke

package/xochitl/package
timower: `> /dev/null` which we already had in the postremove should fix this.

Eeems: > `> /dev/null` which we already had in the postremove should fix this.

This will make it not display the error, but it will also hide real errors from the user. I'm not comfortable doing that. I'd much rather add a `unit-exists` method and check that before seeing if it's masked.

```bash
unit-exists() {
  systemctl --quiet list-unit-files "${1}" > /dev/null
}
```

Eeems: Upon testing on device, it looks like the version of systemctl that ships with it in 2.15 doesn't give failed error codes. This should work though:
```bash
unit-exists() {
  [ $(systemctl --quiet list-unit-files "${1}" | grep "${1}" | wc -l) -eq 1 ]
}

**Comment by Eeems** [comment]
@timower poke

File: package/xochitl/package (hunk 1 above)
Comment (Eeems) [reply package/xochitl/package]: #721 contains the unit-exists method, you can pull the implementation from there, or wait for it to be merged first.

Comment (timower) [reply package/xochitl/package]: I now always checks for the binary name, so it shouldn't log anymore

**Comment by timower** [comment]
Verified on 2.15 (upgrade, fresh install & remove) and 3.5 (fresh install & remove). Seems to work

File: package/xochitl/package (hunk 1 above)
Comment (Eeems) [reply package/xochitl/package]: That's another way to accomplish it I guess

File: package/xochitl/package [hunk 2]
```diff
@@ -102,3 +116,9 @@ postremove() {
         systemctl enable xochitl.service 2> /dev/null
     fi
 }
+
+preupgrade() {
+    if is-enabled manual-sync.service; then
+        systemctl disable manual-sync.service
+    fi
```
Eeems:
```suggestion
    if is-enabled manual-sync.service; then
        systemctl disable manual-sync.service
    fi
    if is-active manual-sync.service; then
        systemctl stop manual-sync.service
    fi
```

Eeems (changes requested):
Upgrading an existing install doesn't error, but I'm left with `manual-sync.service` still running and `manual-sync@sync` in a failed state with the following error:
```
Failed to start reMarkable Document Sync without the xochitl.service dependency.
```
Stopping `manual-sync.service` and then starting `manual-sync@sync` allows them to start as expected.

timower:
I cannot build any packages locally anymore, as I'm always running into:
```
# make xochitl
./scripts/package_build.py  "xochitl"
Traceback (most recent call last):
  File "/src/./scripts/package_build.py", line 44, in <module>
    repo = Repo(paths.RECIPE_DIR, paths.REPO_DIR)
  File "/src/scripts/toltec/repo.py", line 60, in __init__
    self.generic_recipes[name] = GenericRecipe.from_file(
  File "/src/scripts/toltec/recipe.py", line 46, in from_file
    return GenericRecipe(name, path, recipe.read())
  File "/src/scripts/toltec/recipe.py", line 59, in __init__
    variables, functions = bash.get_declarations(definition)
  File "/src/scripts/toltec/bash.py", line 147, in get_declarations
    assert next_token == "("
AssertionError
```
I tried different python versions and some docker image, without success.

Eeems:
Python 3.10 is currently required for the toltec build system. Toltecmk supports 3.11, but hasn't been fully updated to do everything the toltec build system does.

Eeems:
@timower unfortunately preupgrade won't run when upgrading from a previous package to this package, only configure from this package runs: https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#install-section

@matteodelabre I'm not sure how the current pre/post upgrade scripts work makes sense as this keeps coming up. It's basically impossible to know what pre/post upgrade steps should be run until you are writing the new package. I think it might be worth us changing this longer term so that pre/post upgrade are run from package B along with configure. That or we start building some standards on what should actually be in those scripts by default in a package.

Eeems (approved):
I've only tested this on versions before the new rm-sync service, as that's all toltec supports at the moment. When we move to add support for 3.5+ we'll be retesting this along with other things.

PR #719, opened 2023-08-14
Add sill package

kiwiz:


Eeems (approved):
It would be nice if we had an icon for it, but it works as-is.

PR #716, opened 2023-08-02
Update yaft to 0.0.8

Eeems:
Add landscape support

PR #711, opened 2023-07-26
Add issue templates

Eeems:

PR #684, opened 2023-05-07
Update to python 3.10

Eeems:
Fixes #683

PR #657, opened 2022-12-12
package/linux-mainline: Bump to 6.2

alistair23:
Bump the mainline based kernel to the ~~6.1~~ 6.2 release.

Besides the mainline changes from 6.0 to 6.2, this includes a refactoring of the `rohm,bd71815` code to use the mainline driver instead of the reMarkable version. It has also dropped a Koreader related hack from the kernel, see https://github.com/koreader/koreader/issues/10012.

This is the ~~6.1~~ 6.2 mainline kernel and ~~18~~ 12 rM2 specific patches on top.

PR #722, opened 2023-08-18
Stable merge for Week 33 of 2023

Eeems:
### New Packages
- `reterm` - 0230628.1927-1 (#703)
- `sill` - 0.1.1-alpha (#719)
- `changescrn` - 1.0.0-1 (#571)

### Updated Packages
- `yaft` - 0.0.8-2 (#715 #716)
  - Update to 0.0.8
  - Depend on `terminfo`
  - Add patch to start bash as a login shell
- `toltec-base` - 1.2-2 (#715)
  - Adds `/opt/etc/profile` to `/etc/profile.d` so that it automatically loads the environment variables provided by the entware source
- `linux-mainline` - 6.2.0-1 (#657)

### Tooling
- Update build process to use Python 3.10 (#684)
- Added issue templates to the repository (#711)


FouzR:
I’ve tested Sill and toltec-base on 3.2. Haven’t tested changescrn and ReTerm yet. And due to a newer version of YAFT that was self-built already being present, I haven’t tested that either

Jayy001:
# Reterm
* 57MB Averge mem usuage
* Only works when foloio activated before hand
* Can draw stars with your fingers which is cool
* Interesting background
reterm

# Sill
* Around 2MB averge mem usuage
* Beauitfull UI - however, hard to navigate a bit
* When writing, text is drawn and disappears (not sure what the error is here but viewing files works fine)

# Yaft
* Wow, amazing as usual - super smooth and works both landscape and potrait 
* Around 11MB usuage
yaft

# Changescrn
* Requires xochitl restart
* Simple to use, good UX
changescr

# Toltec-base /opt/etc/profile
* LGTM! No problems.

# New kernel
reMarkable: ~/toinstall/rm2/ opkg install linux-mainline_6.2.0-1_rm2.ipk 
Installing linux-mainline (6.2.0-1) to root...
Configuring linux-mainline.
The new kernel files have been copied, but not installed.
Please use kernelctl to select the kernel to boot.

Known issues with the mainline kernel:
 - No support for low power mode (suspend uses more power then it should)
 - Device doesn't always resume from suspend, a reboot is required
 - GUI shutdown in Oxide doesn't work
 - Wacom stylus doesn't work in Xochitl (https://github.com/reMarkable/linux/issues/15)
 - Finger input doesn't work in Koreader (https://github.com/koreader/koreader/issues/1
 - No OTG control support

**I have not tested using this kernel.**

Eeems:
@alistair23 do you know why GUI shutdown in Oxide doesn't work? I would expect it to work since it's just calling `systemctl poweroff`:
https://github.com/Eeems-Org/oxide/blob/b5b37efe5e7b6e4c461e545ebcec7a7f29a92622/applications/system-service/systemapi.h#L378-L388

In your testing, did you maybe have something inhibiting poweroff?

There also doesn't seem to be an issue tracking this?

Eeems:
Alright, I think this is all the testing that is required. I'm okay with the kernel not being tested further at this point.

alistair23:
@Eeems  `systemctl poweroff` doesn't work either. I actually don't think it's specifically tied to the kernel, I think it's something to do with the command line arguments that I'm using (for extra debugging). I haven't ever really looked into it though

Eeems:
> @Eeems `systemctl poweroff` doesn't work either. I actually don't think it's specifically tied to the kernel, I think it's something to do with the command line arguments that I'm using (for extra debugging). I haven't ever really looked into it though

Thanks for the context @alistair23 I think it would be worth updating the message that the package provides with a bit more information about that, or maybe a link to a bug where we can keep up-to-date information? It won't be a blocker for this stable merge, but I would appreciate it for the next one.

alistair23:
I was just thinking the same thing. I'll prep something

raisjn (approved):

PR #723, opened 2023-08-22
package/linux-mainline: Use a custom defconfig

alistair23:
Use a reMarkable specific defconfig that is checked into toltec. Currently this is an exact copy of the mainline defconfig (so no functional changes) but in the future this can be patched to enable/disable features as required.


Eeems:
@alistair23 you need to bump the package version for it to be built, I've done so, and I've confirmed that the defconfig matches the one from the source repo at the same commit, so I'm approving this for merge as soon as it finishes building.

Eeems (approved):

PR #725, opened 2023-08-23
Added reGenda

tenJirka:
reGenda is agenda app for reMarkable tablet 

It is written in Python and uses wrapper around SAS. Can connect to caldav server and show events from there. (read only for now). For more info check my [repo](https://github.com/tenJirka/reGenda).

I'm author of the package and I tested it on reMarkable 2 but I hope it should work on reMarkable 1 too.


package/regenda/package
```diff
@@ -0,0 +1,32 @@
...
+    install -D -m 644 -t "$pkgdir"/opt/etc/reGenda/ "$srcdir"/config.yml.example
+    install -D -m 644 "$srcdir"/icon.png "$pkgdir"/opt/etc/draft/icons/reGenda.png
+    install -D -m 755 -t "$pkgdir"/opt/reGenda/ "$srcdir"/run.py
+    install -D -m 644 -t "$pkgdir"/opt/reGenda/ "$srcdir"/languages.py
+    install -D -m 644 -t "$pkgdir"/opt/reGenda/ "$srcdir"/calendar_caldav.py
+    ln -s /opt/reGenda/run.py "$pkgdir"/opt/bin/reGenda
+}
+
+configure() {
+    pip install --upgrade pip caldav pyyaml pytz rm_pysas
```
Eeems: I'm not comfortable at all with this, these will need to be added as packages that are depended on instead.

package/regenda/package
```diff
@@ -0,0 +1,32 @@
...
+    echo "You have to config server in /opt/etc/reGenda/config.yml, if you have any trouble, see https://github.com/tenJirka/reGenda for help"
```
Eeems: This should be flushed out a bit more, and a good enough starting point configuration added to the package.

package/regenda/package
```diff
@@ -0,0 +1,32 @@
...
+flags=(patch_rm2fb)
+
+source=(https://github.com/tenJirka/reGenda/archive/refs/tags/0.0.1.zip)
+sha256sums=(00d6a1e2f6b59f86ba47ea5c3444f24e112d24eb62082802653fe4649153da01)
+
+package() {
+    install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$srcdir"/regenda.draft
+    install -D -m 644 -t "$pkgdir"/opt/etc/reGenda/ "$srcdir"/config.yml.example
+    install -D -m 644 "$srcdir"/icon.png "$pkgdir"/opt/etc/draft/icons/reGenda.png
+    install -D -m 755 -t "$pkgdir"/opt/reGenda/ "$srcdir"/run.py
```
Eeems: While we do have a couple packages that put their files directly in `/opt/{pkgname}`, it would be better to place them further down the file structure. If possible, I would recommend moving files to places under `/opt` that mirror the [file-hierarchy](https://man.archlinux.org/man/file-hierarchy.7.en)

Eeems (changes requested):

Eeems:
I did a quick pass to restructure the package with the following:

- Removed display from depends and instead used the patch_rmfb flag
- Moved all filesystem layout related stuff from configure() to package()
- Removed all the extra mkdir calls as they are not needed when using install
- Simplified the install statements that don't rename the file

package/regenda/package
Eeems: Especially since this upgrades pip, which is a system package.

`python3-pytz` is already a package that can be used. caldav, pyyaml, and rmpysas are not, and will need to be added.

Eeems: One option for you would be to vendor these files and store them alongside your scripts for now, instead of creating separate packages for all of them. You can do this as part of the build by `pip` installing them in the build process, and copying the python files/folders. That said, if they have any compiled portions, that will not work.

raisjn:
just want to comment and say: cool app!!

Eeems:
> just want to comment and say: cool app!!

Oh yeah, ditto

package/regenda/package
tenJirka: I will look at that or I have an idea to install them via pip as written but not system wide but only to the directory to reGenda using `pip -t` ? Updating pip is not necessary as well. 
I just want to know if it can be an option.

tenJirka: Should be ok now. I added all modules that cant be installed from toltec to package alongside script. I hope I managed to add all of them.

package/regenda/package
tenJirka: I'm not sure how to do this. The configuration files are already in /opt/etc, I can't move the executable to /opt/bin because it has to be in the same directory as all the libraries. And now all pip modules must be there too.
If you give more clues as to how you'd like it, I'll look into it.

package/regenda/package
tenJirka: Should be better now.

package/regenda/package
Eeems: `/opt/usr/lib/reGenda` would probably be a good location.

> /usr/lib/
Static, private vendor data that is compatible with all architectures (though not necessarily architecture-independent). Note that this includes internal executables or other binaries that are not regularly invoked from a shell. Such binaries may be for any architecture supported by the system. Do not place public libraries in this directory, use $libdir (see below), instead.

package/regenda/package
```diff
@@ -0,0 +1,64 @@
...
+    cp -r "$srcdir"/python_modules/rm_pySAS* "$pkgdir"/opt/reGenda/
+    cp -r "$srcdir"/python_modules/recurring_ical_events* "$pkgdir"/opt/reGenda/
+    cp -r "$srcdir"/python_modules/tzlocal* "$pkgdir"/opt/reGenda/
+    cp -r "$srcdir"/python_modules/vobject* "$pkgdir"/opt/reGenda/
+    cp -r "$srcdir"/python_modules/x_wr_timezone* "$pkgdir"/opt/reGenda/
+}
+
+configure() {
+    if [ ! -f /opt/etc/reGenda/config.yml ]; then
+        while true; do
```
Eeems: We don't have a precedent for user interaction on installation, I'd prefer not to add it as it would break using nao to install this package. I think it would probably be best to just copy the file if it doesn't exist and output information letting the user know that it will need to be modified.

package/regenda/package
```diff
@@ -0,0 +1,64 @@
...
+
+package() {
+    install -D -m 644 -t "$pkgdir"/opt/etc/draft/ "$srcdir"/regenda.draft
+    install -D -m 644 -t "$pkgdir"/opt/etc/reGenda/ "$srcdir"/config.yml.example
+    install -D -m 644 "$srcdir"/icon.png "$pkgdir"/opt/etc/draft/icons/reGenda.png
+    install -D -m 755 -t "$pkgdir"/opt/reGenda/ "$srcdir"/run.py
+    install -D -m 644 -t "$pkgdir"/opt/reGenda/ "$srcdir"/languages.py
+    install -D -m 644 -t "$pkgdir"/opt/reGenda/ "$srcdir"/calendar_caldav.py
+    mkdir "$pkgdir"/opt/bin/
+    ln -s /opt/reGenda/run.py "$pkgdir"/opt/bin/reGenda
```
Eeems: Have you tested this package yet? I just realized that I'm not sure if this will work properly as a soft link, since the libraries you included are not in the same folder as the link, which is the main entrypoint.

I believe that python packages tend to create a separate python executable that will just run the main one's main function. For example:

```python
#!/usr/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from remarkable_mouse.remarkable_mouse import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

Eeems (changes requested):

package/regenda/package
tenJirka: Yeah I did, works fine. This should not be necessary.
I was considering it as an option but when I find out it works well as it is, I decided to keep the simple solution.

package/regenda/package
tenJirka: Oh, I totally agree. 
Already moved here.

package/regenda/package
tenJirka: I understand, already implemented.

tenJirka:
I've added postremove function because a lot of empty directories were left over after uninstall in /opt/usr/lib/reGenda.

Eeems:
> I've added postremove function because a lot of empty directories were left over after uninstall in /opt/usr/lib/reGenda.

This shouldn't be the case unless runtime is creating files? That said, blanked removing the directory could remove user created files, which is not something we want to do. Empty folders being left behind is not as large of a concern.

tenJirka:
> > I've added postremove function because a lot of empty directories were left over after uninstall in /opt/usr/lib/reGenda.
> 
> This shouldn't be the case unless runtime is creating files? That said, blanked removing the directory could remove user created files, which is not something we want to do. Empty folders being left behind is not as large of a concern.

Some pycache files should be only created files by runtime. I don't know how bad it would be if they they will last.

```
root@reMarkable:/opt/usr/lib/reGenda# find . -name "*.pyc"
./__pycache__/languages.cpython-311.pyc
./__pycache__/calendar_caldav.cpython-311.pyc
root@reMarkable:/opt/usr/lib/reGenda# ls -lh __pycache__/
-rw-r--r--    1 root     root        3.7K Aug 24 19:18 calendar_caldav.cpython-311.pyc
-rw-r--r--    1 root     root        2.4K Aug 24 19:18 languages.cpython-311.pyc
```

Eeems:
> > > I've added postremove function because a lot of empty directories were left over after uninstall in /opt/usr/lib/reGenda.
> > 
> > 
> > This shouldn't be the case unless runtime is creating files? That said, blanked removing the directory could remove user created files, which is not something we want to do. Empty folders being left behind is not as large of a concern.
> 
> Some pycache files should be only created files by runtime. I don't know how bad it would be if they they will last.
> 
> ```
> root@reMarkable:/opt/usr/lib/reGenda# find . -name "*.pyc"
> ./__pycache__/languages.cpython-311.pyc
> ./__pycache__/calendar_caldav.cpython-311.pyc
> root@reMarkable:/opt/usr/lib/reGenda# ls -lh __pycache__/
> -rw-r--r--    1 root     root        3.7K Aug 24 19:18 calendar_caldav.cpython-311.pyc
> -rw-r--r--    1 root     root        2.4K Aug 24 19:18 languages.cpython-311.pyc
> ```

It might be best to target those folders and files explicitly then instead of a blanket delete. Ideally, we could generate them as part of the package process and include them, but we don't really have the build process setup for that yet, as they can only be created on the same architecture.

```shell
rm -f /opt/usr/lib/reGenda/**/*.cpython-*.pyc
```

Eeems:
I took a look at what entware distributes for their python packages, they just distribute the pyc files instead of the source. So they don't have to do any magic cleanup. We'll have to sort out a way to generate them for python packages in the future. I'm fine with just removing the generated pyc files as part of cleanup for now in this package.

tenJirka:
> ```shell
> *.cpython-*.pyc
> ```



> > > > I've added postremove function because a lot of empty directories were left over after uninstall in /opt/usr/lib/reGenda.
> > > 
> > > 
> > > This shouldn't be the case unless runtime is creating files? That said, blanked removing the directory could remove user created files, which is not something we want to do. Empty folders being left behind is not as large of a concern.
> > 
> > 
> > Some pycache files should be only created files by runtime. I don't know how bad it would be if they they will last.
> > ```
> > root@reMarkable:/opt/usr/lib/reGenda# find . -name "*.pyc"
> > ./__pycache__/languages.cpython-311.pyc
> > ./__pycache__/calendar_caldav.cpython-311.pyc
> > root@reMarkable:/opt/usr/lib/reGenda# ls -lh __pycache__/
> > -rw-r--r--    1 root     root        3.7K Aug 24 19:18 calendar_caldav.cpython-311.pyc
> > -rw-r--r--    1 root     root        2.4K Aug 24 19:18 languages.cpython-311.pyc
> > ```
> 
> It might be best to target those folders and files explicitly then instead of a blanket delete. Ideally, we could generate them as part of the package process and include them, but we don't really have the build process setup for that yet, as they can only be created on the same architecture.
> 
> ```shell
> rm -f /opt/usr/lib/reGenda/**/*.cpython-*.pyc
> ```

I was thinking more of `find /opt/usr/lib/reGenda/ -name "*.cpython-*.pyc" -exec rm -r {} +` because modules in subdirectories can produce pycache as well.

Eeems:
Okay, I was doing some more digging. I misremembered, pyc is not architecture dependant, that would be pyd files. So we could compile them easily as part of the build process and only include those. This would require some extra work to make sure to update this package when entware pushes a new version of python.

https://docs.python.org/3/library/py_compile.html
https://docs.python.org/3/library/compileall.html#module-compileall
https://docs.python.org/3/tutorial/modules.html#compiled-python-files

Eeems:
> > It might be best to target those folders and files explicitly then instead of a blanket delete. Ideally, we could generate them as part of the package process and include them, but we don't really have the build process setup for that yet, as they can only be created on the same architecture.
> > ```shell
> > rm -f /opt/usr/lib/reGenda/**/*.cpython-*.pyc
> > ```
> 
> I was thinking more of `find /opt/usr/lib/reGenda/ -name "*.cpython-*.pyc" -exec rm -r {} +` because modules in subdirectories can produce pycache as well.

That is functionally the same as my example. `**` matches all levels of folders.

tenJirka:
> > > It might be best to target those folders and files explicitly then instead of a blanket delete. Ideally, we could generate them as part of the package process and include them, but we don't really have the build process setup for that yet, as they can only be created on the same architecture.
> > > ```shell
> > > rm -f /opt/usr/lib/reGenda/**/*.cpython-*.pyc
> > > ```
> > 
> > 
> > I was thinking more of `find /opt/usr/lib/reGenda/ -name "*.cpython-*.pyc" -exec rm -r {} +` because modules in subdirectories can produce pycache as well.
> 
> That is functionally the same as my example. `**` matches all levels of folders.

My bad, didn't know that.

Eeems:
No worries

tenJirka:
> Okay, I was doing some more digging. I misremembered, pyc is not architecture dependant, that would be pyd files. So we could compile them easily as part of the build process and only include those. This would require some extra work to make sure to update this package when entware pushes a new version of python.
> 
> https://docs.python.org/3/library/py_compile.html https://docs.python.org/3/library/compileall.html#module-compileall https://docs.python.org/3/tutorial/modules.html#compiled-python-files

I guess the decision is up to you because the entware connection is probably nothing that can be done from package recipe.
I can inplement pycache building if you want, but just right clean up looks good enought for me.

Eeems:
> > Okay, I was doing some more digging. I misremembered, pyc is not architecture dependant, that would be pyd files. So we could compile them easily as part of the build process and only include those. This would require some extra work to make sure to update this package when entware pushes a new version of python.
> > https://docs.python.org/3/library/py_compile.html https://docs.python.org/3/library/compileall.html#module-compileall https://docs.python.org/3/tutorial/modules.html#compiled-python-files
> 
> I guess the decision is up to you because the entware connection is probably nothing that can be done from package recipe. I can inplement pycache building if you want, but just right clean up looks good enought for me.

The decision here is actually up to you. I'm fine with merging either at the moment. Do you feel like just getting this done and merged, or trying something slightly new?

Yes, pushing a new version of the package when entware pushes the next major version of python will be something we have to do manually. To some extent, it might still be required if you don't go with the pyc route as well. Up to you.

Longer term I'd like to get a solution for native compilation in toltec of stuff that can't be cross compiled, that way you could make use of pyinstaller or nuitka to make your application truly standalone.

tenJirka:
> > > Okay, I was doing some more digging. I misremembered, pyc is not architecture dependant, that would be pyd files. So we could compile them easily as part of the build process and only include those. This would require some extra work to make sure to update this package when entware pushes a new version of python.
> > > https://docs.python.org/3/library/py_compile.html https://docs.python.org/3/library/compileall.html#module-compileall https://docs.python.org/3/tutorial/modules.html#compiled-python-files
> > 
> > 
> > I guess the decision is up to you because the entware connection is probably nothing that can be done from package recipe. I can inplement pycache building if you want, but just right clean up looks good enought for me.
> 
> The decision here is actually up to you. I'm fine with merging either at the moment. Do you feel like just getting this done and merged, or trying something slightly new?
> 
> Yes, pushing a new version of the package when entware pushes the next major version of python will be something we have to do manually. To some extent, it might still be required if you don't go with the pyc route as well. Up to you.
> 
> Longer term I'd like to get a solution for native compilation in toltec of stuff that can't be cross compiled, that way you could make use of pyinstaller or nuitka to make your application truly standalone.

I will stick with none pyc compilation and right clean up, because I know it will work. 

For longer term if this way of installation will be unwanted I will update it. Right now I'm not sure about pyinstaller because caldav is dependent on lxml which have to be compiled and I don't know how to setup cross compilation for python module right now. 
Lxml actually can't be installed via pip on reMarkble as well because of lack of build dependencies and compilation fail.

Eeems:
> 
> I will stick with none pyc compilation and right clean up, because I know it will work.
> 
> For longer term if this way of installation will be unwanted I will update it. Right now I'm not sure about pyinstaller because caldav is dependent on lxml which have to be compiled and I don't know how to setup cross compilation for python module right now. Lxml actually can't be installed via pip on reMarkble as well because of lack of build dependencies and compilation fail.

pyinstaller and nuitka can't do cross compilation, and the solution I'm thinking of to support them would be to have an arm version of the toolchain images that can be used instead for compilation, so you'd be able to install lxml as a build dependency and then pyinstall/nuitka would be able to compile the executable using them. It'll be a while before that is implemented. Which reminds me, I need to open an issue to track this idea.

Lxml is the bane of my existence.

Eeems:
I've cancelled the build for a bit, we have to wait for testing to finish updating with the latest change going in, otherwise this build will be building the linux kernel. This is due to how we detect what packages need to be built, we compare against what is in testing, and any that don't exist, or have a different version number will be built. Since the testing build with the new package is still in progress and hasn't been deployed, it thinks that this PR has also modified the linux kernel package.

https://github.com/toltec-dev/toltec/actions/runs/5968320607

tenJirka:
> I've cancelled the build for a bit, we have to wait for testing to finish updating with the latest change going in, otherwise this build will be building the linux kernel. This is due to how we detect what packages need to be built, we compare against what is in testing, and any that don't exist, or have a different version number will be built. Since the testing build with the new package is still in progress and hasn't been deployed, it thinks that this PR has also modified the linux kernel package.
> 
> https://github.com/toltec-dev/toltec/actions/runs/5968320607

Okay totally fine, I was i bit worried if reGenda can't be build or something.
This is actually my first Pull requests so I'm curiously watching how it's going. I'm new to all of this.

tenJirka:
I would also like to ask. I'm right that I will have to open new Pull requests when i would like to update this package?

Eeems:
> I would also like to ask. I'm right that I will have to open new Pull requests when i would like to update this package?

Correct.

We may also update it independent of you with pull requests in the future as well. Usually this would be to make it consistent with changes to our packaging standards, force a rebuild when a dependency change in entware requires it, or to fix a bug before it gets fixed in your upstream source.

Eeems (approved):
Tested and it works fine on my reMarkable 1.

PR #727, opened 2023-08-25
Package regenda - added simple as install dependency to recipe

tenJirka:
Hi, 
I'm sorry to realize that late but I forgot to add Simple dependency. I realized that when making replies to some issues. 
Dependencies should be all now.


tenJirka:
I will close this pull requests because I didn't create new branch for it and do it again and properly.
Sorry for spamming, I'm just inexperienced.

PR #728, opened 2023-08-25
Package regenda - added simple as install dependency to recipe 

tenJirka:
Hi,
I'm sorry to realize that late but I forgot to add Simple dependency. I realized that when making replies to some issues.
Dependencies should be all now.

This request is opened instead of this [pull request](https://github.com/toltec-dev/toltec/pull/727), because I didn't create new branch for it.


Eeems:
And I missed that as I already has simple installed, whoops!

Eeems (approved):

PR #736, opened 2023-09-04
Update YAFT to v0.0.10

FouzR:
- Physical keyboard key repeat support
- Handle rotation when folio is attached/removed
- Support for F1-F12 keys
- Use shell login to make sure to load user profile


Eeems (changes requested):
It looks like the `yaft-256color` file is no longer being provided, but it's still the default `TERM` value for yaft. This is breaking opening nano, vi, etc which was previously working due to having that included with the correct `TERMINFO` set. The correct `TERMINFO` is now set though.

Eeems:
@FouzR poke

timower:
For the terminfo files, I moved the install logic to libYaft: https://github.com/timower/rM2-stuff/blob/4174e7c70645285be47f7ab20d83e76fc8fd3835/libs/libYaft/CMakeLists.txt
This means you'll have to `cd libs/libYaft && make install` as well as just installing yaft.

FouzR:
Alright, so I installed the .ipk from the actions

All the noted changes seem to be working, but I don't know what the function keys do except in VIM =P

Eeems (approved):

PR #741, opened 2023-09-16
Don't overwrite other LD_PRELOAD env values

Eeems:
This allows other `/opt/etc/xochitl.env.d/*.env` files to set `$LD_PRELOAD` to something without losing their value.


Eeems:
This is not tested yet.

raisjn (dismissed):

raisjn (approved):

PR #742, opened 2023-09-23
linux-stracciatella: update sources to RM1XX_5.4.70_v1.3.4 and add wireguard

Etn40ff:
Synced upstream to tag RM1XX_5.4.70_v1.3.4.

Usb network is now usb1.

It seems to work on RM2 2.15.1.1189 without keyboard folio; it needs testing with 3.x and on RM1

EDIT: now we build also wireguard as a module


Etn40ff:
I just tested wg using #743 and the Demo Server available at https://www.wireguard.com/quickstart/#demo-server. Everything seems in order.

Eeems:
> It seems to work on RM2 2.15.1.1189 without keyboard folio; it needs testing with 3.x and on RM1

No testing is currently needed for 3.x as toltec doesn't yet support it. Everything will be tested for 3.x as part of adding support for it.

Eeems:
I've been trying to test this on my rM1, but I'm unable to due to the root partition not having enough space to extract the kernel. It looks like the kernelctl archive for this is 11MB while the stock one is 4MB.

Upon inspecting the kernel package it appears that every single one of the .ko files are larger than the stock ones. Some of the worst offenders are:

- kernel/net/wireless/cfg80211.ko - stock: 362kB, stracciatella: 6MB
- kernel/drivers/net/wireless/broadcom/brcm80211/brcmfmac/brcmfmac.ko - stock: 283kB, stracciatella: 4.7MB
- kernel/drivers/input/touchscreen/cyttsp5.ko - stock: 81kB, stracciatella: 722.4kB
- kernel/drivers/net/wireless/broadcom/brcm80211/brcmutil/brcmutil.ko - stock: 11kB, stracciatella: 204.6kB
- kernel/drivers/input/touchscreen/cyttsp5_loader.ko - stock: 15kB, stracciatella: 196kB
- kernel/drivers/input/touchscreen/cyttsp5_i2c.ko - stock: 7kB, stracciatella: 171.6kB
- kernel/crypto/aes_generic.ko - stock: 29kB, stracciatella: 139.5kB


I've already gone through the exercise of moving systemd-journald logs off of the root partition and into the home partition, so I'm not really sure what else would be reasonable to remove.

Eeems:
Adding [INSTALL_MOD_STRIP=1](https://mjmwired.net/kernel/Documentation/kbuild/kbuild.txt#148) to strip the debug symbols would probably reduce the size to something closer to what the stock one has.

Eeems:
It does not seem to be happy about finding the correct strip executable. It might make more sense to just strip the ko files as a manual step before compressing them.

Etn40ff:
On my machine I get a weird error:
```strip: Unable to recognise the format of the input file `/tmp/toltec/build/package/linux-stracciatella/rm1/src/staging/lib/modules/5.4.70/kernel/crypto/aes_generic.ko'```
Is it complaining because of cross compilation?

**Comment by Eeems** [comment]
> On my machine I get a weird error: `` strip: Unable to recognise the format of the input file `/tmp/toltec/build/package/linux-stracciatella/rm1/src/staging/lib/modules/5.4.70/kernel/crypto/aes_generic.ko' `` Is it complaining because of cross compilation?

Correct, it's trying to use the x64 strip to strip an arm binary. I have a pending change that may fix it, but github is having issues atm.

**Comment by Etn40ff** [comment]
It is still failing on my laptop: 
```
$ make linux-stracciatella
./scripts/package_build.py  "linux-stracciatella"
The build directory 'build/package/linux-stracciatella' for recipe 'linux-stracciatella' already exists.
Would you like to [c]ancel, [r]emove that directory, or [k]eep it (not recommended)? [C/r/k] r
[    INFO] toltec.builder: linux-stracciatella [rm1]: Fetching source files
[    INFO] toltec.builder: linux-stracciatella [rm1]: Preparing source files
[    INFO] toltec.builder: linux-stracciatella [rm1]: Building artifacts
[    INFO] toltec.builder: linux-stracciatella [rm1] (linux-stracciatella): Packaging build artifacts
[    INFO] toltec.builder: linux-stracciatella [rm1] (linux-stracciatella): Only showing up to 50 lines of context. Use --verbose for the full output.
[   ERROR] toltec.builder: linux-stracciatella [rm1] (linux-stracciatella): make[1]: arm-remarkable-linux-gnueabihf-gcc: No such file or directory
[   ERROR] toltec.builder: linux-stracciatella [rm1] (linux-stracciatella): /bin/sh: line 1: arm-remarkable-linux-gnueabihf-strip: command not found
[   ERROR] toltec.builder: linux-stracciatella [rm1] (linux-stracciatella): make[2]: *** [scripts/Makefile.modinst:33: crypto/aes_generic.ko] Error 127
[   ERROR] toltec.builder: linux-stracciatella [rm1] (linux-stracciatella): make[1]: *** [Makefile:1354: _modinst_] Error 2
[   ERROR] toltec.builder: linux-stracciatella [rm1] (linux-stracciatella): package() failed
```

The path seems to be wrong

Etn40ff:
Is it possible that `package()` is not run in the docker image? I ran ` find / -iname \*-gcc` and got a listing of the gcc's I have on my system.

Eeems:
> Is it possible that `package()` is not run in the docker image? I ran ` find / -iname \*-gcc` and got a listing of the gcc's I have on my system.

Ah, yes. That is the case as documented here: https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#package-function

Since this is the case, I would recommend moving the make install to the `build()` method to a folder inside the srcdir, and then just copying that over as part of `package()`. It's a little odd that `make install` is where the strip is done.

Etn40ff:
I just realized that the config file they provided for rm1 has `CONFIG_DEBUG_INFO=y` while it is not set for rm2. I guess that this is where the problem lies. Let me try to get rid of it

Etn40ff:
@Eeems Can you try now?

Eeems:
Good catch! This does reduce it to near the same size as the stock kernel and allows it to be activated with kernelctl. I'm running it now with nothing obviously wrong.

Eeems (approved):

PR #744, opened 2023-10-01
Add package for keyd

kiwiz:
This PR adds `keyd`, a tool for remapping keys. It has been tested on an rM2.

No configuration is currently bundled. My intent is to create a separate `keyd-configs` package with community configs (and possibly a mechanism to automatically apply the right one for the active keymap).


package/keyd/default.conf
Eeems, line 1: This probably should be rm2 only.
You can accomplish this by setting [archs](https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#archs-field) to rm1 and rm2, and then only installing this in `package()` if `$arch` is rm2.

kiwiz: `keyd` will fail to launch without a config. I can create an empty config for `rm1`:
```
[ids]
```

or just update the default config for both to:

```
[ids]
# 2edd:0001 # type folio
```

kiwiz: I opted for the latter, since users will probably want to set up their own configs anyway.

Eeems: > I opted for the latter, since users will probably want to set up their own configs anyway.

I'm not sure all users will. I'd expect that many users would expect this to work out-of-box on the rM2. Since this does nothing out of box now, it should not be enabled by default, and instead provide information on how to enable the service if it isn't running, restart the service if it's running, and provide a message on how to configure it.

package/keyd/package
```diff
@@ -0,0 +1,51 @@
...
+    install -D -m 755 -t "$pkgdir"/opt/bin "$srcdir"/bin/keyd
+    install -D -m 644 -t "$pkgdir"/etc/systemd/system "$srcdir"/keyd.service
+    install -D -m 644 -t "$pkgdir"/opt/etc/keyd "$srcdir"/default.conf
+}
+
+configure() {
+    systemctl daemon-reload
+    systemctl enable keyd
+    systemctl start keyd
+}
```
Eeems:
```suggestion
configure() {
    systemctl daemon-reload
    if is-active keyd; then
        systemctl restart keyd
    elif ! is-enabled "$pkgname.service"; then
        echo ""
        echo "Run the following command(s) to enable the keyd service"
        how-to-enable "$pkgname.service"
        echo ""
        echo "You will need to modify /opt/etc/keyd/default.conf before starting the service"
    fi
}
```

package/keyd/default.conf
kiwiz: Fair enough. In that case, I'll remove the default config and use the same pattern as other optional services.

kiwiz: ... or I'll just apply your suggested change.

Eeems:
I'll have to give this a test on my rM1 to make sure it works as expected. I assume you've tested the latest changes on your rM2 to make sure it works as expected?

kiwiz:
> I'll have to give this a test on my rM1 to make sure it works as expected. I assume you've tested the latest changes on your rM2 to make sure it works as expected?

Yep, all the keys appear to remap properly.

Testing tips:
* Remaps should be as simple as:
```
[main]
a = x # remaps the a key to output x
```
* Output of `keyd` should give you device ids, or you can just specify `*`

Eeems:
image
So it looks like this won't work on the rM1 in 2.15.1. This will also fail to work on the rM2 for the same reason.

The keyd group also doesn't exist, and the default configuration expects it to exist. This will either need to fixed by the install scripts, or the configuration changed to not use the group.

package/keyd/package
```diff
@@ -0,0 +1,58 @@
...
+    elif ! is-enabled "$pkgname.service"; then
+        echo ""
+        echo "Run the following command(s) to enable the keyd service"
+        how-to-enable "$pkgname.service"
+        echo ""
+        echo "You will need to modify /opt/etc/keyd/default.conf before starting the service"
+    fi
+}
+
+preremove() {
```
Eeems, line 52: This will need to be updated to also disable the service if it's enabled, It could be enabled but not active.

kiwiz:
Do you know what version the uinput module was enabled from? Should a version check be added to the package script?

Eeems:
Not offhand, it was one of the earlier 3.x versions I believe.
A version check doesn't make too much sense in the package, in reality we need to better support version limiting in the toltec repos themselves.

At this point, we will have to hold off on merging this until 3.x support is in progress.

FouzR:
I just want to note that, when I installed the ipk from the build artifact, in the `keyd.service`, the expected line for ExecStart is `/opt/bin/keyd`. However, when I installed it, it was `/usr/bin/keyd`

So either we add a symlink, or try and fix this(which shouldn't happen considering it was _supposed_ to be /opt/bin/keyd)

Eeems:
> I just want to note that, when I installed the ipk from the build artifact, in the `keyd.service`, the expected line for ExecStart is `/opt/bin/keyd`. However, when I installed it, it was `/usr/bin/keyd`
> 
> So either we add a symlink, or try and fix this(which shouldn't happen considering it was _supposed_ to be /opt/bin/keyd)

No we wouldn't want to have it be in /usr/bin. That's very odd that the file provided isn't being used. My guess is that the keyd source has its own copy of the service file that is replacing the one included in this repo.

FouzR:
Yup, so I thought, why not copy over keyd.service later in the package code instead of at the very beginning(So maybe put it in a temp folder first)
So maybe just copy it over to `/etc/systemd/system/` in configure()

Eeems:
I haven't looked too closely, but if there is a keyd.service file in the source that's overwriting the one included in the package folder, changing the order you install the files from the source directory wont do anything. The file isn't there to copy, because it's been replaced. The file will need to be renamed in this repo, and then renamed as part of package() when it's installing it to the final location in the package. Ideally the toltec build system (and toltecmk) should be updated to detect this kind of thing and error if there are files that are going to be clobbered.

Etn40ff:
Checking for clobbered files it is probably hard: in this particular instance there is no overlapping in the sources, the problem arises at compiel time: https://github.com/rvaiya/keyd/blob/master/Makefile#L37

Etn40ff:
In this particular case having a separate keyd.service is not needed: it suffices to add `PREFIX=/opt/` to the build environment

Eeems:
> In this particular case having a separate keyd.service is not needed: it suffices to add `PREFIX=/opt/` to the build environment

A tiny bit more than that was required to do it properly, but it should be working now.

Etn40ff:
This is still not working: the binary is `/opt/bin/keyd` but the service calls `/opt/usr/bin/keyd`. I think you are over engineering it a little, setting `PREFIX` should take care of the path in keyd.service

Eeems:
> This is still not working: the binary is `/opt/bin/keyd` but the service calls `/opt/usr/bin/keyd`. I think you are over engineering it a little, setting `PREFIX` should take care of the path in keyd.service
> 

It does not, which is why I added the sed line. I forgot that the default prefix is /usr. The latest code does, but we are dealing with the [v2.4.3 tag](https://github.com/rvaiya/keyd/tree/v2.4.3) which is missing some build process fixes.

Etn40ff:
>  The latest code does, but we are dealing with the [v2.4.3 tag](https://github.com/rvaiya/keyd/tree/v2.4.3) which is missing some build process fixes.

This is annoying

Eeems:
> > The latest code does, but we are dealing with the [v2.4.3 tag](https://github.com/rvaiya/keyd/tree/v2.4.3) which is missing some build process fixes.
> 
> This is annoying

It is :laughing: , the next time they push a release it will no longer be an issue.

Eeems:
For future reference when we get here, I think this is a good/sane default for the US layout. It at least makes the legends match what the keyboard actually outputs, and exposes some needed keys for terminal usage.

```
[ids]
2edd:0001 # type folio

[main]
end = layer(opt)
leftshift+end = layer(opt+shift)
= = -
\ = ~

[shift]
= = _
\ = `

[opt]
1 = esc
= = =
\ = \
0 = +
, = [
. = ]
right = end
left = home

[opt+shift]
\ = |
, = {
. = }
right = pagedown
left = pageup
```

Xochitl doesn't seem completely happy with allowing +/= on the 0/- keys though.

FouzR:
Keyd seems to freak out on boot and has key-repeats even when the key is clicked once.. I have to restart the service to fix it

Eeems:
> Keyd seems to freak out on boot and has key-repeats even when the key is clicked once.. I have to restart the service to fix it

I haven't experienced this. What settings are you using?

FouzR:
```
[ids]
2edd:0001 # type folio

[main]
capslock = overload(control, esc) # Sets the capslock to act as Esc when tapped and Control when held
leftcontrol = overload(control, capslock) # similar or left control

# below 2 lines setup the layers for
# Opt and AltOpt on the Type Folio
end = layer(opt)
rightalt = layer(alternate)

[opt]
1 = esc
, = [
. = ]
/ = \
= = -
right = end
left = home
up = pageup
down = pagedown
backspace = delete

[alternate:G]
1 = f1
2 = f2
3 = f3
4 = f4
5 = f5
6 = f6
7 = f7
8 = f8
9 = f9
0 = f10
= = f11
backspace = f12

```

That’s my config fire

Eeems:
Have you checked the keyd logs to see if it has anything useful? It works by grabbing all other physical keyboard input and creating it's own evdev device with uinput to emit events from. uinput handles automatically emitting the key repeat events. Because of this, keyd might just have missed the event that the key was released.

gustaphe:
For the scandinavian keyboard (and others that need non-ascii input keys), one would need xkbmap to make the compose-hack work.

plredmond:
I've tested https://github.com/toltec-dev/toltec/pull/744/commits/a2ab07855dfebca2f1d845a2d5b1527edab7ede6 (build [9670591399](https://github.com/toltec-dev/toltec/actions/runs/9670591399)) on an rm2 at `3.3.2.1666`.

<details><summary>Upon installation, it instructed me to disable the shutdown target, which seems not relevant.</summary>

```shell-session
reMarkable: ~/ opkg install keyd_2.4.3-1_rmall.ipk 
Installing keyd (2.4.3-1) to root...
Configuring keyd.

Run the following command(s) to enable the keyd service
$ systemctl disable --now shutdown.target
$ systemctl enable --now keyd

You will need to modify /opt/etc/keyd/default.conf before starting the service
```

</details>

<details><summary>I was able to set the `dvorak` layout with the following config file</summary>

```
[ids]
2edd:0001 # type folio

include layouts/dvorak

[global]
default_layout = dvorak

[main]
```

</details>

Having set the dvorak layout, the shortcuts in https://github.com/toltec-dev/toltec/pull/744#issuecomment-1890389124 do not work correctly. I'll post a full config later that makes those all work as expected.

Eeems:
> Upon installation, it instructed me to disable the shutdown target, which seems not relevant.

What launcher are you using? And what is the output of `toltecctl status`?

> I was able to set the `dvorak` layout with the following config file
> 
> Having set the dvorak layout, the shortcuts in [#744 (comment)](https://github.com/toltec-dev/toltec/pull/744#issuecomment-1890389124)) do not work. I'll post a config later to make them work again.

If you have a configuration to change the layout to dvorak, it would conflict with the settings I provided. I would recommend reviewing the keyd logs to see if it complains about anything (`journalctl -ae keyd`), and then you'd need to modify the dvorak layout instead of using the configration.

plredmond:
> What launcher are you using?

```
reMarkable: ~/ opkg list-installed remux
remux - 0.3.0-4
```


> And what is the output of toltecctl status?

```
reMarkable: ~/ toltecctl status
Enabled: Yes
Supported: Yes
Branch: stable
Model: rm2
OS: 3.3.2.1666
```

> reviewing the keyd logs to see if it complains about anything

Nothing beyond the usual startup messages.
```
[[0;1;39m-- Reboot --[[0m
Jun 27 21:33:32 reMarkable systemd[1]: Started key remapping daemon.
Jun 27 21:33:32 reMarkable keyd[188]: CONFIG: parsing /opt/etc/keyd/default.conf
Jun 27 21:33:32 reMarkable keyd[188]: Starting keyd v2.4.3 ()
Jun 27 21:33:32 reMarkable keyd[188]: DEVICE: ignoring 0fac:1ade  (keyd virtual pointer)
Jun 27 21:33:32 reMarkable keyd[188]: DEVICE: match    2edd:0001  /opt/etc/keyd/default.conf    (rM_Keyboard)
Jun 27 21:33:32 reMarkable keyd[188]: DEVICE: ignoring 056a:0000  (Wacom I2C Digitizer)
Jun 27 21:41:04 reMarkable keyd[188]: DEVICE: removed   2edd:0001 rM_Keyboard
Jun 27 21:53:58 reMarkable keyd[188]: DEVICE: match    2edd:0001  /opt/etc/keyd/default.conf    (rM_Keyboard)
```


> you'd need to modify the dvorak layout instead of using the configration.

Yes, I'll do that and post it.

Eeems:
`launcherctl status` would be another output you could share to show what launcher you are on. Forgot to ask for that instead. What you provided works though.

Everything looks fine here, so it would be just that the two configurations are compatible with each other, as one overrides the other.

Eeems (changes requested):
Since uinput isn't available on OS 2, this will need to be limited to the [rmallos3 architecture](https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#archs-field), so it's only available on 3.x.

It would probably be worth sorting out how we intend on other packages providing configuration for keyd, but that is probably better served by the first package who does.

plredmond:
here is my dvorak keyd config, along with an image of the keyboard with the symbols overlaid

https://gist.github.com/plredmond/6da75360e660a7e240a9013f56af0603

346374933-3dd8e10c-3f75-46fb-8f41-87f77b1a099e

Eeems (approved):

PR #746, opened 2023-10-08
rmfm: Update to 1.5.0

TheLastZombie:
Changelog:

- Fixed reloading in a deleted directory resulting in a crash
- Append /opt/bin to PATH if not already included[^1]
- The installation instructions, license and e-mail address were updated/changed

Commits/diff: https://forgejo.sny.sh/sun/rmFM/compare/1.4.0...1.5.0

[^1]: I also updated path_fix.patch accordingly


Eeems:
Just to warn you, but from what I understand, Unlicensed reverts to All Rights Reserved in places like Germany. You may want to stick with MIT.

Eeems (approved):

PR #757, opened 2023-11-29
rmfakecloudctl: Don't error if there are no metadata files

Eeems:


Eeems:
Tested and works on my new rM2 :sunglasses:

raisjn (approved):

PR #764, opened 2023-12-06
Update tilem to v0.1.1

Eeems:
[Release notes v0.1.0](https://github.com/timower/rM2-stuff/releases/tag/v0.1.0)
[Release notes v0.1.1](https://github.com/timower/rM2-stuff/releases/tag/v0.1.1)


Eeems:
@timower For some reason, the icon that gets included is broken.

2023-12-06T205544 318
tilem

timower:
```
> file opt/etc/draft/icons/yaft.png
opt/etc/draft/icons/yaft.png: ASCII text
> cat opt/etc/draft/icons/yaft.png
version https://git-lfs.github.com/spec/v1
oid sha256:5e34a13870db6d6cffb1f3801678b93cfd91a93cfc64465c5dcc0b663601630e
size 17040
```

I use git LFS now, so `git lfs pull` might be required

Eeems:
Tested on my new rM2 and it works :sunglasses:

package/tilem/package
```diff
@@ -5,23 +5,37 @@
 pkgnames=(tilem)
 pkgdesc="TI-84+ calculator emulator"
 url=https://github.com/timower/rM2-stuff/tree/master/apps/tilem
-pkgver=0.0.7-3
-timestamp=2021-04-30T10:42Z
+_tag=0.1.1
```
bkirwi, line 8: It would be somewhat more deterministic to depend on a commit sha instead of a tag, in case the tag ever gets deleted & recreated. 0.1.1 corresponds to `a383c22ff2ea7f4abe7aaaf936ed055a34f8523b` in case you're interested in making that change!

bkirwi (approved):
Looks good - left one suggestion anyways!

package/tilem/package
Eeems: While this is true, we've been happy to use tags as a standard throughout toltec, As they are supposed to be deterministic and not moved if you are using git the way it's meant to be used. So far, I haven't noticed timower moving tags, so I'm not too concerned.

Using tags also makes it easier to update the package, as you just need to update the version number without looking, the hash, and the timestamp.

bkirwi: For sure. It does feel inconsistent to check hashes for fetching a tarball over HTTPS, but not a git fetch... but if this is already how many packages work, this is not the time / place to shake that out!

Eeems: That is a fair point. I'd be happy to approve and merge a new PR to switch this over to a commit hash if you'd like. I had forgotten that this was doing a git checkout.

PR #768, opened 2023-12-06
Update webinterface-wifi to v2.0.0

rM-self-serve:
[Release Notes v2.0.0](https://github.com/rM-self-serve/webinterface-wifi/releases/tag/v2.0.0)
[Docs](https://github.com/rM-self-serve/webinterface-wifi)

Quite a bit of new functionality. 

Should work on all devices+versions.

For anyone testing this, there are [test ssl certs in the repository](https://github.com/toltec-dev/toltec/pull/github.com/rM-self-serve/webinterface-wifi/tree/master/ssl_test_cert). It is necessary the wifi is connected and the web interface is accessible at 10.11.99.1. The following tests should suffice.

---
```
$ webint-wifi net-info
# ensure the wifi ssid is correct 
```
---
```
$ systemctl start webinterface-wifi
```
```
$ webint-wifi create-login
# enter a username/password 
```
```
$ webint-wifi edit
# change [undefined_networks].login_enforced to true
```
```
$ webint-wifi reload
```
Then navigate to the wifi ip address in a browser, enter the credentials, and download a file.

### DIFF
--- a/package/webinterface-wifi/package
+++ b/package/webinterface-wifi/package
@@ -18,12 +18,12 @@ _configdir="/home/root/.config/$_pkgname"
 _etcdir="/opt/etc/$_pkgname"
 
 source=(
-    "$url/archive/feb573ac5f16a75d3d7ce66b3957d5c753ad5561.zip"
+    "$url/archive/4513d5cbc5e323f2959987f3bc9e300b0aaddb19.zip"
     "$_pkgname-toltec.service"
 )
 
 sha256sums=(
-    516f4c1684b8d50fc8c9c63a7f11f3183b259ba52243b719a73a56519dc45947
+    a23c05faf4ccaafea9222255399c8c3a121079dd07970c8768b4315606eb7834
     SKIP
 )
 
--- a/package/webinterface-wifi/package
+++ b/package/webinterface-wifi/package
@@ -57,6 +57,11 @@ configure() {
 
     systemctl daemon-reload
 
+    if is-active "$pkgname"; then
+        echo "Restarting $pkgname"
+        systemctl restart "$pkgname"
+    fi
+
     echo ""
     echo "Run '\$ $pkgname' for usage information and a link to"
     echo "the documentation. You can also find the documentation locally"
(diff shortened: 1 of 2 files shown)


Eeems (changes requested):
Upon testing, trying to enable the service doesn't work as it's trying to run `/opt/bin/webinterface-wifi --run 80` which is no longer valid.

It may also be worth adding some more helpful output to `configure()` to point the user in the right direction for configuring it?

rM-self-serve:
Sorry for that oversight, the correct service file should now be installed.  

I mentioned where to find usage and docs locally and remotely.

Thanks for testing! I was not aware of the release window so my apologies for cutting it so close.

Eeems:
> Sorry for that oversight, the correct service file should now be installed.
> 
> I mentioned where to find usage and docs locally and remotely.

No worries, this is why we test :stuck_out_tongue: 

> 
> Thanks for testing! I was not aware of the release window so my apologies for cutting it so close.

No worries about the release window. I just picked the earlier one as I don't have to rely on another maintainer to test this.

package/webinterface-wifi/package
```diff
@@ -2,43 +2,65 @@
...
 configure() {
+    [[ -f "$_configdir/config.toml" ]] \
+        || cp "$_etcdir/docs/config.default.toml" \
+            "$_configdir/config.toml"
+
     systemctl daemon-reload
 
     echo ""
-    echo "Run the following command to use $pkgname"
-    how-to-enable "$pkgname.service"
```
Eeems, line 41: I think it's still worth leaving this in as it's standard output for toltec.

rM-self-serve:
Sounds good to me!

Eeems:
Okay, testing this, and I realized it doesn't automatically restart the service if it's already enabled and running. Since this isn't a launcher which could have various applications running in an unsaved state, it should be safe to restart the service.

```bash
if is-active webinterface-wifi.service; then
    systemctl restart webinterface-wifi
fi
```

Eeems (approved):

PR #770, opened 2023-12-06
Add package webinterface-onboot v1.2.2

rM-self-serve:
[Release Notes v1.2.2](https://github.com/rM-self-serve/webinterface-onboot/releases/tag/v1.2.2)
[Docs](https://github.com/rM-self-serve/webinterface-onboot/)

This program will start the web interface without the usb cable being plugged in. It can also keep the web interface accessible after unplugging the usb cable.

I am the author of this software. This should work on all devices+versions.

Has been verified on both devices up until v3.3 and the rM1 until v3.8, though the relevant part of both xochitl binaries seem the same. 

If applying other binary modifications such as [rM-hacks](https://github.com/mb1986/rm-hacks) or [ddvk's hacks](https://github.com/ddvk/remarkable-hacks), they will fail if webint-onboot has been applied. Thus, it is necessary to apply webint-onboot after applying the other modifications. If reverting other binary modifications after webint-onboot has been applied, it is necessary to revert webint-onboot before reverting these modifications.

webint-onboot will ensure the modifications are possible before performing them by checking for the "guilty" strings rather than a shasum. It will attempt to revert the modifications when uninstalled, if they are applied.

---

How it works:

Before xochitl starts:
- set WebInterfaceEnabled=true in /home/root/.config/remarkable/xochitl.conf
- give the usb0 network interface the ip address 10.11.99.1/32

For versions >= 2.15:
Requires simple binary hack.

In xochitl:
- change the string 'usb0' to 'usbF'
- change the string 'usb1' to 'usb0'

---

It also has built in functionality to keep the web interface accessible after unplugging the usb cord. 
Add the following line to /etc/ifplugd/ifplugd.action in order to run after dhcpd is stopped: 
- ip addr change 10.11.99.1/32 dev usb0

### DIFF
--- a/package/webinterface-onboot/package
+++ b/package/webinterface-onboot/package
@@ -13,8 +13,6 @@ maintainer="rM-self-serve <122753594+rM-self-serve@users.noreply.github.com>"
 license=MIT
 conflicts=(ddvk-hacks)
 
-_pkgalias="webint-onboot"
-
 source=(
     "$url"/archive/cdfe457435974f7ca309b1ac50f1b2ef67000813.zip
     "$_pkgname-toltec.service"
@@ -28,6 +26,9 @@ package() {
     install -D -m 755 -t "$pkgdir/opt/bin" "$srcdir/$_pkgname"
     install -D -m 644 "$srcdir/$_pkgname-toltec.service" \
         "$pkgdir/lib/systemd/system/$_pkgname.service"
+
+    touch "$srcdir"/emptyfile
+    install -D -m 666 -t "$pkgdir"/usr/share/toltec/reenable.d/"$_pkgname" "$srcdir"/emptyfile
 }
 
 configure() {
@@ -37,7 +38,7 @@ configure() {
     echo "Applying usb0 ip persistence"
     webinterface-onboot apply-prstip -y > /dev/null
     echo "Success"
-    if webinterface-onboot is-hack-vers > /dev/null; then
+    if webinterface-onboot is-hack-version > /dev/null; then
         echo
         echo "Applying binary modification"
         webinterface-onboot apply-hack -y > /dev/null
--- a/package/webinterface-onboot/package
+++ b/package/webinterface-onboot/package
@@ -4,53 +4,89 @@
 
 _pkgname='webinterface-onboot'
 pkgnames=("$_pkgname")
-pkgdesc="Start the web interface without the cable"
+pkgdesc="Start the web interface without the cable, on boot."
 url="https://github.com/rM-self-serve/$_pkgname"
 pkgver=1.2.2-1
 timestamp=2023-12-03T11:43:00Z
 section="utils"
 maintainer="rM-self-serve <122753594+rM-self-serve@users.noreply.github.com>"
 license=MIT
+conflicts=(ddvk-hacks)
 
 _pkgalias="webint-onboot"
 
 source=(
-    "$url"/archive/cbb2b936f0d27898701695f4d166dcf01b0174e2.zip
+    "$url"/archive/cdfe457435974f7ca309b1ac50f1b2ef67000813.zip
     "$_pkgname-toltec.service"
 )
 sha256sums=(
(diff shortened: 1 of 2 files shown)


package/webinterface-onboot/package
```diff
@@ -0,0 +1,56 @@
...
+            if ! webinterface-onboot revert-hack --reverse -y > /dev/null 2>&1; then
+                echo "Could not revert /usr/bin/xochitl"
+            fi
+        fi
+        echo "Success"
+    fi
+    if webinterface-onboot is-prstip-applied > /dev/null 2>&1; then
+        echo "Reverting /etc/ifplugd/ifplugd.action"
+        if ! webinterface-onboot revert-prstip -y > /dev/null 2>&1; then
+            echo "Could not revert /etc/ifplugd/ifplugd.action"
```
Eeems: Is this safe to ignore? I would assume we would want to error out and have the user investigate what went wrong, so that opkg keeps track of the partially broken state and will reattempt the next install.

package/webinterface-onboot/webinterface-onboot-toltec.service
Eeems, line 1: This file doesn't appear to be installed in `package()`

Eeems (changes requested):
I think it would be worth updating the ddvk-hacks package to detect this package and temporarily remove it for you. Otherwise, we'll just need to mark them as conflicting.

I also don't see how these changes are applied? I think we are missing the `configure()` step to automatically apply them.

The goal of toltec is to simplify the installation process for users as much as possible, and make it as close to one line as we can.

package/webinterface-onboot/package
rM-self-serve: Ah great catch, I thought I was copying ddvks _restore() fn but this is a different situation. 

I suppose that applies for the inner revert here as well:
```
    if webinterface-onboot is-hack-applied > /dev/null 2>&1; then
        echo "Reverting /usr/bin/xochitl"
        if ! webinterface-onboot revert-hack --backup -y > /dev/null 2>&1; then
            if ! webinterface-onboot revert-hack --reverse -y > /dev/null 2>&1; then
                echo "Could not revert /usr/bin/xochitl"
            fi
        fi
        echo "Success"
    fi
```

rM-self-serve:
I was not able to see the reinstall on upgrade functionality when upgrading from v2.10 -> v2.15. The package was installed but the modifications were not applied. I did see the file at v2.10:/usr/share/toltec/reenable.d/webinterface-wifi but not in v2.15. Is this something that works with packages from .cache/toltec?

Eeems:
> I was not able to see the reinstall on upgrade functionality when upgrading from v2.10 -> v2.15. The package was installed but the modifications were not applied. I did see the file at v2.10:/usr/share/toltec/reenable.d/webinterface-wifi but not in v2.15. Is this something that works with packages from .cache/toltec?

Is there anything in the `toltecctl reenable` output about this package? I would expect it to not be happy since it's not in toltec yet, so it wouldn't be able to reinstall the package. You can work around this by creating `/opt/etc/opkg.conf.d/16-local.conf` with the following content:
```
src/gz local-rmall file:///home/root/packages/rmall
```
Then you'll need to place the `rmall` folder from the build artifacts in `/home/root/packages` on your device and run `toltecctl generate-opkg-conf`. At this point, the package will be available by just running `opkg install webinterface-onboot` without having to point at the `ipk` file.

rM-self-serve:
With a21721d05a4f9a1c83fe45055456a4c346397e58, both files are modified upon re-enabling after an upgrade. 

The systemd service is not started though, is this standard?

onboot_reinstall_update

Eeems:
> With [a21721d](https://github.com/toltec-dev/toltec/commit/a21721d05a4f9a1c83fe45055456a4c346397e58), both files are modified upon re-enabling after an upgrade.
> 
> The systemd service is not started though, is this standard?

It's only restarted if a package that is being installed restarts it in `configure()`, which I don't believe any other than `display` (on the rM2) will currently.

Eeems (approved):

PR #772, opened 2023-12-06
Add package: signature-rm v1.0.2

rM-self-serve:
[Release Notes v1.0.2](https://github.com/rM-self-serve/signature-rM/releases/tag/v1.0.2)
[Docs](https://github.com/rM-self-serve/signature-rM)

Remove the signature from the bottom of emails sent from the device. It will also take a backup before the modification, and allow reverting the modification from the backup or by performing the modification in reverse.

I am the author of this software. This should work on all devices+versions.

It will ensure the modification is possible before performing it by checking for the "guilty" string rather than a shasum. It will attempt to revert the modification when uninstalled, if it is applied.

---

How it works:
In xochitl: 
- change 'Sent from my reMarkable' to '\0ent from my reMarkable'

### DIFF
--- a/package/signature-rm/package
+++ b/package/signature-rm/package
@@ -12,6 +12,7 @@ section="utils"
 maintainer="rM-self-serve <122753594+rM-self-serve@users.noreply.github.com>"
 license=MIT
 image=rust:v3.1
+conflicts=(ddvk-hacks)
 
 source=(
     "$url"/archive/b5561af4eb6a0f5aa6e98e1a1279066f0c4bd9b7.zip
--- a/package/signature-rm/package
+++ b/package/signature-rm/package
@@ -12,7 +12,7 @@ section="utils"
 maintainer="rM-self-serve <122753594+rM-self-serve@users.noreply.github.com>"
 license=MIT
 image=rust:v3.1
-conflicts=(ddvk-hacks)
+conflicts=(ddvk-hacks webinterface-onboot)
 
 source=(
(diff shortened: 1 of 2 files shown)


Eeems (changes requested):
Same comments as the webinterface-onboot package.

This should be automatically applied with `configure()`, and `ddvk-hacks`'s `configure()` should be updated to check for this, remove the signature, and then reapply it after it's done patching. That or this should be marked as `conflicts` with ddvk-hacks

rM-self-serve:
This can be tested with:
`signature-rm is-applied`

Reinstall on reenable:
sigrm_reinstall_update

rM-self-serve:
If both #770 and #772 are installed when the device is reenabled, its possible the first binary modification "fills up" (until xochitl is restarted) the root partition so that the second binary modification fails. Thus, make them conflict.

Eeems:
> If both #770 and #772 are installed when the device is reenabled, its possible the first binary modification "fills up" (until xochitl is restarted) the root partition so that the second binary modification fails. Thus, make them conflict.

Since this is also a concern for ddvk-hacks, could you open a separate issue to track this. I think we should explore moving xochitl off of the root partition for modifications. Or at the very least move the modified copy of it to `/opt` instead of running the main one.

Edit: I do think we need to resolve this before 3.x support. And since it's a pre-existing issue with our packages, I don't see a reason it should keep this from being merged.

Eeems (approved):

PR #777, opened 2023-12-06
Split stable notification into two steps

Eeems:
- Stop using `hub` and instead use the github cli
- Create new draft release as part of stable build
- Split out discord notification to its own action that triggers on release. This means someone needs to go and publish the draft release


Eeems:
I haven't tested this as there is no easy way in the current repo, and I don't have capacity to set up a standalone test repo for it right now.

I would appreciate someone else setting up a test repo though.

PR #778, opened 2023-12-08
Update ddvk-hacks: conflict with new hacks

rM-self-serve:
Conflict with https://github.com/toltec-dev/toltec/pull/770 and https://github.com/toltec-dev/toltec/pull/772.

### DIFF
--- a/package/ddvk-hacks/package
+++ b/package/ddvk-hacks/package
@@ -6,12 +6,13 @@ archs=(rm1 rm2)
 pkgnames=(ddvk-hacks)
 pkgdesc="Enhance Xochitl with additional features"
 url=https://github.com/ddvk/remarkable-hacks
-pkgver=39.01-1
+pkgver=39.01-2
 timestamp=2022-11-09T18:31:51Z
 section="readers"
 maintainer="Mattéo Delabre <spam@delab.re>"
 license=MIT
 flags=(nostrip)
+conflicts=(webinterface-onboot signature-rm)
 
 source=(https://github.com/ddvk/remarkable-hacks/archive/90e7e3e7ffc269373de191085453be50c9f8da0c.zip)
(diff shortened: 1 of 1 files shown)


Eeems:
Is the conflicts in the other packages not enough?

rM-self-serve:
Ah, I thought it was best practice to put conflicts on both packages.

Eeems:
> Ah, I thought it was best practice to put conflicts on both packages.

I would say it probably is best practice. I was under the impression we were going to try to resolve the conflict next, so I was just making sure there wasn't an issue this was trying to resolve.

rM-self-serve:
No issues other than the issue of ddvk-hacks failing to install/remove if webinterface-onboot or signature-rm has been installed.

Eeems:
> No issues other than the issue of ddvk-hacks failing to install/remove if webinterface-onboot or signature-rm has been installed.

Just to clarify, by this you mean that without those packages including conflicts it would fail, right? And now that they include it, you can't get them and ddvk-hacks to install at the same time?

rM-self-serve:
Yes, I do mean that without https://github.com/toltec-dev/toltec/pull/770 and https://github.com/toltec-dev/toltec/pull/772 including the conflicts, it would fail. 

The conflicts are working correctly when ddvk-hacks is installed then attempting to install https://github.com/toltec-dev/toltec/pull/770 or https://github.com/toltec-dev/toltec/pull/772.
onboot_conflict

But, when https://github.com/toltec-dev/toltec/pull/770 or https://github.com/toltec-dev/toltec/pull/772 is installed then attempting to install ddvk-hacks, ddvk-hacks will fail in configure() as the previous packages have changed the shasum.

 ddvk-noconflict

Eeems:
Alright, so we do need to make sure that ddvk-hacks has conflicts as well to avoid this. Thanks!

Eeems (approved):

PR #785, opened 2023-12-12
Update puzzles to 0.2.4

Eeems:
Only includes rmkit updates, which contain various fixes that are mostly invisible to users at the moment.


Eeems:
Tested with no issues on my rM2

raisjn:
does this work properly? I'm surprised by the reordering of the environment vars on the make line

raisjn (approved):

Eeems:
> does this work properly? I'm surprised by the reordering of the environment vars on the make line

Yup! Make supports either variables from the environment, as an argument, and some other places: https://www.gnu.org/software/make/manual/make.html#Origin-Function

I made the change to be consistent with the build workflow on the puzzles repo: https://github.com/mrichards42/remarkable_puzzles/blob/main/.github/workflows/build.yml

PR #786, opened 2023-12-12
Add upower and dependencies

Eeems:
https://upower.freedesktop.org/

Unfortunately, due to how the toolchain currently is set up, I've had to install libglib and libgudev to their own folders as they contain libraries that are already provided in the base OS, or in entware. We'll have to sort out a better longer term solution for this.

This also introduces a change to toltec-base to load dbus configuration from `/opt/share/dbus-1`, and updates `tarnish` to use the new location.


Eeems:
@raisjn do you have any opinions on this? I don't completely feel great about providing libraries a second time, but I don't really see a way around it right now with the deficiencies in the toolchain, as well as the issue we have trying to link against the glibc that comes with the device, but putting our executables and libraries in `/opt` alongside entware packages that link against entware's glibc version.

raisjn (dismissed):

raisjn:
this is fine with me, I don't have a strong opinion on the libraries. when it comes up again, we can think about how to address it properly.

Eeems:
Thanks! I'm leaving this in draft until it can be tested on a rM2, as I'm concerned that it may not work as expected due to the testing with udev I did in the past for oxide: https://github.com/Eeems-Org/oxide/issues/280#issuecomment-1402931866

Eeems:
It looks like this is no longer working as expected on 3.x. I'm not entirely sure what the issue is, but the only thing I'm seeing in the log is `name lost, exiting`. So this might be related to dbus.

Eeems:
The build The build seems to have broken since the build system switched to toltecmk. I'm not sure what the issue is though, as only the final udev package fails, and I've confirmed that the files it's expecting are in place before the build code in the recipe is called.

PR #792, opened 2023-12-17
Add tool to handle entware services

### LINKED ISSUE #791: Add systemd generator for entware init scripts to toltec
**Package information**
We should have a systemd generator for entware init scripts (found in /opt/etc/init.d).

**Reason for addition**
This will allow doing things like installing entware's cron package and then setting it up to run on startup.

**More context**
This will remove the need for systemd specific packages to be created.

Eeems:
- Adds new package named `entware-rc` that adds the `rcctl` command to manage entware services
- Adds new template unit that `rcctl` uses to start entware services with systemd
- Adds bash completions for `rcctl`

image


Eeems:
It's probably worth pulling this into a different package that `toltec-base` depends on, since it should be a `rmall` package.

raisjn:
what's the motivation for this?

raisjn (approved):

Eeems:
> what's the motivation for this?

Allow a user to easily manage entware services with systemd since we don't use the entware init stuff. This avoids spinning up their init binary as a subprocess to manage their services, and instead uses systemd for everything.

PR #793, opened 2023-12-17
Add toltec-python package

Eeems:
Due to https://github.com/Nuitka/Nuitka/issues/2440 it may be worth adding a normal build of python to the device. I'll be using this to test using nuitka to build binaries for the device to see if it works as expected. We may not even want to merge this PR and instead create an arm toolchain image that has python that can just use nuitka to build things that will run natively.


Eeems:
Closing in favour of https://github.com/marketplace/actions/remarkable-nuitka-build-action

PR #794, opened 2023-12-18
webinterface-onboot wait for usb0

rM-self-serve:
Ensure the webinterface-onboot daemon will wait for usb0 to be ready before giving it an ip address.


Eeems (approved):
lgtm

PR #799, opened 2023-12-23
Add debootstrap and debian-chroot packages

Eeems:
This adds the following new packages:
 1. debootstrap - This will let you create a debian install in a subdirectory
 2. debian-chroot - This will install a minimal debian install to /opt/lib/debian-chroot and then chroot into it, or run commands in it's chroot

 debian-chroot example usage:
 ```shell
$ debian-chroot # Will enter an interactive bash session in the chroot for you
$ debian-chroot apt-install nano # Will install nano in the chroot for you
$ debian-chroot bash -x <<'EOT'
  set -e
  apt install -y \
    python3-dev \
    libffi-dev \
    python3-pip \
    python3-venv \
    patchelf \
    ccache \
    upx
  python3 -m venv /opt/lib/nuitka
  source /opt/lib/nuitka/bin/activate
  pip install \
    --extra-index-url https://wheels.eeems.codes/ \
    nuitka
EOT # Will install nuitka in the chroot for you in a python venv
 ```


Eeems:
The primary use-case for this would be to run things that can't be run with entware. The full install size is around 600MB, which is much larger than the package, due to using debootstrap to handle the actual install. I don't really recommend using this on a real device, but it's quite useful for helping automate builds using qemu.

Etn40ff:
Installation fails on rm2: 
```
reMarkable: /tmp/ opkg install rmall/debian-chroot_0.0.0-1_rmall.ipk rmall/debootstrap_1.0.133-1_rmall.ipk
Package debian-chroot (0.0.0-1) installed in root is up to date.
Package debootstrap (1.0.133-1) installed in root is up to date.
Configuring debian-chroot.
/opt/bin/debian-chroot: line 3: /home/root/.config/debian-chroot.conf: No such file or directory
Collected errors:
 * pkg_run_script: package "debian-chroot" postinst script returned status 1.
 * opkg_configure: debian-chroot.postinst returned 1.
 ```

**Comment by Etn40ff** [comment]
Also, `/opt/usr/sbin` is not in `$PATH`

**Comment by Eeems** [comment]
> Also, `/opt/usr/sbin` is not in `$PATH`

https://github.com/toltec-dev/toltec/blob/stable/package%2Ftoltec-bootstrap%2Ftoltecctl#L189

~~It should be, if it's not on your path, something has removed it.~~

I'll have to look at this, as it was on my path.

**Comment by Etn40ff** [comment]
here is some more:
```
reMarkable: /tmp/ opkg install rmall/debian-chroot_0.0.0-1_rmall.ipk
Installing debian-chroot (0.0.0-1) to root...
Configuring debian-chroot.
/opt/usr/sbin/debootstrap: line 21: /usr/share/debootstrap/functions: No such file or directory
cat: can't open '/opt/lib/debian-chroot/debootstrap/debootstrap.log': No such file or directory
Collected errors:
 * pkg_run_script: package "debian-chroot" postinst script returned status 1.
 * opkg_configure: debian-chroot.postinst returned 1.
 ```

Eeems:
~~Okay, that should fix it.~~

Uninstall is broken at the moment, as umount doesn't have the `-q` flag

Eeems:
Uninstall currently requires being run twice, as it fails the first time. I assume something with umount is the issue, but I don't really have time to test a fix for it at the moment.

Eeems:
I've fixed the uninstall issue and this package is ready to be merged.

Jayy001 (approved):

Issue #803, opened 2023-12-25
linux-stracciatella: add 'CONFIG_MODULE_UNLOAD=y'


gbyl:
**Device/Package information:**
 - Device: reMarkable 1
 - OS version: 2.15.1.1189
 - Stock Kernel version: Linux reMarkable 5.4.70-v1.2.5-rm10x #&#x2060;1
 - Stracciatella Kernel version: Linux reMarkable 5.4.70 #&#x2060;2
 - Package: linux-stracciatella

**Describe the issue**
The stock rM1 kernel does not allow unloading modules.

Packages, [such as KOReader](https://github.com/koreader/koreader/issues/11257#issuecomment-1868607390), that rely on `modprobe brcmfmac` and `modprobe -r brcmfmac` to enable and disable Wi-Fi (and to detect Wi-Fi status) do not work as expected.

**Workaround**

1. Blacklist the Broadcom FullMAC WLAN driver modules `/etc/modprobe.d/blacklist.conf`
  ```
  blacklist brcmfmac
  blacklist brcmutil
  blacklist cfg80211
  ```

2. Disable DHCP (and other services) to prevent startup delays and conflicts
  ```
  systemctl disable dhcpcd disable-wowlan manual-sync wpa_supplicant
  ```

3. Modify Wi-Fi enable script `/opt/koreader/enable-wifi.sh`
  ```sh
  #!/bin/sh
  if ! lsmod | grep -q brcmfmac; then
    modprobe brcmfmac
    sleep 1
  fi

  ip link set wlan0 up
  systemctl start dhcpcd
  wpa_supplicant -B -i wlan0 -C /var/run/wpa_supplicant
  ```

4. Modify Wi-Fi disable script `/opt/koreader/disable-wifi.sh`
  ```sh
  #!/bin/sh
  wpa_cli terminate
  systemctl stop dhcpcd
  ip link set wlan0 down
  modprobe -r brcmfmac 2> /dev/null
  ```

**Side Benefits**
With this workaround I am seeing an increase in battery life, and USBnet startup of around 2 seconds

**Questions**
1. Does the rM2 kernel allow unloading modules?
2. @Etn40ff Are there any disadvantages to adding `CONFIG_MODULE_UNLOAD=y` to linux-stracciatella?

Eeems:
Thank you for reporting this compatibility issue. Is there a reason you haven't opened up an upstream issue/PR to link to this one?

[...]

PR #806, opened 2023-12-27
toltec-base: Disable busybox-ifplugd@usb1.service

gbyl:
Disable `busybox-ifplugd@usb1.service` in rM1 to avoid the following errors on boot:

```
systemd[1]: busybox-ifplugd@usb1.service: Bound to unit sys-subsystem-net-devices-usb1.device, but unit isn't active.
systemd[1]: Dependency failed for ifplugd on usb1.
systemd[1]: busybox-ifplugd@usb1.service: Job busybox-ifplugd@usb1.service/start failed with result 'dependency'.
```

Discussed in https://github.com/toltec-dev/toltec/issues/803#issuecomment-1869778997

### DIFF
--- a/package/toltec-base/package
+++ b/package/toltec-base/package
@@ -33,21 +33,21 @@ configure() {
     disable-unit update-engine.service
     if [[ "$arch" == "rm1" ]]; then
         echo "Disabling usb1 network device to avoid long boots"
-        ! is-masked sys-subsystem-net-devices-usb1.device && systemctl mask sys-subsystem-net-devices-usb1.device
-        ! is-masked busybox-ifplugd@usb1.service && systemctl mask busybox-ifplugd@usb1.service
+        ! systemctl is-masked sys-subsystem-net-devices-usb1.device && systemctl mask sys-subsystem-net-devices-usb1.device
+        ! systemctl is-masked busybox-ifplugd@usb1.service && systemctl mask busybox-ifplugd@usb1.service
     elif [[ "$arch" == "rm2" ]]; then
         echo "Enabling usb1 network device to ensure usb SSH works"
-        is-masked sys-subsystem-net-devices-usb1.device && systemctl unmask sys-subsystem-net-devices-usb1.device
-        is-masked busybox-ifplugd@usb1.service && systemctl unmask busybox-ifplugd@usb1.service
+        systemctl is-masked sys-subsystem-net-devices-usb1.device && systemctl unmask sys-subsystem-net-devices-usb1.device
+        systemctl is-masked busybox-ifplugd@usb1.service && systemctl unmask busybox-ifplugd@usb1.service
     fi
 }
 
 postremove() {
-    if is-masked sys-subsystem-net-devices-usb1.device; then
+    if systemctl is-masked sys-subsystem-net-devices-usb1.device; then
         systemctl unmask sys-subsystem-net-devices-usb1.device
         systemctl unmask busybox-ifplugd@usb1.service
     fi
-    if ! is-enabled "update-engine.service"; then
+    if ! systemctl is-enabled update-engine.service; then
         systemctl enable update-engine
     fi
 }
--- a/package/toltec-base/package
+++ b/package/toltec-base/package
@@ -33,18 +33,28 @@ configure() {
     disable-unit update-engine.service
     if [[ "$arch" == "rm1" ]]; then
         echo "Disabling usb1 network device to avoid long boots"
-        ! is-masked sys-subsystem-net-devices-usb1.device && systemctl mask sys-subsystem-net-devices-usb1.device
-        ! is-masked busybox-ifplugd@usb1.service && systemctl mask busybox-ifplugd@usb1.service
+        if ! is-masked sys-subsystem-net-devices-usb1.device; then
+            systemctl mask sys-subsystem-net-devices-usb1.device
+        fi
+        if ! is-masked busybox-ifplugd@usb1.service; then
+            systemctl mask busybox-ifplugd@usb1.service
+        fi
     elif [[ "$arch" == "rm2" ]]; then
         echo "Enabling usb1 network device to ensure usb SSH works"
-        is-masked sys-subsystem-net-devices-usb1.device && systemctl unmask sys-subsystem-net-devices-usb1.device
-        is-masked busybox-ifplugd@usb1.service && systemctl unmask busybox-ifplugd@usb1.service
+        if is-masked sys-subsystem-net-devices-usb1.device; then
+            systemctl unmask sys-subsystem-net-devices-usb1.device
+        fi
+        if  is-masked busybox-ifplugd@usb1.service; then
+            systemctl unmask busybox-ifplugd@usb1.service
+        fi
     fi
 }
 
 postremove() {
     if is-masked sys-subsystem-net-devices-usb1.device; then
         systemctl unmask sys-subsystem-net-devices-usb1.device
+    fi
+    if is-masked systemctl unmask busybox-ifplugd@usb1.service; then
         systemctl unmask busybox-ifplugd@usb1.service
     fi
     if ! is-enabled "update-engine.service"; then
--- a/package/toltec-base/package
+++ b/package/toltec-base/package
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
-# Copyright (c) 2021 The Toltec Contributors
+# Copyright (c) 2023 The Toltec Contributors
 # SPDX-License-Identifier: MIT
 
 archs=(rm1 rm2)
 pkgnames=(toltec-base)
 pkgdesc="Metapackage defining the base set of packages in a Toltec install"
 url=https://toltec-dev.org/
-pkgver=1.2-3
-timestamp=2023-05-08T19:31Z
+pkgver=1.3-1
+timestamp=2023-12-27T08:30Z
 section="utils"
 maintainer="Eeems <eeems@eeems.email>"
 license=MIT
--- a/package/toltec-base/package
+++ b/package/toltec-base/package
@@ -33,21 +33,21 @@ configure() {
     disable-unit update-engine.service
     if [[ "$arch" == "rm1" ]]; then
         echo "Disabling usb1 network device to avoid long boots"
-        ! systemctl is-masked sys-subsystem-net-devices-usb1.device && systemctl mask sys-subsystem-net-devices-usb1.device
-        ! systemctl is-masked busybox-ifplugd@usb1.service && systemctl mask busybox-ifplugd@usb1.service
+        ! is-masked sys-subsystem-net-devices-usb1.device && systemctl mask sys-subsystem-net-devices-usb1.device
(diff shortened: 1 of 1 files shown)


package/toltec-base/package
```diff
@@ -31,20 +31,23 @@ configure() {
...
+        systemctl is-masked sys-subsystem-net-devices-usb1.device && systemctl unmask sys-subsystem-net-devices-usb1.device
+        systemctl is-masked busybox-ifplugd@usb1.service && systemctl unmask busybox-ifplugd@usb1.service
     fi
 }
 
 postremove() {
-    if is-masked sys-subsystem-net-devices-usb1.device; then
+    if systemctl is-masked sys-subsystem-net-devices-usb1.device; then
         systemctl unmask sys-subsystem-net-devices-usb1.device
+        systemctl unmask busybox-ifplugd@usb1.service
```
Eeems:
```suggestion
    if is-masked sys-subsystem-net-devices-usb1.device; then
        systemctl unmask sys-subsystem-net-devices-usb1.device
    fi
    if is-masked systemctl unmask busybox-ifplugd@usb1.service; then
        systemctl unmask busybox-ifplugd@usb1.service
    fi
```
[install-lib contains an `is-masked` method](https://github.com/toltec-dev/toltec/blob/f97179cf68861222b99a3321b13ae9c61c586d8a/scripts/install-lib#L35-L37), which properly hides output, as well as only returns true if the unit is actually masked. This is why it was used previously, and not `systemctl` directly.

This change will also properly handle if one of the units was masked but the other not for some reason. Which may happen from a user poking around.

package/toltec-base/package
```diff
@@ -31,20 +31,23 @@ configure() {
...
     fi
-    if ! is-enabled "update-engine.service"; then
+    if ! systemctl is-enabled update-engine.service; then
```
Eeems:
```suggestion
    if ! is-enabled update-engine.service; then
```
[install-lib contains an `is-enabled` method](https://github.com/toltec-dev/toltec/blob/f97179cf68861222b99a3321b13ae9c61c586d8a/scripts/install-lib#L22-L24) which properly hides output and only returns true if the unit is in an enabled state. This is why it was used previously, and not `systemctl` directly.

package/toltec-base/package
```diff
@@ -31,20 +31,23 @@ configure() {
         /opt/etc/profile
     echo "Disabling automatic update"
     disable-unit update-engine.service
-    if [[ "$arch" == "rm1" ]] && ! is-masked sys-subsystem-net-devices-usb1.device; then
+    if [[ "$arch" == "rm1" ]]; then
         echo "Disabling usb1 network device to avoid long boots"
-        systemctl mask sys-subsystem-net-devices-usb1.device
-    elif [[ "$arch" == "rm2" ]] && is-masked sys-subsystem-net-devices-usb1.device; then
+        ! systemctl is-masked sys-subsystem-net-devices-usb1.device && systemctl mask sys-subsystem-net-devices-usb1.device
+        ! systemctl is-masked busybox-ifplugd@usb1.service && systemctl mask busybox-ifplugd@usb1.service
```
Eeems:
```suggestion
        if ! is-masked sys-subsystem-net-devices-usb1.device; then
            systemctl mask sys-subsystem-net-devices-usb1.device
        fi
        if ! is-masked busybox-ifplugd@usb1.service; then
            systemctl mask busybox-ifplugd@usb1.service
        fi
```
This will cause installation to fail if one of these units is already masked due to having a non-zero return code, and appending a `|| true` would keep it from failing if the mask command failed.

package/toltec-base/package
```diff
@@ -31,20 +31,23 @@ configure() {
...
+    elif [[ "$arch" == "rm2" ]]; then
         echo "Enabling usb1 network device to ensure usb SSH works"
-        systemctl unmask sys-subsystem-net-devices-usb1.device
+        systemctl is-masked sys-subsystem-net-devices-usb1.device && systemctl unmask sys-subsystem-net-devices-usb1.device
+        systemctl is-masked busybox-ifplugd@usb1.service && systemctl unmask busybox-ifplugd@usb1.service
```
Eeems:
```suggestion
        if is-masked sys-subsystem-net-devices-usb1.device; then
            systemctl unmask sys-subsystem-net-devices-usb1.device
        fi
        if  is-masked busybox-ifplugd@usb1.service; then
            systemctl unmask busybox-ifplugd@usb1.service
        fi
```

Same comments as previously.

Eeems (changes requested):
The version number and timestamp for the package need to be updated as well, otherwise it will not be built, and `opkg` will not detect this as an update.

Eeems:
Am I correct in assuming that you added `systemctl` to everything because you were testing with toltecmk? Toltecmk currently doesn't support install-lib or any of the other toltec specific parsing (https://github.com/toltec-dev/build/issues/29). I have an open PR to add them to toltecmk that you can use the build from if you wish: https://github.com/toltec-dev/build/pull/39

gbyl:
Thank you for the review! To be honest, I was not aware of these custom methods. I naively tried them in an interactive shell and they failed. That's why I appended the command name.

To check the `postremove()` method, I ran `systemctl unmask sys-subsystem-net-devices-usb1.device` on its own and it outright deleted the file. Is that expected?

Eeems:
> 
> To check the `postremove()` method, I ran `systemctl unmask sys-subsystem-net-devices-usb1.device` on its own and it outright deleted the file. Is that expected?

I would expect it to remove the symlink to `/dev/null` that masking would have created.

gbyl:
I thought it was odd too, since Toltec installs are supposed to be 100% reversible. I am not able to re-test this without reflashing. Could you make a backup of that file and test unmasking on your device?

Eeems:
> I thought it was odd too, since Toltec installs are supposed to be 100% reversible. I am not able to re-test this without reflashing. Could you make a backup of that file and test unmasking on your device?

Since it's a `*.device`, it probably is a dynamically created file on boot. Does a `systemctl daemon-reload` recreate the unit for you? If not, how about a reboot? From a quick skim of an update file, I don't see a specific unit for the device.

gbyl:
I tried both and `find / -name *.device` confirms that it did not come back. The unmasking command definitely removed it. I think we are running into something similar to this: https://unix.stackexchange.com/a/611152

Eeems:
> I tried both and `find / -name *.device` confirms that it did not come back. The unmasking command definitely removed it. I think we are running into something similar to this: https://unix.stackexchange.com/a/611152

Device files are native to systemd and automatically generated. The files do not exist on the device. Masking one will create a symlink to `/dev/null` and store that in `/etc/systemd/system`. Unmasking will remove the symlink, and thus reenable the built in unit.

What does `systemctl status sys-subsystem-net-devices-usb1.device` return for you?

gbyl:
```
$ systemctl status sys-subsystem-net-devices-usb1.device
● sys-subsystem-net-devices-usb1.device - /sys/subsystem/net/devices/usb1
     Loaded: loaded
     Active: inactive (dead)
```

```
$ systemctl enable sys-subsystem-net-devices-usb1.device
Failed to enable unit: Unit file sys-subsystem-net-devices-usb1.device does not exist.
```

Eeems:
> ```
> $ systemctl status sys-subsystem-net-devices-usb1.device
> ● sys-subsystem-net-devices-usb1.device - /sys/subsystem/net/devices/usb1
>      Loaded: loaded
>      Active: inactive (dead)
> ```
> 
> ```
> $ systemctl enable sys-subsystem-net-devices-usb1.device
> Failed to enable unit: Unit file sys-subsystem-net-devices-usb1.device does not exist.
> ```

As this is a device file, you don't enable it, it's just enabled by default, you should be able to start the unit. I would recommend not blocking while starting it, though:
```bash
systemctl start --no-block sys-subsystem-net-devices-usb1.device
```

gbyl:
Unfortunately, I got the same result. The start command did not report an error, but the status remains the same.

Eeems:
> Unfortunately, I got the same result. The start command did not report an error, but the status remains the same.

And this is after you've also rebooted your device?

gbyl:
That's right. I'm hoping it's only something local to my device from too much poking around. If you are not able to reproduce this, we can safely assume that, and resume the review of this PR

Eeems:
> That's right. I'm hoping it's only something local to my device from too much poking around. If you are not able to reproduce this, we can safely assume that, and resume the review of this PR

I'd like to make sure this is working as expected before merging. I haven't tested this package yet on my device as I've been busy with other things.

What happens if you reload udev? `udevadm control --reload-rules && udevadm trigger`

gbyl:
No worries. I ran that command and got no output. The status remains the same too. I'm not sure how to troubleshoot further without a reflash to better emulate the experience of the average Toltec user.

Eeems:
> No worries. I ran that command and got no output. The status remains the same too. I'm not sure how to troubleshoot further without a reflash to better emulate the experience of the average Toltec user.

You can use [codexctl](https://github.com/Jayy001/codexctl) to install specific OS versions easily. From there you can just [reenable toltec](https://remarkable.guide/guide/software/toltec.html#reenable-toltec-after-a-system-upgrade) and be back where you were. Minus any manual changes to the root partition you made.

gbyl:
Excellent. I can confirm it works fine on a fresh install:

```
$ systemctl status sys-subsystem-net-devices-usb1.device
● sys-subsystem-net-devices-usb1.device - /sys/subsystem/net/devices/usb1
     Loaded: loaded
     Active: inactive (dead)

systemd[1]: sys-subsystem-net-devices-usb1.device: Job sys-subsystem-net-devices-usb1.device/start timed out.
systemd[1]: Timed out waiting for device /sys/subsystem/net/devices/usb1.
systemd[1]: sys-subsystem-net-devices-usb1.device: Job sys-subsystem-net-devices-usb1.device/start failed with result 'timeout'.
```

Then manually replaying the steps Toltec goes through:

```
$ systemctl mask sys-subsystem-net-devices-usb1.device
Created symlink /etc/systemd/system/sys-subsystem-net-devices-usb1.device → /dev/null.
```

```
$ systemctl unmask sys-subsystem-net-devices-usb1.device
Removed /etc/systemd/system/sys-subsystem-net-devices-usb1.device.
```

Despite the "removed" message, this time it does come back on the next reboot as expected.

Eeems:
> Despite the "removed" message, this time it does come back on the next reboot as expected.

If you look closely at the message produced by the `systemctl mask` command you'll see that the file being removed by `unmask` was actually just a symlink to `/dev/null`. So it's not actually removing the unit file, just the symlink used to instruct systemd that it's masked.

I'm not entirely sure what was happening for you previously, as unmasking it should have been enough for systemd to just automatically start seeing it again. Since this is a device, and not a normal unit file.

gbyl:
Agreed. That makes perfect sense now. Is there anything else that should be changed in this PR?

Eeems:
I'll let you know when I get around to testing it :)

Eeems (approved):
Tested and works fine on my rM1, someone will need to retest this on a rM2

Eeems:
Tested on my new rM2 and it works :sunglasses:

PR #807, opened 2023-12-28
Add package: hstr

gbyl:
Add shell command history suggest box and manager: [hstr](https://github.com/dvorka/hstr)


gbyl:
missed the fact that hstr was already recently added to Entware

PR #808, opened 2023-12-28
Update linux-stracciatella to 5.4.70.4

Etn40ff:
Allow module unloading on RM1 and change the versioning scheme


gbyl:
~It works, but with a major caveat. If the `brcmfmac` module is blacklisted, when you try to load it manually, it cannot find the module:~
```
$ modprobe brcfmac # <-- typo: should be 'brcmfmac'
modprobe: FATAL: Module brcfmac not found in directory /lib/modules/5.4.70.4
```

**Update:**
Error caused by a typo in the command above as called out by Etn40ff. Thank you!

Etn40ff:
This is odd, on my rm2 things seems to be in order. Did you manually make any changes to /lib/modules at some point in the past? Can someone else on rm1 please test this?

Eeems:
> This is odd, on my rm2 things seems to be in order. Did you manually make any changes to /lib/modules at some point in the past? Can someone else on rm1 please test this?

I think you missed the edit at the bottom, `depmod` was run to fix `modprobe brcmfmac`

Etn40ff:
I did not miss that comment: the odd part is that is should not be required as modules.dep is part of the files that are installed

Eeems:
Could it be that the file provided is different than what depmod creates?

Etn40ff:
> when you try to load it manually, it cannot find the module:
> 
> ```
> $ modprobe brcfmac
> modprobe: FATAL: Module brcfmac not found in directory /lib/modules/5.4.70.4
> ```

Is this a typo in your comment or a typo in the command you actually ran?

gbyl:
I copy pasted the commands exactly as I ran them. So, you are right! The second time I ran `modprobe`, it worked because it didn't have the typo anymore, not because of `depmod`.

This should be good for merging as everything works as expected in rM1 and rM2.

Eeems (approved):

PR #809, opened 2023-12-29
KOReader wrapper script: fix LD_PRELOAD error

gbyl:
Current wrapper script was probably intended only for rM2. In rM1 it triggers this error indefinitely every 2 seconds:
```
ERROR: ld.so: object '/opt/lib/librm2fb_client.so.1' from LD_PRELOAD cannot be preloaded (internal error): ignored.
```

Fix: Check the device version and execute a different command accordingly

### DIFF
--- a/package/koreader/koreader
+++ b/package/koreader/koreader
@@ -1,4 +1,11 @@
 #!/usr/bin/env bash
-# Copyright (c) 2021 The Toltec Contributors
+# Copyright (c) 2023 The Toltec Contributors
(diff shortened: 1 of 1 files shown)


Eeems:
This is an expected warning, and working as intended. The error is also raised on the rM2 for any executables that it attempts to run from entware.

Ideally this would be updated to use rm2fb-client, but that needs a bit of extra work to avoid having the process show up as rm2fb-client, and instead as what is actually running, which is why the script is implemented as such here.

PR #811, opened 2023-12-31
lf: Fix missing dependency error

gbyl:
Add `libncurses-dev` as an explicit dependency of `lf`. Otherwise, when launched from yaft (`TERM=yaft-256color`), it fails with the following error:

```
creating screen: exec: "infocmp": executable file not found in $PATH
```


Eeems (approved):
lgtm

PR #812, opened 2023-12-31
Stable merge window for week 52 of 2023

Eeems:
### New Packages
- `bottom` - 0.9.6-1 (#804)

### Updated Packages
- `webinterface-onboot` - 1.2.2-3 (#794 #816)
- `lf` r31-2 - (#802 #811)
- `webinterface-upload-button` - 1.1.1-1 (#790)
- `7zip` - 23.01-1 (#805)
- `ripgrep` - 14.0.3-1 (#801)


Jayy001:
Installed and tested on `2.15.1.1189` RM2 - installed fine without any issues apart from `webinterface-onboot` in which I had to then manually run `opkg install libbfd` (but I presume thats just beacuse I manually ran `opkg install` on the singular .ipk package).

Tested `btm`, `7zz` and `rg` - no issues and worked as intended. Also tested `webinterface-upload` and was able to upload a sample file, `webinterface-onboot` was also tested and does come on at boot. Happy to approve this.

Eeems:
> Installed and tested on `2.15.1.1189` RM2 - installed fine without any issues apart from `webinterface-onboot` in which I had to then manually run `opkg install libbfd` (but I presume thats just beacuse I manually ran `opkg install` on the singular .ipk package).

That should still pull in dependencies, or fail if it can't. Looking at the package, it doesn't have an installdepends, so that's something we missed due to already having libbfd installed while testing. Automated install testing would be so nice 😔

I'll get a fix in place later today, unless @rM-self-serve beats me to it.

rM-self-serve:
This is likely due to /opt/bin/strings pointing to /opt/bin/binutils-strings rather than busybox but not correctly installing binutils-strings' dependencies.

Eeems:
> This is likely due to /opt/bin/strings pointing to /opt/bin/binutils-strings rather than busybox but not correctly installing binutils-strings' dependencies. 

I think adding the binutils-strings dependency will be enough to resolve this for now, but you'll probably want to update webinterface-onboot to explicitly run the non-/opt version of strings.

rM-self-serve:
I will do that. This is well outside of my scope but doesn't this elude to a deeper problem with packaging? Had Jay run 'strings' in the terminal it would have failed as well. May this be an upstream entware thing?

Eeems:
> I will do that. This is well outside of my scope but doesn't this elude to a deeper problem with packaging? Had Jay run 'strings' in the terminal it would have failed as well. May this be an upstream entware thing?

This does seem to be an issue with upstream. You could look into getting upstream fixed as well if you'd like.

Jayy001 (approved):

PR #817, opened 2024-01-03
add sudoku

HookedBehemoth:
It's a simple Sudoku game I wrote using [libremarkable](https://github.com/canselcik/libremarkable) and the [sudoku](https://github.com/emerentius/sudoku) rust library.

It's licensed under GPLv3.

While I developed this for the reMarkable 2 and only tested it on there, I don't see why it shouldn't work on the reMarkable 1.

2024-01-03T000951 503


Eeems:
I've noticed the naming is a little inconsistent here. The package is `rm-sudoku`, the repo is `remarkable-sudoku`, the draft file is `sudoku`, but uses the name `rm-sudoku`. What is the actual name you want used here, and presented to the user?

HookedBehemoth:
Oh excuse the confusion.
Initially I wanted to call it rm-sudoku, but later decided for the full name.
User-facing, I wanted to just have it called sudoku without any prefix.

I think I should go with just sudoku. Seems like the badge also doesn't support dashes.

Eeems (approved):

PR #818, opened 2024-01-06
Add sysfs_preload

Eeems:
A simple preload that forces any calls to /sys/power/state to use systemd instead.


Eeems:
This will need to wait for an oxide 2.8 to automatically add sysfs_preload to LD_PRELOAD if it's missing.

Eeems:
Waiting on #833

Eeems:
Closing in favour of #833

PR #825, opened 2024-01-12
[rmkit] update rmkit to latest (2024-01-11)

raisjn:
* simple: fix diacritic issues in font rendering
* genie: add config dir
* harmony: throw away persistent buffer if the size changed (this should be very rare)
* remux: add APIs for launcher-ctrl
* iago: add table shape


Eeems:
There was no updated genie package, so I can't test that.

- The new API added to remux works great!
- Harmony properly discards the buffer if the size changes
- I'm not really setup to reproduce the diacritic issues in nao, but it seems to be working as expected.
- The new table shape for iago works great!

Eeems:
Genie test passed, it's reading configuration from files in `/opt/etc/genie.conf.d`

Eeems (approved):

PR #840, opened 2024-03-11
Add package: tailscale-rm-webint

rM-self-serve:
This program is a reverse proxy that will enable interaction with the USB web interface over Tailscale. 

Tailscale needs a service to run on localhost in order to share it with the network. This program will open a port on localhost and forward requests to the USB web interface at 10.11.99.1:80.

In order to test the package, tailscale is not required. Simply verifying that the USB web interface is accessible at 10.11.99.1:80 and 127.0.0.1:80 will be sufficient.

I am the author.

Question before merging, should this package depend on tailscale? It does not require tailscale to run but does not have any purpose unless tailscale is being used.

### DIFF
--- a/package/tailscale-rm-webint/package
+++ b/package/tailscale-rm-webint/package
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+_pkgname="tailscale-rm-webint"
+pkgnames=("$_pkgname")
+pkgdesc="View the web interface if running, over tailscale"
+url="https://github.com/rM-self-serve/$_pkgname"
+pkgver=1.0.0-1
+timestamp=2024-03-11T13:03:16Z
+section="utils"
+maintainer="rM-self-serve <122753594+rM-self-serve@users.noreply.github.com>"
+license=MIT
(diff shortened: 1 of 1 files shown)


Eeems:
It might be worth just renaming this to not be tailscale specific since this just is a proxy to expose something on localhost.

All that said, it would probably be better to use socat or systemd-socket-proxyd (if it's available) to handle proxyong things here instead of having a custom built executable. This way we only need to handle being the glue to configure it instead ofqking sure a custom binary continue to build and work on the device.

rM-self-serve:
I have a working solution using systemd-socket-proxyd.

Two questions:

1. Is there any way to let a user change the port of the systemd socket? I believe the socket should listen on port 80 by default but a user may want to change this behavior. Right now I have `ListenStream=127.0.0.1:80` in the .socket file. 
2. Would the name 'webinterface-localhost' be a more apt name?

Eeems:
>   1. Is there any way to let a user change the port of the systemd socket? I believe the socket should listen on port 80 by default but a user may want to change this behavior. Right now I have `ListenStream=127.0.0.1:80` in the .socket file.

Users can override the `.socket` file to change the port.

> 
>   2. Would the name 'webinterface-localhost' be a more apt name?

Probably

PR #842, opened 2024-03-12
Add package: webinterface-localhost

rM-self-serve:
Improved version of https://github.com/toltec-dev/toltec/pull/840.

This package is a reverse proxy that will enable interaction with the USB web interface over localhost. It uses systemd-socket-proxyd with no external repository. 

Useful for exposing the USB web interface over tailscale.

I am the author.


package/webinterface-localhost/package
```diff
@@ -0,0 +1,50 @@
...
+configure() {
+    systemctl daemon-reload
+
+    if is-active "$pkgname".socket; then
+        echo "Restarting $pkgname"
+        systemctl restart "$pkgname".socket
+    fi
+
+    echo "Run the following command to use $pkgname"
+    how-to-enable "$pkgname.socket"
```
Eeems: Since this isn't changing an active launcher or something similar, it is probably one of the services we are okay with automatically enabling and starting.

Eeems (approved):

PR #865, opened 2024-05-27
Bump Version of ReTerm to Latest

Willmac16:
* [Changelog of ReTerm between these two releases](https://github.com/i-am-shodan/ReTerm/compare/release_20230628_1927...release_20230801_1556)

* I tested with `toltecmk`. The generated `.ipk` file installed fine and fixed the inverted colors issue I was having with the original version (also possible removed the duplicate icons in oxide).


Eeems:
What OS version did you test on? The testing branch is in the middle of adding 3.3.2 support, so I'll need to verify that it works on 3.3.2 against toltec testing before I merge. I'm also holding off on merging anything until the stable merge is done for ease of testing the stable merge stuff.

Willmac16:
> What OS version did you test on? The testing branch is in the middle of adding 3.3.2 support, so I'll need to verify that it works on 3.3.2 against toltec testing before I merge. I'm also holding off on merging anything until the stable merge is done for ease of testing the stable merge stuff.

This was on 3.3.2.1666 against testing

Eeems:
Alright, this will also need to be tested again 2.15. I'll try to take some time after I get the stable merge accross the finish line to validate on both my devices.

Willmac16:
> Alright, this will also need to be tested again 2.15. I'll try to take some time after I get the stable merge accross the finish line to validate on both my devices.

Although, you can't actually use ReTerm on 2.0 or the rM1 because neither have support for the folio keyboard. So maybe changing the supported device/firmware string would be more apt.

Eeems:
> Although, you can't actually use ReTerm on 2.0 or the rM1 because neither have support for the folio keyboard. So maybe changing the supported device/firmware string would be more apt.

You can connect an external keyboard, or pipe evdev input through to the device via various methods on the rM1, which is why I added ReTerm with both rM1 and rM2 support.

Willmac16:
Oh cool!

Eeems:
fb

When attempting to run this on a rM2 running 2.15.1.1189

Eeems:
Same error on the rM1

Eeems (changes requested):
This update makes ReTerm no longer compatible with 2.15 on either device. This will need to be reworked to provide the older version of ReTerm on 2.x, and the newer one on 3.x.

This would also require testing what happens when you downgrade with it installed and then reenable toltec.

Willmac16:
Is it worth it to figure out/do you know why the changes break 2.0 support?

Also, it looks like you made changes to the package file inside the ReTerm repo, do those need to be propagated into this instance of the package file.

Eeems:
> Is it worth it to figure out/do you know why the changes break 2.0 support?

It looks to be an intentional change to have the application stop if no keyboard is detected.

> Also, it looks like you made changes to the package file inside the ReTerm repo, do those need to be propagated into this instance of the package file.

They may be worth including. I haven't looked at what the changes were.

Eeems:
I've marked this as a draft, as it will need to be reworked to split the package into the OS 2.x and OS 3.x versions.

Eeems:
Due to the lack of updates to this PR, I'm closing it.

PR #869, opened 2024-05-30
DO NOT MERGE, DO NOT UPDATE FROM STABLE

Eeems:
https://github.com/zyedidia/micro/issues/3322

I'm attempting to validate the build of micro that last succeeded. I expect this to fail due to a mismatched sha256sum.


Eeems:
This is very odd, the build ran with no issue, but the generated artifact contains folders and files that should not be generated for the codebase, as the changes required were just merged into stable yesterday.

The sha256sum used in the package recipe doesn't match what I get when I manually download the file to check against as well, which is also not what I'm expecting.

Eeems:
Upon reviewing the logs, it looks like the git checkout runs the following:
```
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +720bd3919fdc80a9267ed25588779368a797cb19:refs/remotes/pull/869/merge
```

I expect that `refs/remotes/pull/869/merge` contains an automatic merge of this branch into stable, which would be why all the changes are there. I'll have to do something else to force it to use the old codebase. This doesn't answer why the sha256sum check isn't failing though.

Eeems:
Now that I've ensured that the old codebase for the build system is running, it seems to no longer work due to changes in the github runners. I'm hesitant to try to resolve them, as it may invalidate this test.

Edit: Well, we won't know if it invalidates it until we try. I'll try to resolve the libc dependency error, and lets see what happens.

PR #878, opened 2024-06-02
Fix mmc-utils build

### LINKED ISSUE #867: Unable to build mmc-utils
**Describe the issue**
git.kernel.org is now returning a 403 error when trying to download the source files

**To Reproduce**
Steps to reproduce the behaviour:
1. Attempt to build mmc-utils
2. Get the following error: `Unexpected status code while fetching source file 'https://git.kernel.org/pub/scm/utils/mmc/mmc-utils.git/snapshot/mmc-utils-7769a4d7abe339ce273c13a203394a79a11fcff9.tar.gz', got 403`

**Expected behaviour**
It should not error.

**Upstream issue**
N/A

**Screenshots**
N/A

**System information:**
 - `toltecctl status` output: N/A
 - Package: mmc-utils 1.0-0

**Additional context**
I've tried manually downloading the tar.gz file, and I also get a 403, even though the link is available on the [web interface](https://git.kernel.org/pub/scm/utils/mmc/mmc-utils.git/commit/?id=7769a4d7abe339ce273c13a203394a79a11fcff9).

Likely the workaround would be to change the recipe to use git to checkout the repository at the correct commit instead. The `wireguard` package should have an example of how to do that.

Eeems:
Fix mmc-utils build by switching to using git as we now get a 403 when trying to download the artifact from git.kernel.org


LinusCDE (approved):
lgtm

PR #890, opened 2024-06-07
7zip: Alias for 7-zip

gbyl:
[Release notes](https://www.7-zip.org/history.txt)


gbyl:
@Eeems This package might also need to be removed. Entware now has 7-zip natively. The only difference is the hyphen in the name (Toltec: `7zip` vs Entware: `7-zip`):

```
$ opkg info 7zip
Package: 7zip
Version: 23.01-1
Status: install user installed
Section: utils
Architecture: rmall
Maintainer: Eeems <eeems@eeems.email>
Size: 971259
Filename: 7zip_23.01-1_rmall.ipk
Description: A file archiver with a high compression ratio.
```
```
$ opkg info 7-zip
Package: 7-zip
Version: 23.01-1
Depends: libc, libssp, librt, libpthread, libpthread, libstdcpp
Status: unknown ok not-installed
Section: utils
Architecture: armv7-3.2
Size: 1085100
Filename: 7-zip_23.01-1_armv7-3.2.ipk
Description: File archiver with a high compression ratio.
```

Eeems:
Instead of removing it, just make a meta package that contains nothing but depends on `7-zip` to force anybody using the old one to update to the entware package.

Eeems:
Getting upstream updated would be good too

gbyl:
Could you point me to an example of a Toltec meta package? I am not sure what the bare minimum parameters are for the build to succeed

Eeems:
https://github.com/toltec-dev/toltec/blob/stable/package/toltec-deletions/package
I believe you just need a noop `package` method, so `package() { ;; }` might be enough (formatted so the linter is happy, it'll probably hate this).

gbyl:
The only thing that worked in the end was an empty `echo`. Should I remove or change other parameters? Maybe `pkgdesc="Alias for 7-zip"` or `pkgdesc="Metapackage for 7-zip"` to better represent the package?

Eeems:
> The only thing that worked in the end was an empty `echo`. Should I remove or change other parameters? Maybe `pkgdesc="Alias for 7-zip"` or `pkgdesc="Metapackage for 7-zip"` to better represent the package?

`true` instead of `echo` would avoid adding an empty line to the build output.

Updating the pkgdesc would be a good idea.

Eeems (approved):

PR #891, opened 2024-06-07
Remove package: ripgrep

gbyl:
Remove ripgrep from Toltec as it was recently added to Entware


package/ripgrep/package
```diff
@@ -1,22 +1,23 @@
...
+installdepends=(libpcre2)
 
 image=rust:v3.1
 source=("https://github.com/BurntSushi/ripgrep/archive/refs/tags/${pkgver%-*}.zip")
-sha256sums=(eb91f202fc25d51755390736775a7e7f8e146eb97e584aaa01fb0c45aa5de654)
+sha256sums=(821cb40ecddb6d539f6d8ea2490acc6b5e2a45723f884f5e14a0c115d1523b4d)
 
 build() {
-    cargo build --release
+    cargo build --release --features 'pcre2'
```
Eeems: What was the reason for this change to the build?

gbyl: PCRE2 is a feature included in all official builds of ripgrep. Entware also added ripgrep recently, but their binary triggers the error below in the remarkable. By declaring an explicit dependency on `libpcre2` in Toltec we can work around it.

```
rg: error while loading shared libraries: libpcre2-8.so.0: cannot open shared object file: No such file or directory
```

Eeems: Does this actually work? The entware libraries are linked against the entware libc, not the one on the device, and stuff built with the toltec toolchain are built against the libc that comes with the device. I'm guessing this isn't an issue due to it being a rust application?

Since entware is now shipping ripgrep, we'll need to either remove it from toltec, or, since we can't, [add an epoch to the version](https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#pkgver-field) to make sure new versions from entware don't replace ours.

Eeems: I would be in favour of getting upstream entware's dependencies fixed and removing it from toltec if you are okay with that.

gbyl: I am not sure about the technical explanation, but the binary from this PR does work.

```
$ rg --version
ripgrep 14.1.0

features:-simd-accel,+pcre2

PCRE2 10.42 is available (JIT is unavailable)
```

Are you able to test Entware's build of ripgrep? I tried their binary on a fresh install on RM1 and still got the shared library error. The only workaround I could think of for their binary is running it as `LD_LIBRARY_PATH="/opt/lib" rg --version`

Eeems: > I am not sure about the technical explanation, but the binary from this PR does work.
> 
> ```
> $ rg --version
> ripgrep 14.1.0
> 
> features:-simd-accel,+pcre2
> 
> PCRE2 10.42 is available (JIT is unavailable)
> ```
> 
> Are you able to test Entware's build of ripgrep? I tried their binary on a fresh install on RM1 and still got the shared library error. The only workaround I could think of for their binary is running it as `LD_LIBRARY_PATH="/opt/lib" rg --version`

So their binary is just broken and assumes libpcre2 is on the host system? I haven't tried anything yet, but I was going to try all your open PRs later tonight. No matter how it's broken, I do recommend getting an upstream issue or PR opened.

gbyl: @Eeems I agree. There is no use in duplicating efforts. Should I change this PR to `Remove ripgrep from Toltec`? Or should I close it and open a new one for that?

Eeems: Feel free to reuse this one. I'm guessing ripgrep is currently broken in toltec due to the entware one having a newer version?

Eeems (approved):

PR #892, opened 2024-06-07
Add package: bandwhich

gbyl:
Add Rust-based network utilization tool [bandwhich](https://github.com/imsnif/bandwhich)


Eeems:
Just a side note, we've switched toltec over to using toltecmk for the build process, so you can test locally with it to speed up the turnaround on building/testing new package recipes.

Eeems (approved):

PR #896, opened 2024-06-08
Temporarily fix remux

Eeems:
Binaries in this package that are being stripped are failing with the following error: `remux: remux: no version information available (required by remux)`

When they aren't stripped there is no error, and when manually stripping a non-stripped binary, the following warning is raised: `warning: allocated section `.dynstr' not in segmen`. The binary will then segfault if you run it on device.


Eeems:
Superceded by #898

PR #897, opened 2024-06-09
Remove package: bottom

gbyl:
[bottom](https://github.com/ClementTsang/bottom) was recently added to Entware.

The only caveat is that Entware's bottom gets installed in `/opt/usr/bin/btm` which is not included in the $PATH set by Toltec.


Eeems:
> The only caveat is that Entware's bottom gets installed in `/opt/usr/bin/btm` which is not included in the $PATH set by Toltec.

That appears to be an upstream bug, as `/opt/etc/profile` only adds `/opt/bin` and `/opt/sbin` to the path.

gbyl:
That would be my guess too going by [this file](https://github.com/Entware/entware-rust/blob/18d022fe8d537834e69896100f89bd9ea48dda4f/bottom/Makefile#L38-L41). There is a second caveat too. Entware's btm, just as with ripgrep, expects the libraries to be in the host system. So, to run it successfully in the remarkable it needs to be run like so: `LD_LIBRARY_PATH="/opt/lib" /opt/usr/bin/btm`

Eeems:
Have you opened upstream issues for either of the packages yet? That said, I didn't have the same issue you did with ripgrep on 3.3.2, I was able to run it without issue.

gbyl:
Unfortunately, I don't understand Entware well enough to open a productive issue. Thank you for testing ripgrep though! The issue might just be in my device after all. I would have to retest on a fresh install.

As far as this PR goes, btm is available directly from Entware. So, we can merge it to avoid duplicating efforts, right?

Eeems:
> Unfortunately, I don't understand Entware well enough to open a productive issue. Thank you for testing ripgrep though! The issue might just be in my device after all. I would have to retest on a fresh install.
> 
> As far as this PR goes, btm is available directly from Entware. So, we can merge it to avoid duplicating efforts, right?

I've done some looking for `/opt/usr/bin` in the entware repo, and it seems to be an expected path: https://github.com/search?q=repo%3AEntware%2FEntware%20%2Fopt%2Fusr%2Fbin&type=code

It's still not part of `/opt/etc/profile` which is part of the entware-opt package: https://github.com/Entware/rtndev/blob/4f0e97fc9ddb46317ea7c4752a81ad3e5dc52044/entware-opt/files/profile#L16-L23

It would probably still be worth it open an upstream issue asking about this discrepancy. For this PR though, we can easily add a file to toltec-base to add it to the path, similar to how we load the entware profile: https://github.com/toltec-dev/toltec/blob/eed41f1abede4150994186faa6a53612d90d88e0/package/toltec-base/package#L28

gbyl:
>we can easily add a file to toltec-base to add it to the path, similar to how we load the entware profile

I was not able to figure out how to accomplish this. Could you push that change to this PR? I would like to see the solution you had in mind.

Also, I found a [minor typo](https://github.com/toltec-dev/toltec/blob/2f13ee47197d0121946b61c6abf03c00dca17efb/package/toltec-bootstrap/toltecctl#L682) while testing solutions. Since it is such a small change, could we include it with this PR?

Eeems:
> > we can easily add a file to toltec-base to add it to the path, similar to how we load the entware profile
> 
> I was not able to figure out how to accomplish this. Could you push that change to this PR? I would like to see the solution you had in mind.

I will try to get to that sometime soon.

> 
> Also, I found a [minor typo](https://github.com/toltec-dev/toltec/blob/2f13ee47197d0121946b61c6abf03c00dca17efb/package/toltec-bootstrap/toltecctl#L682) while testing solutions. Since it is such a small change, could we include it with this PR?

Sure, but the toltec-bootstrap package will need to have its version number bumped to make the package build.

Eeems:
The version number would need to be the one bumped, as the package sources have had a change. The pkgrel is only for when there are no changes to the source, and it was just adjustments to the package recipe.
https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#pkgver-field

Eeems (approved):

gbyl:
Thank you for those commits @Eeems! I would not have guessed that implementation. Also, I reflashed my device, and no longer see those library errors with ripgrep. As you recommended, I tested the new toltecmk too and it works great.

PR #902, opened 2024-06-27
Fix linux-mainline build

Eeems:


Eeems:
@alistair23 would you be able to look at this? It looks like the debian image we are using has become unwilling to install the dependencies require for the package to build. I would try to use the v3.1 image, as it will install the dependencies, but it fails to build.

alistair23:
Having a look

alistair23:
Have a look at https://github.com/toltec-dev/toltec/pull/904 hopefully that works

Eeems:
Closed in favour of #904

PR #906, opened 2024-08-15
Build gocryptfs for rmall

icedream:
This package used to be available for reMarkable 2 on OS 3 but seems to have been limited to OS 2.x only after the split to individual OS version repos. Upgrading currently leads to the uninstallation of this package.

This PR restores sharing of the package between OS 2.x and OS 3.x by including it in rmall instead of rmallos2. As far as I can tell, there is no technical reason for why this package shouldn't be made available this way.

I am not the author of the software or the package, I'm merely contributing a change to the package here.

### DIFF
--- a/package/gocryptfs/package
+++ b/package/gocryptfs/package
@@ -2,12 +2,12 @@
 # Copyright (c) 2020 The Toltec Contributors
 # SPDX-License-Identifier: MIT
 
-archs=(rmallos2)
+archs=(rmall)
 pkgnames=(gocryptfs)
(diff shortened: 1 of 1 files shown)


icedream:
Currently unable to test installation of the generated package due to a weird issue

```
root@reMarkable:~# opkg install ./gocryptfs_2.0-beta2-3_rmall.ipk 
Package toltec-deletions is already installed on root.
```

icedream:
I figured out what's happening, closing the PR for now until we can somehow figure out the fuse situation.

PR #919, opened 2024-09-14
Bump toltec-deletions version as stable is out of sync

Eeems:
Stable is at 0.1-8 but doesn't contain the linux-mainline changes as it had to be removed from the last merge window. We just need to bump both testing and stable to 0.1-9


LinusCDE (approved):
lgtm

PR #922, opened 2024-09-14
Fix launcherctl category

Eeems:
According to our [docs](https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#section-field), it should be `launchers` not `launcher`.


LinusCDE (approved):
lgtm

PR #923, opened 2024-09-14
Fix sysfs_preload category

Eeems:
According to our [docs](https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#section-field) it should be `utils` and not `util`


LinusCDE (approved):
lgtm

PR #924, opened 2024-09-14
Add writing section as we have apps in it already

Eeems:
Sill and Keywriter are both in the writing section. As such, unless we choose to change the section they are in, we should document that section.


LinusCDE (approved):
lgtm

PR #933, opened 2024-12-21
detect architecture and give error if running on reMarkable paper pro

xugro:
Added some groundwork to the installation script for supporting rmpp.
It just quits installation if run on aarch64.

### DIFF
--- a/scripts/bootstrap/bootstrap
+++ b/scripts/bootstrap/bootstrap
@@ -23,9 +23,11 @@ fi
 arch=""
 case "$(uname -m)" in
     "armv7l")
-        arch="armv7sf-k3.2" ;;
+        arch="armv7sf-k3.2"
+        ;;
     #"aarch64)
-    #    arch="aarch64-k3.10" ;;
+    #    arch="aarch64-k3.10"
+    #    ;;
     *)
         echo "Unsupported device architecture"
         exit 1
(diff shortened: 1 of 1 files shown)


Eeems:
We do have code in toltecctl that would reject the rMPP, but we probably would need to rejig rmall to move all the non-architecture-independent packages to a different repo.

xugro:
> We do have code in toltecctl that would reject the rMPP, but we probably would need to rejig rmall to move all the non-architecture-independent packages to a different repo.

My main reason for detecting rmpp within the installation script was to download the correct wget binary for the device as the architecture is different. The rest can be handled by toltec which has existing mechanisms to do so anyways.

Eeems:
> > We do have code in toltecctl that would reject the rMPP, but we probably would need to rejig rmall to move all the non-architecture-independent packages to a different repo.
> 
> My main reason for detecting rmpp within the installation script was to download the correct wget binary for the device as the architecture is different. The rest can be handled by toltec which has existing mechanisms to do so anyways.

Fair enough, I have been thinking about this and I don't think there is a way around duplicating the code instead of sharing it between bootstrap and toltecctl.

I've updated the PR to go back to rmall, and commented out the rMPP architecture handling for now. Followup tasks will be to create a static compile of wget for the architecture, and then make sure rmall is properly architecture independent, and create new repos for architecture specific packages.

Eeems (approved):

PR #939, opened 2025-02-12
Add package for rmstylusbutton

MoritzBoehme:
I am adding the package [RMStylusButton](https://github.com/rschroll/RMStylusButton).
It's a fork of the unmaintained [RemarkableLamyEraser](https://github.com/isaacwisdom/RemarkableLamyEraser).

With the package, one can:
- switch to the eraser by holding down the button of the stylus
- double click to undo
- triple click to redo

The upstream project uses a [custom script for managing the installation](https://github.com/rschroll/RMStylusButton/blob/main/RMStylusButton/manage.sh) which I used as a basis for the package scripts.
Currently, one cannot pass flags, but I plan to add this as well. Maybe using an env file which has a variable FLAGS which is then loaded via the service file.

### DIFF
--- a/package/rmstylusbutton/package
+++ b/package/rmstylusbutton/package
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2025 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+archs=(rm2)
+pkgnames=(rmstylusbutton)
+_pkgver=3.0
+pkgver=$_pkgver-1
+pkgdesc="Use a stylus button with the reMarkable 2"
+timestamp=2024-03-04T04:30:52Z
+maintainer="Moritz <github@moritzboeh.me>"
+license=GPL-3.0
+url=https://github.com/rschroll/RMStylusButton
+section="utils"
+installdepends=(xochitl)
+conflicts=(remarkable-stylus)
+
+image=base:v3.2
+source=(
+    "$url/archive/refs/tags/v$_pkgver.tar.gz"
+    rmstylusbutton.service
+)
+sha256sums=(
+    5f468b9f78b705ca67ea3288ae923b10fb70193c236b31507a4e6148ebb4e60e
+    SKIP
+)
+
+build() {
+    export CC=arm-linux-gnueabihf-gcc
+    make
+}
+
+package() {
+    install -D -m 755 -t "$pkgdir"/opt/bin "$srcdir"/RMStylusButton/RMStylusButton
+    install -D -m 644 -t "$pkgdir"/lib/systemd/system/ "$srcdir"/rmstylusbutton.service
+}
+
+configure() {
+    systemctl daemon-reload
+    systemctl enable --now rmstylusbutton.service
+}
+
+preremove() {
+    disable-unit rmstylusbutton.service
+}
+
+postremove() {
+    systemctl daemon-reload
+}
--- a/package/rmstylusbutton/rmstylusbutton.service
+++ b/package/rmstylusbutton/rmstylusbutton.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Stylus Button Input
+After=xochitl.service
+
+[Service]
+ExecStart=/opt/bin/RMStylusButton
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target


package/rmstylusbutton/package
```diff
@@ -0,0 +1,49 @@
...
+
+package() {
+    install -D -m 755 -t "$pkgdir"/opt/bin "$srcdir"/RMStylusButton/RMStylusButton
+    install -D -m 644 -t "$pkgdir"/lib/systemd/system/ "$srcdir"/rmstylusbutton.service
+}
+
+configure() {
+    systemctl daemon-reload
+    systemctl enable rmstylusbutton.service
+    systemctl start rmstylusbutton.service
```
Eeems:
```suggestion
    systemctl enable --now rmstylusbutton.service
```

package/rmstylusbutton/package
```diff
@@ -0,0 +1,49 @@
...
+}
+
+preremove() {
+    systemctl stop rmstylusbutton.service
+    systemctl disable rmstylusbutton.service
```
Eeems:
```suggestion
    disable-unit rmstylusbutton.service
```

package/rmstylusbutton/package
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2025 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+archs=(rm2)
+pkgnames=(rmstylusbutton)
+pkgver=v3.0
```
Eeems:
```suggestion
_pkgver=3.0
pkgver=$_pkgver-1
```

package/rmstylusbutton/package
```diff
@@ -0,0 +1,49 @@
...
+timestamp=2024-03-03T20:18:54-08:00
+maintainer="Moritz <github@moritzboeh.me>"
+license=GPL-3.0
+url=https://github.com/rschroll/RMStylusButton
+section="utils"
+installdepends=(xochitl)
+
+image=base:v3.2
+source=(
+    "$url/archive/refs/tags/$pkgver.tar.gz"
```
Eeems:
```suggestion
    "$url/archive/refs/tags/v$_pkgver.tar.gz"
```

package/rmstylusbutton/package
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2025 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+archs=(rm2)
+pkgnames=(rmstylusbutton)
+pkgver=v3.0
+pkgdesc="Use a stylus button with the reMarkable 2"
+timestamp=2024-03-03T20:18:54-08:00
```
Eeems:
```suggestion
timestamp=2024-03-04T04:30:52Z
```

Eeems (changes requested):
Please update the package recipe to better match the standards we have for our recipes as a whole. I've added comments with change suggestions.

Please also add a conflicts with the `remarkable-stylus` package, as these both handle the same thing, and installing both at the same time can cause unexpected results.

MoritzBoehme:
> Please update the package recipe to better match the standards we have for our recipes as a whole. I've added comments with change suggestions.
> 
> Please also add a conflicts with the `remarkable-stylus` package, as these both handle the same thing, and installing both at the same time can cause unexpected results.

Thanks for the quick feedback! Will get right on it.

MoritzBoehme:
I have added the conflicts as well as a basic way to configure the package if desired via /opt/etc/rmstylusbutton.env. I do not know how this works in combination with package updates, though. Any changes will probably get overridden?
I'll search how other packages do it...

MoritzBoehme:
Well, it seems other packages do it similarly. Renamed the config file to end in `.conf` to make it clear that this is a config file.
Hope this complies with best practices and the style of this project.

Eeems:
Generally for this kind of configuration, users would create a systemd unit override. Configuration files are generally something that the underlying application reads. I'm not sure it's really something we would desire as a separate file in this case

MoritzBoehme:
> Generally for this kind of configuration, users would create a systemd unit override. Configuration files are generally something that the underlying application reads. I'm not sure it's really something we would desire as a separate file in this case

Ah, okay. Thanks for clarifying. I'll read up on how this would work and change the package accordingly.

Eeems:
> > Generally for this kind of configuration, users would create a systemd unit override. Configuration files are generally something that the underlying application reads. I'm not sure it's really something we would desire as a separate file in this case
> 
> Ah, okay. Thanks for clarifying. I'll read up on how this would work and change the package accordingly.

A user would run `systemctl edit rmstylusbutton`, Make their modifications to pass the flags, save and exit. This would create an override file that overrides the settings they want to override. This would persist on package upgrades. They can remove the override file if they want to revert back to stock settings.

Eeems (approved):

PR #940, opened 2025-03-01
Add Rempack

rempack-b:
Rempack is an opkg frontend for the RM2 and similar.

It was developed on and tested with an RM2, but I expect it runs fine on any device with a similar screen size.

This is the first release stable enough for public testing.

### DIFF
--- a/package/rempack/package
+++ b/package/rempack/package
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(rempack)
+pkgdesc="A user friendly package manager frontend for opkg"
+archs=(rm2)
+url="https://github.com/rempack-b/rempack"
+pkgver=1.0.5
+timestamp=2025-02-24T04:41:34-05:00
+section=admin
+maintainer="Brant Martin <toltec@brantmartin.com>"
+license=MIT
+image=base:v3.1
+installdepends=(display)
+source=(
+    http://gitlab.onionstorm.net/rolenthedeep/rempack/-/archive/1.0.5/rempack-1.0.5.zip
+)
+sha256sums=(
+    754c34ad4b4e5e208da4e27f3c7b9c7a33fdf6617263aa22a67e31e82a22dffe
+)
+build() {
+    cd $srcdir
+    mkdir output
+    cmake -S . -B output --preset "Release Native"
+    cmake --build output --target rempack
+}
+
+package() {
+    ## Commands to copy the build artifacts into a package structure
+    # Use the `install` command when possible
+    # (see `man install` or <https://linux.die.net/man/1/install>)
+
+    # The following variables are available:
+    # "$srcdir" - The directory where build() was run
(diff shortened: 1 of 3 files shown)


package/rempack/package
```diff
@@ -0,0 +1,37 @@
...
+license=MIT
+image=base:v3.1
+installdepends=(display)
+source=(
+    http://gitlab.onionstorm.net/rolenthedeep/rempack/-/archive/1.1.3/rempack-1.1.3.zip
+    rempack.draft
+    rempack.oxide
+)
+sha256sums=(
+    SKIP
```
Eeems: Only local files should skip verification, any externally fetched files need to be verified so we know they haven't been tampered with.

package/rempack/package
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(rempack)
+pkgdesc="A user friendly package manager frontend for opkg"
+archs=(rm2)
```
Eeems: I see you've tagged this as rM2 only, what about it doesn't work for a rM1?

Eeems (changes requested):

package/rempack/package
Eeems: I just tested on a rM1 and it appears to work fine.

package/rempack/package
rempack-b: fair enough

package/rempack/package
```diff
@@ -0,0 +1,37 @@
...
+
+pkgnames=(rempack)
+pkgdesc="A user friendly package manager frontend for opkg"
+archs=(rmall)
+url="https://github.com/rexxar-tc/rempack"
+pkgver=1.1.3
+timestamp=2025-03-01T04:26:09-05:00
+section=admin
+maintainer="Brant Martin <toltec@brantmartin.com>"
+license=MIT
```
Eeems, line 13: I can't find this license in the repo pointed to by the url. Without an explicit license, a repo defaults to "All Rights Reserved"

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#choosing-the-right-license

package/rempack/package
```diff
@@ -0,0 +1,37 @@
...
+image=base:v3.1
+installdepends=(display)
+source=(
+    http://gitlab.onionstorm.net/rolenthedeep/rempack/-/archive/1.1.3/rempack-1.1.3.zip
```
Eeems: This appears to be pulling from a different location than the url. If the official location for people to go is at the url, I would expect to be pulling source archives from there as well.

Eeems (changes requested):
I missed this in the first look

Eeems:
@rempack-b poke

package/rempack/package
rempack-b: License added

package/rempack/package
rempack-b: yeah, I was trying to point it at my internal CI system but really I'm just overcomplicating things

Eeems (approved):

PR #947, opened 2025-07-27
Update bootstrap wget to 1.21.1-2

timower:
Updates the wget version so it statically links against libatomic.
This ensures the toltec bootstrap works on firmware version 3.20.


timower:
Just saw I changed http to https, I'll need to fix that first.

Eeems (dismissed):
lgtm

Eeems:
> Just saw I changed http to https, I'll need to fix that first.

Ah yes, this is important as we need to download it over http so that we can use it for downloads over https.

Eeems (approved):

PR #949, opened 2025-08-25
Add fuseki and gnugo

palfrey:
Adding the fuseki GnuGo frontend (which also packages a copy of GnuGo cross-compiled for the Remarkable). I'm the author, I own both a v1 and a v2 Remarkable, but Toltec et al is only installed on the v1 as it's my "hacking around" Remarkable as opposed to my "I keep my work and personal notes in it and need not to break it" Remarkable, so I've marked this as RM1-only currently.

So far, I've implemented a Atari Go and human-vs-machine mode (the latter gets a bit slow at times, but that's because GnuGo only goes so fast running on a non-desktop machine) and it all works pretty well.


docs/building.md
```diff
@@ -18,7 +18,7 @@ Before running the build, make sure you have all the required dependencies:
 
 * Docker
 * bsdtar
-* Python 3.10
+* Python 3.11
```
palfrey, line 21: https://github.com/toltec-dev/build requires 3.11 at this point, so I couldn't install the `requirements.txt` packages on 3.10.

package/fuseki/package
```diff
@@ -0,0 +1,72 @@
...
+# See <https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#section>
+# for available sections
+section=games
+## This should be your contact information
+maintainer="Tom Parker-Shemilt <palfrey@tevps.net>"
+## A license under which the upstream source is available
+# Choose among the list of valid SPDX license identifiers
+# <https://spdx.org/licenses/>
+license=AGPL-3.0-or-later
+archs=(rm1)
```
Eeems: Since you are using libremarkable, and don't appear to be doing anything rM1 specific in your codebase, I don't see a reason to make this rmall, I'll be testing on both rM1 and rM2 anyways before this gets merged.

package/fuseki/package
Eeems, line 1: Please remove all the extra comments from this file, we have a lot of packages in the repo, we don't want to have the docstrings on every single one instead of in the documentation.

package/fuseki/package
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright (c) 2025 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(fuseki)
```
Eeems, line 5: This should be split into two packages, fuseki and gnugo. fuseki would depend on the gnugo package. You can have both packages defined in this recipe, or split them into two recipes.

package/fuseki/package
```diff
@@ -0,0 +1,72 @@
...
+sha256sums=(
+    2624c57641c7cb28699df693c6da65f6d192f9867554fcfcf830b17db851a702
+    37e4ff419a5a697039e9aa9e5b156248fcf6ed7a97867aae13aa623a7ee6a6ea
+)
+
+noextract=(638191be13be06a8d7428e7739ed947904ef3e56.zip 51f99b7a00b7acf295475b56c349c857a9aa2d0e.zip)
+
+prepare() {
+    cd build/package/fuseki/rm1/src
+    unzip -o 638191be13be06a8d7428e7739ed947904ef3e56.zip
```
Eeems: It would simplify the maintenance of this recipe to extract these archives to named folder for what they are that doesn't include the git commit hash. That way you can reference them by gnugo and fuseki in the other methods and not have to update them when a version number bump happens.

package/fuseki/package
```diff
@@ -0,0 +1,72 @@
...
+    unzip -o 51f99b7a00b7acf295475b56c349c857a9aa2d0e.zip
+}
+
+makedepends=("build:autoconf" "build:automake")
+
+build() {
+    cd gnugo-51f99b7a00b7acf295475b56c349c857a9aa2d0e
+    autoreconf || true
```
Eeems: What are you swallowing with the `|| true` here?

package/fuseki/package
```diff
@@ -0,0 +1,72 @@
...
+    make
+    cd ../fuseki-638191be13be06a8d7428e7739ed947904ef3e56
+    cargo build --release
+}
+
+package() {
+    install -D -m 755 "$srcdir"/gnugo-51f99b7a00b7acf295475b56c349c857a9aa2d0e/interface/gnugo "$pkgdir"/opt/bin/gnugo
+    install -D -m 755 "$srcdir"/fuseki-638191be13be06a8d7428e7739ed947904ef3e56/target/armv7-unknown-linux-gnueabihf/release/fuseki "$pkgdir"/opt/bin/fuseki
+    install -D "$srcdir"/fuseki-638191be13be06a8d7428e7739ed947904ef3e56/fuseki.draft "$pkgdir"/etc/draft/fuseki.draft
+    sed -i -e 's#/home/root/fuseki#env GNUGO_BINARY=/opt/bin/gnugo /opt/bin/fuseki#g' "$pkgdir"/etc/draft/fuseki.draft
```
Eeems: Instead of modifying this file, a copy of it should be in this repository next to the package recipe, and have the correct values.

The modification is also not compatible with oxide, so an oxide application registration would be needed, that or to use an intermediary script that gets placed in /opt/bin instead that sets up the environment and calls fuseki instead.

Eeems (changes requested):

package/fuseki/package
palfrey: I'll do them as two recipes. Do you want one PR for both, or shall I submit gnugo one first separately?

Eeems: A single PR is fine.

package/fuseki/package
palfrey: Done

package/fuseki/package
palfrey: Done

package/fuseki/package
palfrey: I've removed the `archs` line

package/fuseki/package
palfrey: No longer needed now I've split out the packages

package/fuseki/package
palfrey: Mostly that the autotools are a terrible piece of software. AFAIK running it twice like this with automake in the middle is both the easiest and the advised way to do things, in that it's an idempotent program intended to fix what it can, which isn't everything. Running it again after automake works, but automake needs the first autoreconf run to work.

package/fuseki/package
palfrey: Done with an intermediary script

package/fuseki/package
Eeems: Could you add a comment with information as to why it's being done this way, a link to automake documentation that recommends this would be fine.

palfrey: I have seen those instructions _somewhere_ but it's been some time and I can't find them again. Turned out to be easier to find the set of commands autoreconf runs (which includes the automake) and figure out the set that needed running to make autoreconf work properly.

Eeems:
I will try to getting to testing this sometime this week, along with the other changes in this milestone. No promises though, as I'm in the middle of a heatwave, which makes doing anything outside of my dayjob kinda difficult at the moment.

palfrey:
> I will try to getting to testing this sometime this week, along with the other changes in this milestone. No promises though, as I'm in the middle of a heatwave, which makes doing anything outside of my dayjob kinda difficult at the moment.

No worries. I finally got closed out a PR I opened 2 years ago, so if you're able to do faster than that, all good :) More seriously, I know well the load of maintainership, so whenever's good for you, no rush.

Eeems:
Life has other plans, it'll have to be later. Perhaps next weekend, but with how things are looking likely later than that.

palfrey:
> Life has other plans, it'll have to be later. Perhaps next weekend, but with how things are looking likely later than that.

How's this going? Any way I can help?

Eeems:
Sorry for how long this is taken, I've had several life events happen in a row, including one today that is taking all of my attention. If you can get one of the other toltec maintainers to give this a test, approve the PR, and merge it, that could help move things forward faster. Most have them have gotten extremely busy in their personal lives as well though.

package/fuseki/fuseki-runner.draft
```diff
@@ -0,0 +1,5 @@
+name=fuseki
+desc=Play go
+call=/opt/bin/fuseki-runner
```
Eeems, line 3: When installing the package, this like ends up being `call=/home/root/fuseki`, so something is overwriting this in the build.

package/fuseki/package
```diff
@@ -0,0 +1,36 @@
...
+
+build() {
+    cargo build --release
+}
+
+package() {
+    install -D -m 755 "$srcdir"/target/armv7-unknown-linux-gnueabihf/release/fuseki "$pkgdir"/opt/bin/fuseki
+    install -D -m 644 -t "$pkgdir"/opt/etc/draft "$srcdir"/fuseki.draft
+    install -D -m 755 -t "$pkgdir"/opt/bin "$srcdir"/fuseki-runner
+    install -D "$srcdir"/fuseki.png "$pkgdir"/etc/draft/icons/fuseki.png
```
Eeems:
```suggestion
    install -D "$srcdir"/fuseki.png "$pkgdir"/opt/etc/draft/icons/fuseki.png
```

Eeems (changes requested):
Issues to resolve:

- Upon install, oxide will not pick up the new application as the draft config points at an executable that doesn't exist
- fuseki icon is placed in `/etc` instead of `/opt/etc`

package/fuseki/fuseki-runner.draft
palfrey: Fixed. This was due to https://github.com/palfrey/fuseki/blob/main/fuseki.draft

palfrey:
>     * Upon install, oxide will not pick up the new application as the draft config points at an executable that doesn't exist
>     * fuseki icon is placed in `/etc` instead of `/opt/etc`

Both are now resolved

package/fuseki/package
```diff
@@ -0,0 +1,36 @@
...
+
+build() {
+    cargo build --release
+}
+
+package() {
+    install -D -m 755 "$srcdir"/target/armv7-unknown-linux-gnueabihf/release/fuseki "$pkgdir"/opt/bin/fuseki
+    install -D -m 644 "$srcdir"/fuseki-runner.draft "$pkgdir"/opt/etc/draft/fuseki.draft
+    install -D -m 755 -t "$pkgdir"/opt/bin "$srcdir"/fuseki-runner
+    install -D "$srcdir"/fuseki.png "$pkgdir"/opt/etc/draft/icons/fuseki.png
```
Eeems: This results in the image being executable, which doesn't make sense, it probably should include `-m 644`

Eeems (approved):
Will merge if the final test passes

PR #952, opened 2025-09-11
Update zerotier-one to 1.16.0

Eeems:
[Release Notes](https://github.com/zerotier/ZeroTierOne/blob/1.16.0/RELEASE-NOTES.md#2025-08-21----version-1160)


Eeems:
Upon reviewing the release notes, they have changed some code to MPL, and other code to a commerical source available license. The code that is commerical source available is not something we are compiling/distributing, so that should be fine.

PR #954, opened 2025-10-22
Change Oxide sentry endpoint

Eeems:
Update Sentry URL in oxide_sentry.cpp for GlitchTip.

I'm migrating away from sentry due to how resource hungry it is to self-host.


Eeems:
I've handled this another way

PR #957, opened 2025-11-03
Add syncthingctl

### LINKED ISSUE #841: Add syncthingctl to toltec
**Package information**
syncthingctl (from https://github.com/Martchus/syncthingtray)

**Reason for addition**
This is a CLI that orchestrates syncthing, which is already in the repo.  With it, it'd be possible to configure / manage syncthing via an on-device terminal, without needing a browser or another computer on the same network to access the web interface.

palfrey:
This in theory should fix https://github.com/toltec-dev/toltec/issues/841. However, it's a hard one as syncthingctl depends on _multiple_ libraries we don't have available in the base image versions of Debian that are all made by the same author, and AFAIK are only really used by his projects.

This PR so far does the following:
* [cpp-utilities](https://github.com/Martchus/cpp-utilities) - core requirement. Seems to build fine, but I think there's some issues with the cmake packaging given complaints later on
* [qtutilities](https://github.com/Martchus/qtutilities) - currently build fails due to paths for cpp-utilities headers that are installed fine, but get the wrong paths. That I've got a `mkdir /opt/include` hack in it's build so far is also a bad sign.
* [qtforkawesome](https://github.com/Martchus/qtforkawesome) - feels like we shouldn't need this to build the cli, but no, the apps depends on this anyways. Doesn't build yet due to a lack of `qtutilities` build.
* [syncthingtray](https://github.com/Martchus/syncthingtray) - probably needs a few libraries adding, haven't gotten far enough with the other dependencies to really check that list.

I don't have enough cmake knowledge to really follow this one up, so pushing to ask a wider audience if they've got any ideas.


Eeems:
Should syncthingctl have `installdepends=(syncthing)`?

palfrey:
> Should syncthingctl have `installdepends=(syncthing)`?

Probably! I'll check that if I can ever get it building...

palfrey:
I've managed to fix the `libcpp-utilities` build, but I think I've now hit a hard block. The entware qt packages seem to contain just the `.so` files, not any of the headers or the cmake files. I'd somewhat worked around this by just installing the Debian qt libraries, which is fine up until you actually try and link with any of them 😆 Given the debian and entware versions are different (and are likely to get more so over time v.s. the single minor version difference right now), mixing those two won't work.

AFAIK, the only way this will ever work is going through entware and making qt dev packages (including a qt gui one, because apparently qtutilities always needs that for some reason). Or theoretically making qt packages here as well, but both of those are a giant pile of work, and I think I'm done on this one.

Eeems:
Toltec doesn't use the entware qt libraries, as qt is shipped on the device. You'll want to use the qt image and use the system qt libraries instead. These should already have header files. If there is a specific library that is not included in the system qt libraries, that would need to be built and added. You can reference https://github.com/toltec-dev/toolchain/blob/v3.x/qt/Dockerfile for how qt was built for the device, and how it's installed.

palfrey:
> You'll want to use the qt image and use the system qt libraries instead

I had somehow completely forgotten that was a thing! So I've unfudged the qtutilities build, but qtforkawesome is another matter, as despite it's readme claiming it supports 5.6, it [actually uses at least one 6.0 function](https://github.com/Martchus/qtforkawesome/issues/6). OTOH, it's a small one, so can probably patch around it...

Eeems:
The qt:v4.0 image has qt6, as rM has updated to qt6 in later OS versions. This wont be available in toltec until #955 has been merged though. For now (and to keep supporting older OS versions) patching around it does seem like the easiest solution.

palfrey:
`qtforkawesome` is fixed with a patch, and I've gotten as far as the actual `syncthingctl`! Brief moment of panic when it looked like it needed Boost, but turns out you can switch that off. OTOH, [it also needs Qt 6](https://github.com/Martchus/syncthingtray/issues/381). Hopefully again that's a fairly simple patch, but that's one for another day right now.

Martchus:
If you just need `syncthingctl` you don't need `qtforkawesome`. If you nevertheless do then I should also fix that.

Martchus:
> Hopefully again that's a fairly simple patch, but that's one for another day right now.

I pushed that patch for you. Maybe further changes are required, though.

I also pushed a patch that let's you avoid dealing with `qtforkawesome` altogether if you are only interested in `syncthingctl`.

palfrey:
I've tested this locally on my rm1 (started `syncthing`, went to `/opt/etc/syncthing/` and ran `syncthingctl -f config.xml --cert https-cert.pem status`) and LGTM. All seems to be running fine.

package/syncthingctl/package
```diff
@@ -0,0 +1,42 @@
...
+url="https://martchus.github.io/syncthingtray/"
+pkgver=2.0.2-1
+timestamp=2025-09-26T16:18Z
+section=utils
+maintainer="Tom Parker-Shemilt <palfrey@tevps.net>"
+license=GPL-2.0-or-later
+image=qt:v3.3
+makedepends=(build:cmake host:libcpp-utilities-dev host:libcpp-utilities host:libqtutilities host:libqtutilities-dev)
+installdepends=(libcpp-utilities libqtutilities syncthing)
+# We need commit version for Qt version and qtforkawesome removal patches
```
Eeems: This comment does not seem relevant anymore.

package/libcpp-utilities/package
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(libcpp-utilities libcpp-utilities-dev)
+pkgdesc="Common C++ classes and routines used by Martchus' applications"
+url=https://github.com/Martchus/cpp-utilities
+pkgver=5.31.0-1
+timestamp=2025-09-26T16:18Z
```
Eeems:
```suggestion
timestamp=2025-09-20T23:53Z
```
This is the timestamp of the release tag.

https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#timestamp-field

package/qtforkawesome/package
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(libqtforkawesome libqtforkawesome-dev)
+pkgdesc="Library that bundles ForkAwesome for use within Qt applications"
+url=https://github.com/Martchus/qtforkawesome
+pkgver=0.3.1-1
+timestamp=2025-09-26T16:18Z
```
Eeems:
```suggestion
timestamp=2025-11-04T20:30Z
```
This timestamp matches the 42fe0be74ad6ee075d91f2b96d9bc3eb7717c9d5 commit.

https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#timestamp-field

package/qtutilities/package
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(libqtutilities libqtutilities-dev)
+pkgdesc="Common Qt related C++ classes and routines used by Martchus' applications"
+url=https://github.com/Martchus/qtutilities
+pkgver=6.18.2-1
+timestamp=2025-09-26T16:18Z
```
Eeems:
```suggestion
timestamp=2025-09-24T09:03Z
```
This is the timestamp of the v6.18.2 release tag.

https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#timestamp-field

package/syncthingctl/package
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+# Copyright (c) 2025 The Toltec Contributors
+# SPDX-License-Identifier: MIT
+
+pkgnames=(syncthingctl)
+pkgdesc="CLI that orchestrates syncthing"
+url="https://martchus.github.io/syncthingtray/"
+pkgver=2.0.2-1
+timestamp=2025-09-26T16:18Z
```
Eeems:
```suggestion
timestamp=2025-11-04T08:50Z
```

This is the timestamp of the 67a4ced74d79b5152e438464f60b7da6f40f9a4a commit.

https://github.com/toltec-dev/toltec/blob/stable/docs/package.md#timestamp-field

Eeems (changes requested):
I'll have to figure out how to test on rM1/2 sometime soon. I have some nits about the packages though that will need to be addressed.

package/syncthingctl/package
palfrey: It is. That particular commit version gets us both https://github.com/Martchus/syncthingtray/commit/4e9979f6ceed8ebd8e4129cdfe55329db3882b46 (qt 5.15.1 compat) and https://github.com/Martchus/syncthingtray/commit/67a4ced74d79b5152e438464f60b7da6f40f9a4a (optional qtforkawesome). Admittedly, the latter we could do without, but we don't actually need it for syncthingctl anymore. I've left the qtforkawesome package in this PR anyways given I've done the work though.

Eeems: To me this initially read like a TODO style comment. Which means that it's not clear enough to be useful in it's current form as it's adding confusion. The comment should be clarified. Perhaps something like:
```suggestion
# This specific commit fixes qt5 support and removes the need for qtforkawesome.
# We can target version tags after the next release where these fixes are included.
```

The comment in qtforkeawesome should similarly be updated.

Martchus: I have released 2.0.3 which contains the needed commits.

package/qtforkawesome/package
Martchus: I have released 0.3.2 which contains this commit.

Eeems:
@palfrey do you have any recommendations for how to test this? I've never used syncthing, and I'm not really planning on becoming a user.

palfrey:
I don't no. I'm also not a user, it was just on the list of things people wanted and an interesting problem! @mooomooo as you originally asked for this, can you make suggestions?

Martchus:
If you execute unit/integration tests as part of the package build you should be good. You can additionally invoke `syncthingctl --help` to see whether all dependencies are present.

Eeems:
`--help` does work, but I'm seeing the following warnings we'll need to resolve as part of the packaging:

```shell
reMarkable: ~/packages/syncthing/ syncthingctl -s
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
QFSFileEngine::open: No file name specified
Error: Unable to locate Syncthing config file and no API key specified
reMarkable: ~/packages/syncthing/ syncthingctl edit
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
QFSFileEngine::open: No file name specified
Error: Unable to locate Syncthing config file and no API key specified
```
Likely we'll just want to create a wrapper script for syncthingctl that sets XDG_RUNTIME_DIR before calling it, similar to how we have `/opt/bin/xochitl` that handles setting environment variables for `xochitl` before launching it.

Eeems (changes requested):
See above comment

palfrey:
> Likely we'll just want to create a wrapper script for syncthingctl that sets XDG_RUNTIME_DIR before calling it, similar to how we have `/opt/bin/xochitl` that handles setting environment variables for `xochitl` before launching it.

Done

package/syncthingctl/package
```diff
@@ -36,6 +38,6 @@ build() {
 }
 
 package() {
-    install -D -m 755 -t "$pkgdir"/opt/bin \
-        "$srcdir"/install/opt/bin/syncthingctl
+    install -D -m 755 "$srcdir"/syncthingctl "$pkgdir"/opt/bin/syncthingctl
+    install -D -m 755 "$srcdir"/install/opt/bin/syncthingctl "$pkgdir"/opt/bin/syncthingctl.real
```
Eeems:
```suggestion
    install -D -m 755 "$srcdir"/install/opt/bin/syncthingctl "$pkgdir"/opt/libexec/syncthingctl
```

package/syncthingctl/syncthingctl
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+export XDG_RUNTIME_DIR=/tmp/runtime-root
+
+exec /opt/bin/syncthingctl.real $@
```
Eeems:
```suggestion
exec /opt/libexec/syncthingctl.real $@
```

Eeems (changes requested):
Lets put the real executable in libexec so it's not polluting the PATH.

Eeems:
@palfrey poke

palfrey:
> Lets put the real executable in libexec so it's not polluting the PATH.

Done

Eeems (approved):

